PR #165, opened 2012-05-29
[2.0] Registration timeouts and modules

aquanight:
Allow m_dnsbl, m_ident, m_sqlauth, and similar modules to not cause 'registration timeouts' if they are being particularly slow.

One side effect is that modules blocking the registration process must now handle timing out clients on their own.


attilamolnar:
I don't think this is necessary. The current definition of the registration timeout (the maximum amount of time we are willing to wait for a client to become registered) is fine.

PR #335, opened 2012-10-12
New threadengine based on C++11

ChrisTX:
The idea behind this is to add a C++11 threading primitive based engine. I've tested it with VC11 and it does seem to work fine.

I'm not sure how useful this is after all; the actual reason I made this was because I was bored.

Please note that this class does not implement ThreadSignalSocket since I was unsure how to exactly do this and the class is little documented. As it's only used in m_mssql and m_mysql, this shouldn't be a problem unless you use these modules.

### DIFF
--- a/include/threadengines/threadengine_stdlib.h
+++ b/include/threadengines/threadengine_stdlib.h
@@ -1,0 +1,123 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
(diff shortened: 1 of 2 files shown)


attilamolnar:
I appreciate your work, however, I'll close this as no branch of inspircd is using C++11 presently and won't be until the next major release at least, and it is also abandoned and targets the insp20 branch.

SaberUK:
Plus, when we do target C++11 we can just get rid of the thread engines entirely and just use std::thread.

PR #466, opened 2013-04-03
Removed config option for +L

Shawn-Smith:
Since the config option was only there to preserve backwards compatibility in 2.0, I've removed it for the new version.


SaberUK:
Was there a reason this was not merged? It seems to be okay to me.

SaberUK:

PR #468, opened 2013-04-03
Usermode z

Shawn-Smith:
This adds usermode +z for ssl-only queries and notices.


SaberUK:
A rebased version of this which retains 2.0 compatibility by default is available at https://github.com/inspircd/inspircd/compare/master...SaberUK:master+sslumode.

Issue #463, opened 2013-04-02
[SUGGESTION]: +B caps-percent


westor7:
Hello,

I wanna request to make the +B working with a custom set for any channel with an custom percent of CAPS block and not using the networks configuration setting.

e.g: /mode #chan +B 70
this will block all caps messages that has 70% characters.
- Thanks!

PR #473, opened 2013-04-05
Made blockcaps a per-channel setting.

Shawn-Smith:
This changes the `m_blockcaps` from a config-defined, network-wide setting to a per-channel setting defined with mode parameters.

The new syntax is `+B percent-to-match:minimum-length`
Example: `+B 50:10` will kick for anything over 10 characters that contains 50% or more caps.

You can also specify `~` or `*` for blocking or banning in the same manner as m_messageflood

Example: `+B *55:10` will ban for anything over 10 characters that contains 55% or more caps.
Example: `+B ~55:10` will block anything over 10 characters that contains 55% or more caps.

The current `m_blockcaps` module was moved to `m_blockcaps_compat` so networks that still have 2.0 servers linked may use that.

Requested in: #463

### DIFF
--- a/docs/conf/helpop-full.conf.example
+++ b/docs/conf/helpop-full.conf.example
@@ -867,7 +867,9 @@ Closes all unregistered connections to the local server.">
  A                  Allows anyone to invite users to the channel
                     (normally only chanops can invite, requires
                     allowinvite module).
- B [percent]:[len]  Blocks messages with too many capital letters.
+ B [~*][percent]:[len] kicks users for messages with too many capital letters.
+                    If the ~ is used it will block instead of kick.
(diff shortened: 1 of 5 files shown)


SaberUK:
I'm wondering if making [my anticaps module from extras](https://github.com/inspircd/inspircd-extras/blob/master/3.0/m_anticaps.cpp) a vendor module and deprecating blockcaps might be a better alternative to this. What do people think?

SaberUK:
Anticaps was added in 36e6dc7bc8c21db2f8fa63c96b3e3aa23886e056 which deprecates blockcaps and includes the behaviour of this pull request. Thanks for your contribution though even if it did not end up being merged.

PR #581, opened 2013-07-04
Custom channel prefixes

Adam-:


docs/conf/inspircd.conf.example
```diff
@@ -616,7 +616,10 @@
 
          # nosnoticestack: This prevents snotices from 'stacking' and giving you
          # the message saying '(last message repeated X times)'. Defaults to no.
-         nosnoticestack="no">
+         nosnoticestack="no"
+
+         # channelprefixes: Prefixes allowed for channels.
```
SaberUK: It'd be better if this was calculated automatically from the chantypes tags.

PR #598, opened 2013-08-01
Spoofs!

Adam-:


attilamolnar:
I'm not convinced that we need this. It is a niche feature and I believe it is simpler to run connections through localhost. A dumb connection forwarder does the trick and is also safer as the inspircd process has no way to access the remote IP so it is impossible to leak the IP due to a bug.

About the implementation, how does this interact with other modules that try to use the user's IP, such as ident and dnsbl?
There appears to be no prioritization for the event used so it is possible that some modules will operate on the original IP when handling the user while other modules may get the spoofed IP, no?

SaberUK:
After discussing this with Adam we have decided that it would be better as an extras module or an external piece of software.

PR #635, opened 2013-09-15
Add more connect class debug logs.

m4z:
For me, these helped understand why some connections did not match the connect class I expected they would.

Please feel free to adjust the messages to your liking.


SaberUK:
Apologies for the late response but the insp20 branch is not receiving feature changes at this time. 

Please re-file this against the master branch if you are still interested. Thanks!

PR #640, opened 2013-10-08
Patch to m_ldapauth to allow whitelisting multiple nicks and idents

mivok:
This change allows you to put nickname and ident patterns inside ldapwhitelist
tags, similarly to how IPs are currently whitelisted with the cidr attribute.

This lets you supply multiple nickname patterns to whitelist (vs allowpattern,
which only allows one), as well as allowing whitelisting based on ident
instead (which is most useful when you have userfield set to use the ident
instead of nickname for ldap searches).

Example usage:

```
<ldapwhitelist nick="Guest*">
<ldapwhitelist nick="*Bot">
<ldapwhitelist ident="testing">
```

### DIFF
--- a/src/modules/extra/m_ldapauth.cpp
+++ b/src/modules/extra/m_ldapauth.cpp
@@ -104,6 +104,8 @@ class ModuleLDAPAuth : public Module
 	std::string password;
 	std::string vhost;
 	std::vector<std::string> whitelistedcidrs;
+	std::vector<std::string> allownickpatterns;
+	std::vector<std::string> allowidentpatterns;
 	std::vector<std::pair<std::string, std::string> > requiredattributes;
 	int searchscope;
 	bool verbose;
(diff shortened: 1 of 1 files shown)


SaberUK:
Hey,

Apologies for how long it has taken for this to be reviewed.

You can already whitelist nicks with `<ldapauth:allowpattern>` but this method is inconsistent from matching hosts. Also, instead of adding separate fields for nick/ident/host/ip it would probably be better if we had `<ldapwhitelist:mask> which took a n!u@h or n@u!ip/cidr mask and then compared that against the user.

Thanks!

mivok:
Hi! Thanks for taking the time to review it. Sadly I haven't touched inspircd in 4 years since I switched companies, and won't be able to work on this. I don't have a need for this any more, so feel free to close if you wish.

SaberUK:
No worries, I have opened #1478 instead for if someone else wants to implement this. Thanks anyway.

Issue #648, opened 2013-10-24
[correctness] BufferedSocket::GetNextLine() rfc1459 framing overrun


kaniini:
`\r`, `\0`, `\n` and `\r\n` should all be handled as line terminators.

attilamolnar:
> Then you are not in compliance with section 8 of RFC1459 (section 8 describes, as you mentioned in your quote, the defacto state of ircd).  `\r` and `\n` should be treated the same, per section 8 of RFC1459.

I don't think we should be bug for bug compatible with other implementations, especially when the RFC uses the words "unfortunately" and "backwards compatibility". At this time, there is no need for this backwards compatibility measure.

I think @kaniini's comment in https://github.com/inspircd/inspircd/issues/1123#issuecomment-168048947 applies here as well:

> either way, this is an example of the RFC codifying sub-optimal behavior. as I said on #charybdis the other day to the reporter, interpreting the RFC literally will result in the same type of bad decisions as interpreting other religious texts literally.

PR #664, opened 2013-11-08
[Blatant RFC Violation] src/users: Comply to RFC standards

Xe:
\r == \n

This closes issue #648: [security] BufferedSocket::GetNextLine() rfc1459 framing overrun

### DIFF
--- a/src/users.cpp
+++ b/src/users.cpp
@@ -274,7 +274,6 @@ void UserIOHandler::OnDataReady()
 				c = ' ';
 				break;
 			case '\r':
-				continue;
 			case '\n':
 				goto eol_found;
 			}


attilamolnar:
This is not needed. Please see https://github.com/inspircd/inspircd/issues/648.

To summarize:

Messages should always be terminated with a CRLF:
> IRC messages are always lines of characters terminated with a CR-LF (Carriage Return - Line Feed) pair
> ...
> `<message>  ::= [':' <prefix> <SPACE> ] <command> <params> <crlf>`
> ...
> `<crlf>     ::= CR LF`

Section 8 says

> Unfortunately, due to backward compatibility requirements, the implementation of some parts of this document varies with what is laid out.  On notable difference is:
> * recognition that any LF or CR anywhere in a message marks the end of that message (instead of requiring CR-LF);

The current way of handling this is correct, because comforming clients always send CRLF, so acting upon the `\n` and ignoring the `\r` is an implementation detail.

I don't think we should be bug for bug compatible with other implementations, especially when the RFC uses the words "unfortunately" and "backwards compatibility". At this time, there is no need for this backwards compatibility measure.

PR #687, opened 2013-11-20
modules/m_ircv3: Add support for the CHGHOST client capability

Xe:
And also add the relevant documentation from the ircv3 specifications.


attilamolnar:
This has been implemented as a separate module.

PR #744, opened 2014-01-23
[WIP] Implement the IRCv3.3 XINFO command.

SaberUK:
This is still very much work in progress. A lot of stuff needs to still be implemented and the code needs to be cleaned up somewhat. However, the XINFO command should work and some basic commands have been implemented.

/cc @Elizafox


SaberUK:
Closing this as it seems like XINFO is not going to be a thing.

PR #892, opened 2014-07-25
m_spanningtree: Add an optional KEY param to SVSJOIN

ShutterQuick:
So ajoin on services can join users to keyed channels with SVSJOIN without issuing an invite first.


CuleX:
@attilamolnar This affects 2.0, too, and breaks Anope 2.0's NickServ AJOIN command. Any chances this could get backported to 2.0?

SaberUK:
@CuleX This involves a S2S protocol change which can not happen in a patch release.

CuleX:
@SaberUK This is a non-breaking change: The new parameter is optional. Not supplying it does not modify exiting behavior.

This fixes interoperability issues that have confusing semantics to users. The message shown when SVSJOIN for a keyed channel without key is issued is "Cannot join channel (Incorrect channel key)", which confuses users. This is an issue that is exposed to end-users right now.

I believe these benefits outweigh an auto-no based on semantics. Please correct me if I'm reaching and this change is not actually non-breaking.

PR #902, opened 2014-08-06
Replacement for m_nationalchars

Googolplexed:
I've had this sitting around on my HDD for a while, it replaces m_nationalchars.
Unpolished, but I've kinda forgotten what I was working on before putting it up here.

Might want a few eyes on it, and such.


attilamolnar:
Please update the coding style to match the style of the project, I'll point out the different instances as comments in the code

src/modules/m_nationalcodepage.cpp
```diff
@@ -0,0 +1,118 @@
...
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
```
attilamolnar: Extra newline missing

src/modules/m_nationalcodepage.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include "inspircd.h"
+#include "caller.h"
+
+class ModuleNationalCodePage : public Module,public HandlerBase1<bool, const std::string&>
+{
+	caller1<bool, const std::string&> rememberer;
+	const unsigned char * lowermap_rememberer;
```
attilamolnar: `*` (and `&`) should be next to the type

src/modules/m_nationalcodepage.cpp
```diff
@@ -0,0 +1,118 @@
...
+	std::bitset<256> m_allowedchar;
+	unsigned char m_casemap[256];
+	std::string charset,casemapping;
+
+public:
+	ModuleNationalCodePage()
+		: rememberer(ServerInstance->IsNick), lowermap_rememberer(national_case_insensitive_map) {
```
attilamolnar: See init list style here: https://github.com/inspircd/inspircd/blob/master/src/modules/m_delayjoin.cpp#L42

src/modules/m_nationalcodepage.cpp
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	//Check if Nick is valid using the current codepage
+	bool Call(const std::string& nick)
+	{
+		if(nick.length() > ServerInstance->Config->Limits.NickMax)
```
attilamolnar: There should be a space after `if`, `while`, etc.

src/modules/m_nationalutf.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include "inspircd.h"
+#include "caller.h"
+
+class ModuleNationalUTF : public Module,public HandlerBase1<bool, const std::string&>
+{
+	caller1<bool, const std::string&> rememberer;
+	struct Range {
+		unsigned long min;
+		unsigned long max;
+		Range(unsigned long mi,unsigned long ma)
```
attilamolnar: Space missing between the parameters

src/modules/m_nationalutf.cpp
```diff
@@ -0,0 +1,206 @@
...
+		 */
+
+		//MSB is 7, but we've already handled that in ltr, so start at 6
+		//Only a maximum of 4 bits can be set, including the initial
+		int bytes = 1; //Already got a header byte
+		for(int i=6; ltr[i]!=0 && i>3; i--)
+		{
+			//For true bit in the UTF header, we grab an extra byte.
+			it++;
+			if(it == str.end()) return 0;         //Prevent invalid header from iterating too far.
```
attilamolnar: These should be one separate lines, and the comment should be either on its own line or next to the `if` separated by one space. C++-style single line comments should have a space between the `//` and the text.

    if (it == str.end()) // Prevent invalid header from iterating too far.
	    return 0;

or

    // Prevent invalid header from iterating too far.
    if (it == str.end())
	    return 0;

src/modules/m_nationalcodepage.cpp
```diff
@@ -0,0 +1,118 @@
...
+			return false;
+
+		for(std::string::const_iterator it = nick.begin(); it != nick.end(); ++it)
+		{
+			//Check if character in normal range
+			if ((*it >= 'A') && (*it <= '}'))
```
attilamolnar: It's preferable to assign the result of `*it` to a variable once and inspect that instead of using the `*it` syntax in many places

src/modules/m_nationalcodepage.cpp
```diff
@@ -0,0 +1,118 @@
...
+				continue;
+
+			//Check if character is okay, but needs to be after the initial
+			if ((((*it >= '0') && (*it <= '9')) || (*it == '-')) && (it != nick.begin()))
+				continue;
+
+			//std::string's iterator returns signed chars.
```
attilamolnar: Incorrect, the signedness of `char` is defined by the implementation

src/modules/m_nationalcodepage.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+		//Use InspIRCd config engine to read tags for characters in decimal
+		//Split a range of codepoints in decimal form
+		ConfigTagList chars = ServerInstance->Config->ConfTags("cpchars");
+		for (ConfigIter i = chars.first; i != chars.second; ++i)
+		{
+			ConfigTag* tag = i->second;
+			std::string range = tag->getString("range");
+			irc::portparser coderange(range, false);
+			int j=-1;
```
attilamolnar: Spaces are missing before and after the `=`, but in this case the assignment is pointless as the value is overwritten on the next line

src/modules/m_nationalcodepage.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+	void On005Numeric(std::map<std::string, std::string>& tokens) CXX11_OVERRIDE
+	{
+		tokens["CASEMAPPING"] = charset;
+	}
+
+	void ReadConfig(ConfigStatus& status) CXX11_OVERRIDE
+	{
+		//Copy the existing generic ASCII casemap into place
+		memcpy(m_casemap, rfc_case_insensitive_map, 256);
```
attilamolnar: Use sizeof() instead of a raw number here

src/modules/m_nationalutf.cpp
```diff
@@ -0,0 +1,206 @@
...
+		//If it didn't, the below shift will compensate, moving them into
+		//proper position
+		unsigned long utf32 = total >> ((bytes * -6)+30);
+
+		if(utf32 > 0x10FFFF) { return 0; } //Decoded to greater than the maximum value in the UTF8 spec
+
+		return utf32;
+	}
+
+	bool isValidUTFChar(const unsigned long& utf32)
```
attilamolnar: Pass by value

src/modules/m_nationalutf.cpp
```diff
@@ -0,0 +1,206 @@
...
+		{
+			min = mi;
+			max = ma;
+		};
+	};
+	std::vector<Range> ranges;
+
+	//Returns the next character in UTF32 encoding and advances the iterator
+	//past it. Returns 0 on failure
+	unsigned long utf8to32(std::string str,std::string::const_iterator& it)
```
attilamolnar: Either pass `str` by const ref or pass the .end() iterator if that's all what you use `str` for

src/modules/m_nationalutf.cpp
```diff
@@ -0,0 +1,206 @@
...
+{
+	caller1<bool, const std::string&> rememberer;
+	struct Range {
+		unsigned long min;
+		unsigned long max;
+		Range(unsigned long mi,unsigned long ma)
+		{
+			min = mi;
+			max = ma;
+		};
```
attilamolnar: `;` shouldn't be here

src/modules/m_nationalcodepage.cpp
```diff
@@ -0,0 +1,118 @@
...
+	{
+		return Version("8bit national character codepage support", VF_VENDOR|VF_COMMON);
+	}
+
+	~ModuleNationalCodePage()
+	{
+		ServerInstance->IsNick = rememberer;
+		national_case_insensitive_map = lowermap_rememberer;
+	}
+
```
attilamolnar: No need for a blank line after the last method of a class

SaberUK:
@Googolplexed Is this obsoleted by #1224?

Googolplexed:
#1224 includes Unicode support, but not 8-bit character set support. The Unicode part of this is entirely obsolete though.

PR #916, opened 2014-09-09
Advertise m_kicknorejoin's upper bound. Closes #913

### LINKED ISSUE #913: m_kicknorejoin.cpp: implement customizable upper-bound and advertise it on CAPAB
Per today's discussion on #atheme, it was determined that if inspircd announced an upper-bound on +J during linking, that atheme would prefer that value over the hardcoded upper-bound it presently uses.

The upper-bound should probably be restricted to the current hardcoded upper-bound in inspircd 2.2 (60 seconds), to avoid resource exhaustion attacks.

ShutterQuick:


src/modules/m_spanningtree/capab.cpp
```diff
@@ -33,6 +33,11 @@ std::string TreeSocket::MyModules(int filter)
 	std::string capabilities;
 	for (ModuleManager::ModuleMap::const_iterator i = modlist.begin(); i != modlist.end(); ++i)
 	{
+		// 2.2 advertises its maxbound for the benefit of services
```
attilamolnar: Not sure what maxbounds is, say "settings" instead to be clear

attilamolnar:
Ok to me, please rebase and I'll merge

attilamolnar:
Thanks!

PR #959, opened 2014-12-29
Fix a regression in m_cloaking

Alef-Burzmali:
5c7db14a5 made the cloak use hex-encoded hashes instead
of the binary output in master.


SaberUK:
Perhaps we should rename Generate to GenerateHex to avoid this problem in the future?

attilamolnar:
Thanks!

PR #960, opened 2014-12-30
Improvements to UpdateTime().

SaberUK:
See the "Availability" section of "man 3 clock_gettime" for more
information on this #define.

http://linux.die.net/man/3/clock_gettime


src/inspircd.cpp
```diff
@@ -578,21 +583,29 @@ InspIRCd::InspIRCd(int argc, char** argv) :
 
 void InspIRCd::UpdateTime()
 {
-#ifdef _WIN32
+#if defined __MACH__
+	clock_serv_t cs;
+	host_get_clock_service(mach_host_self(), CALENDAR_CLOCK, &cs);
+	mach_timespec_t mts;
+	clock_get_time(cs, &mts);
+	mach_port_deallocate(mach_task_self(), cs);
```
attilamolnar: Is there a need to do the `host_get_clock_service()` and `mach_port_deallocate()` calls every UpdateTime()? Wouldn't it be more wiser to do it once on startup and shutdown?

SaberUK:
I'm closing this as:

1. Apparently OpenBSD doesn't support _POSIX_TIMERS in violation of the specification.
2. The OS X change doesn't give a noticeable performance boost.

PR #961, opened 2015-01-03
Fix a snotice in m_spanningtree incorrectly using an unsigned long.

SaberUK:
This value can be negative so it should be a signed long not an unsigned long.


attilamolnar:
Thanks!

PR #966, opened 2015-01-16
Fix a blank error message when trying to load a module with a path.

SaberUK:
Without this patch when you try to load a module with a path it does this:

     [*] Loading module:	/path/to/module.so

     [*]


attilamolnar:
Thanks! It actually showed the last error message in this case, not always a blank one.

PR #967, opened 2015-01-16
Fix modules not being versioned on Windows.

SaberUK:
Also, deduplicate MODULE_INIT (we are using a default implementation of DllMain).


attilamolnar:
Please make this as simple as possible by limiting code changes to what is necessary to fix the issue. We can remove dllmain in master.

SaberUK:
@attilamolnar Done.

attilamolnar:
Thanks!

PR #968, opened 2015-01-17
cmd_list: Give +s precedence to +p

ShutterQuick:
When +ps is set, handle it as +s (most restrictive) instead of +p


attilamolnar:
Nice spot, thanks!

PR #970, opened 2015-01-22
Various preprocessor-related improvements.

SaberUK:
- Merge INSPIRCD_REVISION into INSPIRCD_VERSION.
- Remove the Windows-specific definition of MODULE_INIT.
- Use DLLManager::RetrieveLastError on all platforms.
- Fix various includes.
- Don't define DLLManager and ModuleDLLManager on static builds.


attilamolnar:
Why use/replace `#ifdef` with `#if defined`?

SaberUK:
In which file? In `inspircd.cpp` there is no real reason. I just used that when I rewrote it.

PR #971, opened 2015-01-24
Use ERR_YOUREBANNEDCREEP instead of NOTICE when a user is banned.

SaberUK:
This behaviour is specified in RFC 1459 and is used by Charybdis and other IRC server implementations so we should probably also use it.


attilamolnar:
LGTM, what do you think @Adam-?

Adam-:
Yes

attilamolnar:
Thanks!

PR #981, opened 2015-02-02
* Add characterset option to MySQL database connection

flopraden:
Hi,

I loaded the nationalchar for utf-8 and utf-8 sqlauth isn't working since the MySQL connection wasn't configured for utf-8

Here the patch to add an option for character-set.
It may be interessant to put it for others SQL drivers

Best

### DIFF
--- a/src/modules/extra/m_mysql.cpp
+++ b/src/modules/extra/m_mysql.cpp
@@ -259,10 +259,10 @@ class SQLConnection : public SQLProvider
 		std::string characterset;
 		if (config->readString("characterset", characterset))
 		{
-		  if (!mysql_set_character_set(connection,characterset.c_str()))
-		  {
-		    ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT, "WARNING: Could not change character-set to " + characterset);
-		  }
+			if (!mysql_set_character_set(connection, characterset.c_str()))
+			{
+				ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT, "WARNING: Could not change character-set to " + characterset);
+			}
 		}
 		std::string initquery;
 		if (config->readString("initialquery", initquery))
--- a/src/modules/extra/m_mysql.cpp
+++ b/src/modules/extra/m_mysql.cpp
@@ -255,6 +255,15 @@ class SQLConnection : public SQLProvider
 		bool rv = mysql_real_connect(connection, host.c_str(), user.c_str(), pass.c_str(), dbname.c_str(), port, NULL, 0);
 		if (!rv)
 			return rv;
+		// enable character-set settings
+		std::string characterset;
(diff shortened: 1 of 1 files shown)


src/modules/extra/m_mysql.cpp
```diff
@@ -255,6 +255,15 @@ class SQLConnection : public SQLProvider
 		bool rv = mysql_real_connect(connection, host.c_str(), user.c_str(), pass.c_str(), dbname.c_str(), port, NULL, 0);
 		if (!rv)
 			return rv;
+		// enable character-set settings
+		std::string characterset;
+		if (config->readString("characterset", characterset))
+		{
+		  if (!mysql_set_character_set(connection,characterset.c_str()))
```
SaberUK: Your indentation is a bit wonky here. Tabs only please.

Also, you might want to fix the lack of a space after `connection,`

flopraden:
It should be corrected.

src/modules/extra/m_mysql.cpp
```diff
@@ -255,6 +255,15 @@ class SQLConnection : public SQLProvider
 		bool rv = mysql_real_connect(connection, host.c_str(), user.c_str(), pass.c_str(), dbname.c_str(), port, NULL, 0);
 		if (!rv)
 			return rv;
+		// enable character-set settings
+		std::string characterset;
+		if (config->readString("characterset", characterset))
+		{
+			if (!mysql_set_character_set(connection, characterset.c_str()))
```
attilamolnar: This check is incorrect, quote from the [manual](http://mirror.metrocast.net/mysql/doc/refman/5.5/en/mysql-set-character-set.html):

```
Return Values

Zero for success. Nonzero if an error occurred.
```

src/modules/extra/m_mysql.cpp
```diff
@@ -255,6 +255,15 @@ class SQLConnection : public SQLProvider
 		bool rv = mysql_real_connect(connection, host.c_str(), user.c_str(), pass.c_str(), dbname.c_str(), port, NULL, 0);
 		if (!rv)
 			return rv;
+		// enable character-set settings
+		std::string characterset;
+		if (config->readString("characterset", characterset))
```
attilamolnar: Are you deliberately using `readString()` to allow empty strings? If not, please use `getString()` as the code above and test for empty() to determine whether or not to call the mysql_ func.

flopraden:
In fact no.
I just use the same config-read block that m_mysql use for initialquery.
In fact, for initialquery, it should be rewrite this getString no?

attilamolnar:
Thanks, squashed your commits and made some style changes to match our coding style.

attilamolnar:
https://github.com/inspircd/inspircd/commit/b3bafc6aaee4b6b95c6aaefcc6f2e80eca6bd370

Issue #978, opened 2015-01-31
m_sqloper doesn't support swhois


B00mX0r:
The m_sqloper module does not support an swhois field. This is a suggestion to make it so m_sqloper will recognize a column in the MySQL table called "swhois", and treat it like the swhois in opers.conf. This suggestion can also apply to other fields, such as override.

B00mX0r:
https://github.com/inspircd/inspircd/pull/983

Issue #292, opened 2012-09-01
Opers in SQL table does not appear in /stats o


Polsaker:
hi, i'm using the module sqloper, and the opers that i add to the mysql table not appear in the /stats o (only appears the opers from the config file)

SaberUK:
Fixed in #983. Thanks for reporting!

PR #983, opened 2015-02-03
Add m_sqloper compatibility for custom columns

B00mX0r:
m_sqloper reads additional columns in its MySQL table. This means columns like swhois and override can be added, and they'll automatically be used. Adding a column in the ircd_opers table is like adding a tag in an oper block.

This fixes https://github.com/inspircd/inspircd/issues/978


aeular:
Any chance of getting a fixed version of this into live? Would be nice to be able to use cert fingerprints with sql db

SaberUK:
@B00mX0r Can you rebase and squash this?

B00mX0r:
@SaberUK @aeular I updated the PR to add compatibility with other modules that hook into /OPER, including m_sslinfo for fingerprints.

B00mX0r:
Another note: if this is merged, it will be trivial to PR a fix to #292 by modifying
https://github.com/inspircd/inspircd/blob/master/src/coremods/core_stats.cpp#L319 to pull from Config->oper_blocks.

attilamolnar:
@B00mX0r Thanks for your continued interest. I like the capabilities enabled by your work, but I disagree with the new direction taken recently. Namely, only updating opers on /REHASH is a step backwards from what we had previously. Can you make it so the oper block is created from SQL at oper-up time?

B00mX0r:
@attilamolnar Do you mean make it so /OPER causes the server to re-fetch the oper blocks, or revert to the old method and remove compatibility with m_sslinfo completely?

If you mean revert to the old method, the problem with m_sslinfo is that it has an OnPostConnect hook which triggers oper-up instead of an /OPER command. As such, I could figure out no way to issue m_sslinfo an oper block from sql without either putting it in the config ahead of time, or by giving m_sqloper a function that returns an oper block. I asked @SaberUK about the latter option in #inspircd.dev yesterday and, as I suspected, module functions can only be void or ModResult.

Like you said, these can be two separate issues, but the problem is that I don't see any way to make m_sslinfo's OnPostConnect hook compatible with the old method. Unless, of course, you make it hacky and make the OnPostConnect actually call HandleCommand for OPER. 


On a separate note, as it stands right now with the current PR, sql oper blocks override config oper blocks. Is this how we want m_sqloper to work, or should config oper blocks override?

attilamolnar:
I mean revert to the old method for now to allow additional fields for e.g. swhois, then I'll take a look at refactoring the hooks to be more flexible. If you put the code doing the SQL query in a method then if we later change our minds and decide to have it called on every /OPER it will be easy to move the call.

B00mX0r:
@attilamolnar I made oper blocks update again upon executing /oper. Please let me know if I misunderstood your request.

As it stands, oper blocks from sql are stored in the config, and updated upon either execution of /oper or server rehash. The only module compatibility issue should be that m_sslinfo auto-login blocks will be outdated if they were changed after the last /rehash and after the last execution of the /oper command. I think the best solution to this last compatibility issue is an OnPreOper hook, but that's not my decision to make.

aeular:
Where does this stand on being merged into master?  Also is this ready at this point for use on a live server?

B00mX0r:
@aeular: I've done robust testing on this and found no current errors besides the one I mentioned in my previous comment. This conflicts with #1267 but if that is implemented then it is trivial for me to change a few lines of code to make it compatible. I use this on my server and have used it for over a month, so I personally think it's ready.

Of course, the code hasn't been reviewed by @attilamolnar yet, so take this with a grain of salt. This is not production code. It is still in development and has not been reviewed by someone else yet.

B00mX0r:
Retested and updated to work with the latest changes from master. I also fixed some bugs, and deduped some of the code.

Note: oper blocks from the config file will now override oper blocks from m_sqloper. This is different from the current behavior. Scenario: an IRC network that gives leaf owners netadmin on their own leafs and globop everywhere else. Now, the leaf owner will have a row in the sqloper DB table giving them globop, but their own leaf will have them as netadmin in the opers.conf file.

src/modules/m_sqloper.cpp
```diff
@@ -19,46 +20,83 @@
...
-#include "modules/hash.h"
 
-class OpMeQuery : public SQLQuery
+class OpGetQuery : public SQLQuery
 {
  public:
-	const std::string uid, username, password;
-	OpMeQuery(Module* me, const std::string& u, const std::string& un, const std::string& pw)
-		: SQLQuery(me), uid(u), username(un), password(pw)
+	bool OnUserCommand; //Record whether or not the OpGetQuery (AKA reloading the list of SQLopers) is done as a result of /OPER, or something like a /rehash or module reload
```
SaberUK: Is this necessary? If there is no uid you can just not pass one and check if `uid.empty()`.

src/modules/m_sqloper.cpp
```diff
@@ -19,46 +20,83 @@
...
+	const std::string uid, username, password; //We want to store the username and password if it's because of an /OPER, as this SQL class will never return anything to the SQLOper module, meaning we're responsible for actually executing the /OPER command post-DB fetch.
```
SaberUK: Comment like this should go on the previous line with a space after the `//`

src/modules/m_sqloper.cpp
```diff
@@ -19,46 +20,83 @@
...
+	const ServerConfig::OperIndex conf_blocks; //This variable will store all the OPER blocks that are from the config file. This is so we are always aware of which OPER blocks aren't from this module.
+	OpGetQuery(Module* me, bool ouc, const std::string& u, const std::string& un, const std::string& pw, const ServerConfig::OperIndex& co) //oper_blocks passed as opers in conf
+		: SQLQuery(me), OnUserCommand(ouc), uid(u), username(un), password(pw), conf_blocks(co)
```
SaberUK: Can you split these out so they are more readable? i.e.

```
   : SQLQuery(me)
   , uid(u)
   , username(un)
    (etc)
```

src/modules/m_sqloper.cpp
```diff
@@ -19,46 +20,83 @@
...
-		while (res.GetRow(row))
+		while (res.GetRow(row)) //Iterate through DB results to create oper blocks from sqloper rows
 		{
-			if (OperUser(user, row[0], row[1]))
-				return;
+			parameterlist cols;
+			res.GetCols(cols);
+
+			ConfigItems* items;
+			reference<ConfigTag> tag = ConfigTag::create("oper", "<m_sqloper>", 0, items); //Create the oper tag as if we were the conf file.
```
SaberUK: You can use MODNAME instead of "<m_sqloper" here.

src/modules/m_sqloper.cpp
```diff
@@ -121,40 +144,47 @@ class ModuleSQLOper : public Module
...
+				return MOD_RES_DENY; //We need to reload oper blocks from the DB before other hooks can run (i.e. sslinfo). We will re-call /OPER later.
 			}
 			ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT, "database not present");
 		}
-		return MOD_RES_PASSTHRU;
+		return MOD_RES_PASSTHRU; //There is either no DB or we successfully reloaded oper blocks
 	}
 
-	void LookupOper(User* user, const std::string &username, const std::string &password)
+	void GetOLines(bool OnUserCommand, const std::string u, const std::string &un, const std::string &pw)
```
SaberUK: OLines is a really archaic term. Can you use something more appropriate like OperBlocks?

Also, when modifying lines where the &* is against the variable name you can move them to be against the type name.

src/modules/m_sqloper.cpp
```diff
@@ -121,40 +144,47 @@ class ModuleSQLOper : public Module
...
-		userinfo["username"] = username;
-		userinfo["password"] = hash ? hash->Generate(password) : password;
+		SQL->submit(new OpGetQuery(this, OnUserCommand, u, un, pw, conf_blocks), query); //Call to SQL query to fetch oper list from SQL table.
+		return;
+	}
 
-		SQL->submit(new OpMeQuery(this, user->uuid, username, password), query, userinfo);
+	void Prioritize() CXX11_OVERRIDE
+	{
+		ServerInstance->Modules.SetPriority(this, I_OnPreCommand, PRIORITY_FIRST); //We need to run before m_sslinfo's hook to ensure that the DB is updated before it checks fingerprints. We do this by having first priority and then issuing a DENY, prevent other hooks from running until we can update the DB and call them ourselves.
```
SaberUK: Is it just m_sslinfo? If so, then you should just prioritise before it.

src/modules/m_sqloper.cpp
```diff
@@ -121,40 +144,47 @@ class ModuleSQLOper : public Module
 		else
 			SQL.SetProvider("SQL/" + dbid);
 
-		hashtype = tag->getString("hash");
-		query = tag->getString("query", "SELECT hostname as host, type FROM ircd_opers WHERE username='$username' AND password='$password' AND active=1;");
+		query = tag->getString("query", "SELECT * FROM ircd_opers WHERE active=1;");
+		const std::string noexist = "null"; //Because OpGetQuery never returns something to ModuleSQLOper, it is resposible for executing /OPER if necessary. Therefore, it requires params such as username and pwd that aren't necessary if it's updating for /REHASH.
```
SaberUK: Is there any reason why you can't just make these parameters empty?

src/modules/m_sqloper.cpp
```diff
@@ -121,40 +144,47 @@ class ModuleSQLOper : public Module
...
-		SQL->submit(new OpMeQuery(this, user->uuid, username, password), query, userinfo);
+	void Prioritize() CXX11_OVERRIDE
+	{
+		ServerInstance->Modules.SetPriority(this, I_OnPreCommand, PRIORITY_FIRST); //We need to run before m_sslinfo's hook to ensure that the DB is updated before it checks fingerprints. We do this by having first priority and then issuing a DENY, prevent other hooks from running until we can update the DB and call them ourselves.
 	}
 
 	Version GetVersion() CXX11_OVERRIDE
 	{
-		return Version("Allows storage of oper credentials in an SQL table", VF_VENDOR);
+		return Version("Allows storage of oper blocks in an SQL table", VF_VENDOR);
```
SaberUK: As a user-facing description this was probably more accurate before.

src/modules/m_sqloper.cpp
```diff
@@ -121,40 +144,47 @@ class ModuleSQLOper : public Module
 		else
 			SQL.SetProvider("SQL/" + dbid);
 
-		hashtype = tag->getString("hash");
-		query = tag->getString("query", "SELECT hostname as host, type FROM ircd_opers WHERE username='$username' AND password='$password' AND active=1;");
+		query = tag->getString("query", "SELECT * FROM ircd_opers WHERE active=1;");
+		const std::string noexist = "null"; //Because OpGetQuery never returns something to ModuleSQLOper, it is resposible for executing /OPER if necessary. Therefore, it requires params such as username and pwd that aren't necessary if it's updating for /REHASH.
+
+		GetOLines(false, noexist, noexist, noexist); //Update sqloper list from the database.
+		return;
```
SaberUK: In several places there are returns at the end of a function. These are unnecessary and can be removed.

src/modules/m_sqloper.cpp
```diff
@@ -121,40 +144,47 @@ class ModuleSQLOper : public Module
...
+	CullResult cull()
+	{
+		ServerInstance->Config->oper_blocks = conf_blocks; //Reset oper blocks to those from the config
+		return Module::cull();
 	}
 
 	ModResult OnPreCommand(std::string &command, std::vector<std::string> &parameters, LocalUser *user, bool validated, const std::string &original_line) CXX11_OVERRIDE
 	{
-		if (validated && command == "OPER" && parameters.size() >= 2)
+		if (validated && command == "OPER" && parameters.size() >= 2 && original_line != "SQLOPER") //The line will be SQLOPER only if this module already reloaded the DB and is recalling /oper for other hooks
```
SaberUK: Rather than checking original_line here can't you have a bool `active` sentinel in the class? The core (which calls events) is single threaded so this will work fine unless you need to modify it from the SQL thread.

src/modules/m_sqloper.cpp
```diff
@@ -121,40 +144,47 @@ class ModuleSQLOper : public Module
...
-		hashtype = tag->getString("hash");
-		query = tag->getString("query", "SELECT hostname as host, type FROM ircd_opers WHERE username='$username' AND password='$password' AND active=1;");
+		query = tag->getString("query", "SELECT * FROM ircd_opers WHERE active=1;");
+		const std::string noexist = "null"; //Because OpGetQuery never returns something to ModuleSQLOper, it is resposible for executing /OPER if necessary. Therefore, it requires params such as username and pwd that aren't necessary if it's updating for /REHASH.
+
+		GetOLines(false, noexist, noexist, noexist); //Update sqloper list from the database.
+		return;
+	}
+
+	CullResult cull()
```
SaberUK: Is there any reason you are doing this in cull rather than in the destructor?

extras/m_sqloper.postgresql.sql
```diff
@@ -1,9 +1,12 @@
...
     id serial NOT NULL,
-    username text,
+    "name" text,
     "password" text,
-    hostname text,
+    "hash" text,
+    "host" text,
     "type" text,
+    "fingerprint" text,
+    autologin boolean NOT NULL DEFAULT 0,
```
SaberUK: If you're going to quote these can you quote all of them.

Also, you could consider making the mandatory fields in all of the tables `NOT NULL`.

src/modules/m_sqloper.cpp
B00mX0r: I think it's best to leave the possibility for future modules that, like sslinfo, attempt to hook into part of the /OPER system and assume that the OPER blocks are up-to-date. Calling out sslinfo specifically is because right now, as far as I know the only module that does this is sslinfo. I can, however, think of other ones that would need this in the future, such as the TOTP two-factor module proposed elsewhere.

SaberUK: In that case the comment should be updated to be more generic.

B00mX0r:
@SaberUK Thanks for the review. I resolved your comments and updated the description in modules.conf.example.

B00mX0r:
Repushed the same commit to trigger a travis rebuild. The failure it showed was a result of an apt issue in one of the travis confs, and not of the module itself.

B00mX0r:
Changes in the last commit:
- Switched conf_blocks to my_blocks, where a vector of the names of sqlopers are recorded (allows compat with other modules that modify conf_blocks, i.e. ldapoper in the future)
- Removed the snomask notice for duplicate oper blocks, since a large server that used sqloper in the scenario I described would then receive 50 notices about duplicate oper blocks every time someone /oper'd

B00mX0r:
Fixed merge conflict with 2.0 docs change.

B00mX0r:
Rebased.

src/modules/m_sqloper.cpp
```diff
@@ -19,46 +20,104 @@
 
 #include "inspircd.h"
 #include "modules/sql.h"
-#include "modules/hash.h"
 
-class OpMeQuery : public SQLQuery
+class OpGetQuery : public SQLQuery
 {
  public:
+	// This variable will store all the OPER blocks that are from the config file. This is so we are always aware of which OPER blocks aren't from this module.
```
SaberUK: Can you add line breaks to your comments. Ideally they shouldn't be longer than the line of code associated with them.

src/modules/m_sqloper.cpp
```diff
@@ -19,46 +20,104 @@
...
 	{
 	}
 
 	void OnResult(SQLResult& res) CXX11_OVERRIDE
 	{
-		ServerInstance->Logs->Log(MODNAME, LOG_DEBUG, "result for %s", uid.c_str());
-		User* user = ServerInstance->FindNick(uid);
-		if (!user)
-			return;
+		// Make it so we can use a legible variable
```
SaberUK: legible variable?

src/modules/m_sqloper.cpp
```diff
@@ -19,46 +20,104 @@
...
+		ServerConfig::OperIndex* oper_blocks =& ServerInstance->Config->oper_blocks;
```
SaberUK: =& is not an operator. The space should be before the & not after.

Better still, can oper_blocks just be a reference?

src/modules/m_sqloper.cpp
```diff
@@ -19,46 +20,104 @@
...
+
+		// Remove our previous blocks from oper_blocks for a clean update
+		for (std::vector<std::string>::const_iterator i = my_blocks.begin(); i != my_blocks.end(); ++i)
+		{
+			(*oper_blocks).erase(*i);
```
SaberUK: `->`

src/modules/m_sqloper.cpp
```diff
@@ -121,35 +175,57 @@ class ModuleSQLOper : public Module
...
+	}
+	void GetOperBlocks(const std::string u, const std::string& un, const std::string& pw)
+	{
+		active = true;
+		// Call to SQL query to fetch oper list from SQL table.
+		SQL->submit(new OpGetQuery(this, my_blocks, u, un, pw), query);
+	}
 
-		SQL->submit(new OpMeQuery(this, user->uuid, username, password), query, userinfo);
+	 // Run before any other /OPER hooks that expect the oper blocks to be populated, i.e. sslinfo or a TOTP module. We issue a DENY first, and will re-run OnPreCommand later to trigger the other hooks post-DB update.
```
SaberUK: This comment should be inside the Prioritize call.

src/modules/m_sqloper.cpp
```diff
@@ -19,46 +20,104 @@
...
+				continue;
+			}
+
+			OperInfo* ifo = new OperInfo(type); // Create blank oper info.
+
+			ifo->type_block = tblk->second->type_block;
+			ifo->oper_block = tag;
+			ifo->class_blocks.assign(tblk->second->class_blocks.begin(), tblk->second->class_blocks.end());
+			(*oper_blocks)[name] = ifo;
+			my_blocks.push_back(name);
```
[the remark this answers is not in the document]
B00mX0r: Skipped per line 78 (
```
if (oper_blocks->find(name) != oper_blocks->end())
				continue;
```
)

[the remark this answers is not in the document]
B00mX0r: This way, oper blocks from the config file take priority over oper blocks from sqloper. This is intended behavior. Say we're on a server where server admins get netadmin on their own server, and globop on other servers. They have themselves as netadmin in their own opers.conf (which they control-- make sense), and then one globop entry for them in the SQL db for m_sqloper.

src/modules/m_sqloper.cpp
B00mX0r: Not a pointer in this case.

src/modules/m_sqloper.cpp
B00mX0r: I fixed the issues with comments.

src/modules/m_sqloper.cpp
```diff
@@ -19,49 +20,110 @@
 
 #include "inspircd.h"
 #include "modules/sql.h"
-#include "modules/hash.h"
 
-class OperQuery : public SQL::Query
+class OpsGetQuery : public SQL::Query
```
SaberUK: There's no need to change this.

src/modules/m_sqloper.cpp
```diff
@@ -19,49 +20,110 @@
...
+		my_blocks.clear();
 
-		// multiple rows may exist
 		SQL::Row row;
+		// Iterate through DB results to create oper blocks from sqloper rows
 		while (res.GetRow(row))
 		{
-			if (OperUser(user, row[0], row[1]))
-				return;
+			parameterlist cols;
```
SaberUK: This is an inappropriate usage of a protocol interface typedef. Just use a std::vector<std::string>.

systocrat:
Hey, been using this PR on my network for a couple weeks now and I haven't noticed any problems in testing.
I'm looking forward to this being in mainline 3.0, is there any estimate on whether this will be merged @SaberUK ?

PR #984, opened 2015-02-05
Allow U-lined services to bypass m_silence (configurable)

Renegade334:
Currently, it is possible for users to "mute" U-lined services through the use of m_silence. This should rarely, if ever, be desired behaviour.

Adds config options to m_silence to specify whether U-lined services should bypass silence mask(s).

### DIFF
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -1841,7 +1841,10 @@
 #<module name="m_silence.so">
 #
 # Set the maximum number of entries allowed on a user's silence list.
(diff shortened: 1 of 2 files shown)


SaberUK:
:+1:

attilamolnar:
Hello! I'm sure this is not the case with callerid because it's enforced locally, meaning any pseudoserver can bypass it freely. See check on line 382.

Renegade334:
Right you are - didn't read closely enough.

attilamolnar:
Thanks!

PR #988, opened 2015-02-06
* [modules/sqlopers] * Using Compare from HashProviders to check password

flopraden:
comparing salted password could not be done with SQL query

### DIFF
--- a/src/modules/m_sqloper.cpp
+++ b/src/modules/m_sqloper.cpp
@@ -154,8 +154,6 @@ public:
 
 	void LookupOper(User* user, const std::string &username, const std::string &password)
 	{
-
-
 		ParamM userinfo;
 		SQL->PopulateUserInfo(user, userinfo);
 		userinfo["username"] = username;
--- a/src/modules/m_sqloper.cpp
+++ b/src/modules/m_sqloper.cpp
@@ -24,9 +24,9 @@
 class OpMeQuery : public SQLQuery
 {
  public:
-	const std::string uid, username, password;
-	OpMeQuery(Module* me, const std::string& u, const std::string& un, const std::string& pw)
-		: SQLQuery(me), uid(u), username(un), password(pw)
+	const std::string uid, username, password, hashtype;
+	OpMeQuery(Module* me, const std::string& u, const std::string& un, const std::string& pw, const std::string& ht)
+		: SQLQuery(me), uid(u), username(un), password(pw), hashtype(ht)
 	{
 	}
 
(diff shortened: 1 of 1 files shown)


src/modules/m_sqloper.cpp
```diff
@@ -76,9 +76,21 @@ class OpMeQuery : public SQLQuery
...
+		if(hash)
+		{
+			if(!hash->Compare(password,hashedpass))
+		    		return false;
+		}
+		else
+		{
+			if(hashedpass.compare(password) != 0)
+				return false;
+		}
```
attilamolnar: This logic isn't right, if they set a hash and it was not found we shouldn't compare the input with the hash

src/modules/m_sqloper.cpp
```diff
@@ -24,9 +24,9 @@
 class OpMeQuery : public SQLQuery
 {
  public:
-	const std::string uid, username, password;
-	OpMeQuery(Module* me, const std::string& u, const std::string& un, const std::string& pw)
-		: SQLQuery(me), uid(u), username(un), password(pw)
+	const std::string uid, username, password, hashtype;
```
attilamolnar: hashtype should be a const ref

src/modules/m_sqloper.cpp
```diff
@@ -142,14 +154,13 @@ class ModuleSQLOper : public Module
 
 	void LookupOper(User* user, const std::string &username, const std::string &password)
 	{
-		HashProvider* hash = ServerInstance->Modules->FindDataService<HashProvider>("hash/" + hashtype);
+
```
attilamolnar: Remove the blank lines please

src/modules/m_sqloper.cpp
flopraden: I think I've done that because if the HashProvider is not found, I assume that hashedpass is in fact the plaintext password.
One way to do it, is to create a NoHash hash algorithm.

src/modules/m_sqloper.cpp
flopraden: I did not understand why hashtype should be a const ref and not uid, username or password.

SaberUK:
I think this might be indirectly implemented via #983 (which will be merged as soon I have had time to audit it). Any thoughts @B00mX0r?

B00mX0r:
@SaberUK you're correct. Part of the point of #983 is to reduce duplication of oper code by making authentication all run through the same function. That way, if the logic for /OPER changes in the future, sqloper will already support it.

SaberUK:
As this duplicates behaviour in an older pull request i'm going to close this.

Apologies for the delay in responding and thanks for your contribution though! 👍

Issue #989, opened 2015-02-07
[2.0] CGIIRC: Lines don't work if real IP is E-Lined


B00mX0r:
When G-lining and Z-lining a host connected via WEBIRC, it won't do anything if the router's IP is E-lined. Example: 1.1.1.1 sends a WEBIRC for user "User1", telling the server that this user's IP and host mask are 1.1.1.2. If 1.1.1.1 is E-lined, then lines placed on 1.1.1.2 won't work. So if I Z-line 1.1.1.2, it won't actually work. It works fine if 1.1.1.1 isn't E-lined. This behavior also applies to DNSBL, where DNSBL won't check 1.1.1.2 to see if it's blacklisted.

Version: Latest insp20 code (2.0.19 I guess). Untested on other versions (it may also apply to 2.2).

B00mX0r:
If I'm right, then the code sets an exempt flag on a connection if it detects an e-line. Since the user originally connects via the real IP, I think that it's setting the exempt flag but then doesn't unset it on webirc or cgiirc.

JDowny:
I can't replicate this...

<code>* *** ANNOUNCEMENT: Connecting user 490AAAADG detected as using CGI:IRC (127.0.0.1), changing real host to 192.168.0.100 from 127.0.0.1
* *** CONNECT: Client connecting on port 6697 (class mainsubs): asdfafsdasdfasfd!webirc@192.168.0.100 (192.168.0.100) [http://testnet.com]
* *** XLINE: TinMan added permanent E-line for *@127.0.0.1: test
* *** XLINE: TinMan added timed Z-line for 192.168.0.100, expires on Sun Feb  8 01:36:35 2015: test
* *** QUIT: Client exiting: asdfafsdasdfasfd!webirc@192.168.0.100 (192.168.0.100) [Z-Lined: test]</code>

Or am I misunderstanding this?

B00mX0r:
You're unable to replicate this because you add the E-line after the client is connected. The E-line must be added before, because this is what happens:

1) Client connects with 127.0.0.1
2) user->exempt status is set since 127.0.0.1 is E-lined
3) WEBIRC, ok, IP for client now 192.168.0.100
4) We check for lines on 192.168.0.100, but return false since the user's exempt flag is set to true

JDowny:
`* *@127.0.0.1 1423384571 0 TinMan :test
* *@192.168.0.117 1423384289 0 TinMan :test
* e :End of /STATS report
[...]

PR #990, opened 2015-02-08
Update user exempt flag on WEBIRC

B00mX0r:
Solves https://github.com/inspircd/inspircd/issues/989

Simple example: If Mibbit's IP is E-lined, and a user connects through Mibbit (so using WEBIRC), it won't check for bans on the user.

### DIFF
--- a/src/modules/m_cgiirc.cpp
+++ b/src/modules/m_cgiirc.cpp
@@ -144,6 +144,7 @@ class CGIResolver : public Resolver
 			them->host = result;
 			them->dhost = result;
 			them->InvalidateCache();
+			them->exempt = (ServerInstance->XLines->MatchesLine("E",them) != NULL);
 			them->CheckLines(true);
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/modules/m_cgiirc.cpp
```diff
@@ -317,6 +319,7 @@ class ModuleCgiIRC : public Module
 			if(InspIRCd::Match(user->host, iter->hostmask, ascii_case_insensitive_map) || InspIRCd::MatchCIDR(user->GetIPString(), iter->hostmask, ascii_case_insensitive_map))
 			{
 				// Deal with it...
+				user->exempt = (ServerInstance->XLines->MatchesLine("E",user) != NULL);
```
attilamolnar: This is in the wrong place, at the time this runs the user IP etc. hasn't been changed yet

attilamolnar:
Not merging as we've decided to fix the regression this has attempted to fix by reverting the commit after 2.0.18 which caused the regression.

PR #992, opened 2015-02-12
[2.0] Fix erroneous error messages in configure.

SaberUK:
Another day; another misleading error message from configure.


attilamolnar:
Thanks, looks like this was forgotten when moving from gnutls-config to pkg-config.

PR #999, opened 2015-02-21
m_sqlauth: Allow specifying what hash algos to provide hashes from

### LINKED ISSUE #741: bcrypt support in m_sqlauth
Hi

We were using m_sqlauth with our user database that stored passwords using md5. 

We converted these to bcrypt assuming InspIRCd already supported it - no. 

So at the moment we've had to open all connections. 

Can bcrypt support be added?

Thanks

Steve

ShutterQuick:
Defaults to md5,sha256 to be compatible with old configs.
Closes #741

### DIFF
--- a/src/modules/m_sqlauth.cpp
+++ b/src/modules/m_sqlauth.cpp
@@ -34,18 +34,58 @@ class AuthQuery : public SQLQuery
 	const std::string uid;
 	LocalIntExt& pendingExt;
 	bool verbose;
-	AuthQuery(Module* me, const std::string& u, LocalIntExt& e, bool v)
-		: SQLQuery(me), uid(u), pendingExt(e), verbose(v)
+	const std::string& kdf;
+	const std::string& pwcolumn;
+
+	AuthQuery(Module* me, const std::string& u, LocalIntExt& e, bool v, const std::string& kd, const std::string& pwcol)
+		: SQLQuery(me), uid(u), pendingExt(e), verbose(v), kdf(kd), pwcolumn(pwcol)
 	{
 	}
 
 	void OnResult(SQLResult& res) CXX11_OVERRIDE
 	{
-		User* user = ServerInstance->FindNick(uid);
+		LocalUser* user = static_cast<LocalUser*>(ServerInstance->FindUUID(uid));
 		if (!user)
 			return;
+
 		if (res.Rows())
 		{
+			if (!kdf.empty())
+			{
+				HashProvider* hashprov = ServerInstance->Modules->FindDataService<HashProvider>("hash/" + kdf);
+				if (!hashprov)
+				{
+					if (verbose)
+						ServerInstance->SNO->WriteGlobalSno('a', "Forbidden connection from %s (a provider for %s was not loaded)", user->GetFullRealHost().c_str(), kdf.c_str());
+					pendingExt.set(user, AUTH_STATE_FAIL);
+					return;
+				}
+
+				size_t colindex = 0;
+				if (!pwcolumn.empty() && !res.HasColumn(pwcolumn, colindex))
+				{
+					if (verbose)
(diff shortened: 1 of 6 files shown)


src/modules/m_sqlauth.cpp
```diff
@@ -34,22 +34,41 @@ class AuthQuery : public SQLQuery
...
+	const std::string kdf;
+	AuthQuery(Module* me, const std::string& u, LocalIntExt& e, bool v, const std::string& kd)
+		: SQLQuery(me), uid(u), pendingExt(e), verbose(v), kdf(kd)
 	{
 	}
 
 	void OnResult(SQLResult& res) CXX11_OVERRIDE
 	{
-		User* user = ServerInstance->FindNick(uid);
+		LocalUser* user = dynamic_cast<LocalUser*>(ServerInstance->FindNick(uid));
```
attilamolnar: No dynamic_cast in release builds since b705c6426818ba8cedfd00d1a84dab8ff0d0f1a0

src/modules/m_sqlauth.cpp
```diff
@@ -34,22 +34,41 @@ class AuthQuery : public SQLQuery
...
 		if (!user)
 			return;
+
 		if (res.Rows())
 		{
+			if (!kdf.empty())
+			{
+				HashProvider* hashprov = ServerInstance->Modules->FindDataService<HashProvider>("hash/" + kdf);
+				if (!hashprov)
+					goto authfailure;
```
attilamolnar: I don't like this goto, and it also jumps to log the wrong thing ("SQL query returned no matches")

src/modules/m_sqlauth.cpp
```diff
@@ -34,22 +34,41 @@ class AuthQuery : public SQLQuery
 	const std::string uid;
 	LocalIntExt& pendingExt;
 	bool verbose;
-	AuthQuery(Module* me, const std::string& u, LocalIntExt& e, bool v)
-		: SQLQuery(me), uid(u), pendingExt(e), verbose(v)
+	const std::string kdf;
```
attilamolnar: Why does it copy the string here?

src/modules/m_sqlauth.cpp
```diff
@@ -34,18 +34,44 @@ class AuthQuery : public SQLQuery
...
+	const std::string& kdf;
+	AuthQuery(Module* me, const std::string& u, LocalIntExt& e, bool v, const std::string& kd)
+		: SQLQuery(me), uid(u), pendingExt(e), verbose(v), kdf(kd)
 	{
 	}
 
 	void OnResult(SQLResult& res) CXX11_OVERRIDE
 	{
-		User* user = ServerInstance->FindNick(uid);
+		LocalUser* user = IS_LOCAL(ServerInstance->FindUUID(uid));
```
attilamolnar: Crashes here if the user quits, static_cast instead

src/modules/m_sqlauth.cpp
```diff
@@ -34,18 +34,44 @@ class AuthQuery : public SQLQuery
...
+						ServerInstance->SNO->WriteGlobalSno('a', "Forbidden connection from %s (a provider for %s was not loaded)", user->GetFullRealHost().c_str(), kdf.c_str());
+					pendingExt.set(user, AUTH_STATE_FAIL);
+					return;
+				}
+
+				SQLEntries row;
+				while (res.GetRow(row))
+				{
+					if (hashprov->Compare(user->password, row[0].value))
+						return (void)pendingExt.set(user, AUTH_STATE_NONE);
```
attilamolnar: Please put the set and and the return on separate lines

SaberUK:
This has been committed in d38f2ad213e875d40556116cba3230064eacb861  and 066cc66cc541b612a5293093fe1f070b24c68768

PR #1000, opened 2015-02-21
m_sqloper: Add `active` column

### LINKED ISSUE #281: Add an "active" flag to SQL Oper
Hi,

I think it would be a nice idea to have the possibility to block an oper from logging in by having an "active" flag in the SQL Schema. For example if it is "0" the account is blocked and if the value is "1" the oper will be able to login.

ShutterQuick:
closes #281


attilamolnar:
Great, thanks. Please update the wiki as well.

PR #1005, opened 2015-02-28
Various changes mainly relating to the build system.

SaberUK:
ec971cb72c3e9b1f1646a4c76ba1ddaf7c7221e5 will probably break Jenkins.

@tobgen can you change the Jenkinses to pass `--development` to configure?

### DIFF
--- a/configure
+++ b/configure
@@ -272,10 +272,10 @@ if (prompt_bool $interactive, $question, 0) {
 	foreach my $key (qw(BINARY_DIR CONFIG_DIR DATA_DIR LOG_DIR MANUAL_DIR MODULE_DIR)) {
 		$config{$key} =~ s/^\Q$original_base_dir\E/$config{BASE_DIR}/;
 	}
-	$config{BINARY_DIR} = prompt_dir $interactive, 'In what directory should the InspIRCd binary be placed?',     $config{BINARY_DIR};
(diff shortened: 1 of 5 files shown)


md-5:
Seriously seems like using a off the shelf buildsystem is worth considering given how often this one changes

SaberUK:
Even if you used CMake or some other build system we would still need to change it to add features like this.

md-5:
@SaberUK why can't they? http://www.cmake.org/download/

SaberUK:
Good luck explaining how to install CMake to someone who has problems typing `./configure` and `make install`. Also, CMake doesn't offer an interactive configuration system so we would have to maintain the existing one as a wrapper for CMake.

Besides, i'm fine with maintaining the build system.

Zayelion:
Developer vs user mindset. Why most people dont use Linux, and there are installers for most things. Build system 'just makes it work'.

attilamolnar:
> Seriously seems like using a off the shelf buildsystem is worth considering given how often this one changes

The reason why inspircd has its own build system is historical. If the project was started today it would have more dependencies and less custom code in general for sure.

However:

> Even if we used CMake or some other build system we would still need to change it to add features like this.

CMake does not work magically either and has its own problems. @SaberUK is naturally more familiar with their own code than any other build system, and since they're the one who does the lion's share of work on the build system (even if we had another one) ultimately it's their choice.

@SaberUK Thanks!

PR #1008, opened 2015-03-03
m_clones: Show a list over nicks under each CIDR

ShutterQuick:
closes #736

### DIFF
--- a/src/modules/m_clones.cpp
+++ b/src/modules/m_clones.cpp
@@ -21,19 +21,111 @@
 
 #include "inspircd.h"
 
+typedef std::vector<User*> UList;
+typedef std::map<irc::sockets::cidr_mask, UList> CloneList;
+
+class UserTracker : public Timer
+{
+	bool tracking;
+	unsigned long timeout;
+	CloneList cl;
+
+public:
+	UserTracker()
+		: Timer(timeout, false), tracking(false), timeout(0)
+	{
+	}
+
+	bool Tick(time_t) CXX11_OVERRIDE
+	{
+		StopTracking();
+		return false;
+	}
+
+	void SetTimeout(unsigned long t)
+	{
+		timeout = t;
+	}
+
+	const CloneList& GetMap()
+	{
+		if (tracking)
+		{
+			SetInterval(timeout);
+			return cl;
+		}
+
+		if (timeout)
+		{
+			tracking = true;
+			SetInterval(timeout);
+		}
+
+		for (user_hash::const_iterator it = ServerInstance->Users->clientlist.begin(); it != ServerInstance->Users->clientlist.end(); ++it)
+		{
+			User* user = it->second;
+			AddClient(user);
+		}
+		return cl;
+	}
+
+	void StopTracking()
+	{
+		cl.clear();
+		tracking = false;
+	}
+
+	void AddClient(User* user)
+	{
+		if (!tracking)
+			return;
+
+		UList& counts = cl[user->GetCIDRMask()];
+		counts.push_back(user);
+	}
+
+	void DelClient(User* user)
+	{
+		if (!tracking)
+			return;
+
+		CloneList::iterator i = cl.find(user->GetCIDRMask());
+		if (i == cl.end())
+		{
+			ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT, "Potential bug. CIDR was not found in clone map upon removal.");
+			StopTracking();
+			return;
+		}
+
+		UList& counts = i->second;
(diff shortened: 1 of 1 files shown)


src/modules/m_clones.cpp
```diff
@@ -64,10 +151,23 @@ class CommandClones : public Command
...
 	CommandClones cmd;
+
  public:
-	ModuleClones() : cmd(this)
+	ModuleClones() : cmd(this, utrack)
+	{
+		utrack.SetTimeout(60*5);
+	}
+
+	void OnPostConnect(User*user) CXX11_OVERRIDE
```
attilamolnar: Missing space between type and var name

src/modules/m_clones.cpp
```diff
@@ -23,16 +23,97 @@
...
+
+		if (timeout)
+		{
+			tracking = true;
+			SetInterval(timeout);
+		}
+
+		for (user_hash::const_iterator it = ServerInstance->Users->clientlist.begin(); it != ServerInstance->Users->clientlist.end(); ++it)
+		{
+			User* user = &(*it->second);
```
attilamolnar: `it->second`

src/modules/m_clones.cpp
```diff
@@ -23,16 +23,97 @@
 
 /** Handle /CLONES
  */
```
attilamolnar: This comment belongs to the CommandClones class, should put the new code above it

src/modules/m_clones.cpp
```diff
@@ -23,16 +23,97 @@
...
+		tracking = false;
+		return false;
+	}
+
+	void SetTimeout(unsigned long t)
+	{
+		timeout = t;
+	}
+
+	CloneList& GetMap()
```
attilamolnar: Couldn't it return a const ref?

src/modules/m_clones.cpp
```diff
@@ -23,16 +23,97 @@
...
- 	CommandClones(Module* Creator) : Command(Creator,"CLONES", 1)
+	CommandClones(Module* Creator, UserTracker& ut)
+		: Command(Creator,"CLONES", 1), utrack(ut)
 	{
 		flags_needed = 'o'; syntax = "<limit>";
 	}
 
 	CmdResult Handle (const std::vector<std::string> &parameters, User *user)
 	{
+		CloneList& clonemap = utrack.GetMap();
```
attilamolnar: Move this near first use (that is, before the loop like the current code does with the clonemap)

src/modules/m_clones.cpp
```diff
@@ -23,16 +23,97 @@
 
 /** Handle /CLONES
  */
+
+
+typedef TR1NS::unordered_set<User*> UList;
+typedef std::map<irc::sockets::cidr_mask, UList > CloneList;
```
attilamolnar: Extra space after UList and `>`

src/modules/m_clones.cpp
```diff
@@ -21,19 +21,112 @@
...
+	void DelClient(User* user)
+	{
+		if (!tracking)
+			return;
+
+		CloneList::iterator i = cl.find(user->GetCIDRMask());
+		if (i == cl.end())
+		{
+			ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT, "Potential bug. CIDR was not found in clone map upon removal. Did you change clone CIDR?");
+			return StopTracking();
```
attilamolnar: StopTracking();
    return;

src/modules/m_clones.cpp
```diff
@@ -21,19 +21,112 @@
...
+		StopTracking();
+		return false;
+	}
+
+	void SetTimeout(unsigned long t)
+	{
+		timeout = t;
+	}
+
+	const CloneList& GetMap()
```
attilamolnar: Method now can be const

src/modules/m_clones.cpp
```diff
@@ -21,19 +21,112 @@
...
+
+		UList& counts = i->second;
+		UList::iterator it = std::find(counts.begin(), counts.end(), user);
+		if (it == counts.end())
+		{
+			ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT, "Potential bug. User was not found in clone map upon removal. Did you change clone CIDR?");
+			return StopTracking();
+		}
+
+		stdalgo::vector::swaperase(counts, it);
```
attilamolnar: The above if should check the return of this too, if it returns false the user was not found and the same should happen

SadieCat:
The clones module was moved to -contrib for v4.

PR #1009, opened 2015-03-09
[2.0] Fix building when VERBOSE is already set in the environment.

SaberUK:
This is a workaround for now. In 2.2 we should prefix all of the environment variables with INSPIRCD_ to avoid collisions.


attilamolnar:
Shouldn't this be fixed properly in 2.0?

SaberUK:
I could do that but it might cause merge conflicts. If you're okay with that then I can change it.

SaberUK:
I could do that but it might cause merge conflicts. If you're okay with that then I can change it.

attilamolnar:
This is fine, thanks.

PR #1012, opened 2015-03-14
Add CPPFLAGS support

guikcd:
No default value set, but let users define them (ex.: -D_FORTIFY_SOURCE=2)


SaberUK:
There's no need to edit the Perl files. You can append `CPPFLAGS` to `CORECXXFLAGS` at the same time that `CXXFLAGS` is appended.

guikcd:
You're absolutely right; i simply found more clear and standard to have a dedicated variable (and more useful for other distributions vendors if needed).

As your project choose to have a custom build system, i think you can consider reproducing standard behaviour like autotools.

Anyway, if you don't choose to support it, i'll remove this patch from Debian's package and append CPPFLAGS to CFLAGS in next releases.

SaberUK:
CORECXXFLAGS is an internal Makefile variable used to build the flags for `unit-cc` to pass to the compiler. See what we do on line 133 of `main.mk`.

attilamolnar:
@SaberUK has addressed this in #1225.

guikcd:
:+1:

PR #1021, opened 2015-03-23
Convert build system to Perl 5.10.

SaberUK:
Now we have dropped support for the old CentOS release we can use Perl 5.10 instead of 5.8. Yay!

**Note:** I'm not going to update `make/utilities.pm` as that file is scheduled to be removed soon anyway. Similarly, `./modulemanager` and the module bits of `./configure` are due to be entirely rewritten so i'm not going to update that either.

### DIFF
--- a/configure
+++ b/configure
@@ -27,9 +27,10 @@
 
 
 BEGIN {
-	require 5.8.0;
+	require 5.10.0;
 }
 
+use feature ':5.10';
 use strict;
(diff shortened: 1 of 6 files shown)


SaberUK:
@attilamolnar This should be ready to merge now.

attilamolnar:
Thanks, can you rephrase "SSL generation" in the future?

PR #1024, opened 2015-03-28
Start cleaning up the make/*.pl files.

SaberUK:


SaberUK:
This should be done now. There is still some problems but i'll tackle them at a later date.

attilamolnar:
Thanks!

Issue #1023, opened 2015-03-26
[Proposal] Add a new extban to m_gecosban to match n!u@h+gecos


Renegade334:
Currently, `m_gecosban` exports the `r:` extban for globmatching against gecos. However, there is no way to fine-tune this matching behaviour by mask. The ability to narrow down a gecos extban by mask might be useful in circumstances where the target gecos pattern is a common one, and therefore this crude matching against all users is not desirable.

One way this could be implemented is with an additional extban in `m_gecosban` that takes a string containing both n!u@h and a gecos globmask. For example, `a:*!*@*.abusive.host+*bot*` might match users matching both the `*!*@*.abusive.host` mask and the `*bot*` gecos pattern.

Thoughts welcome!

Renegade334:
Example implementation at https://github.com/inspircd/inspircd/compare/master...Renegade334:ext-gecosban.

SaberUK:
:+1:

You should probably file a pull request.

maxteufel:
Can't this be consistent with Charybdis' `$x:nick!user@host#realname` (at least the # instead of +)?

SaberUK:
@maxteufel I imagine that might cause conflicts with `m_banredirect`.

PR #1026, opened 2015-04-02
Add extban 'a' to m_gecosban

Renegade334:
`a:nick!user@host+gecos`
eg. `a:*!*@abusive.host+*MegaBot?v2.0*`

ref: #1023


SaberUK:
Hello, can you rebase this please. 👍

attilamolnar:
I would prefer this to be implemented in a more generic way that does not require new extban characters for every combination of matchers (realname + hostmask in this case).

PR #1030, opened 2015-04-07
Tell git to store .bat files as CRLF

docwhat:
You can also add additional types, such as:

```
*.cpp  text diff=cpp eol=lf
*.h    text diff=cpp eol=lf
```

Ciao!


attilamolnar:
Why is this necessary? Can't Windows execute the .bat file as it is currently?

docwhat:
Sure. But why not use the right line endings? That would let someone open the file in notepad on Windows.

SaberUK:
You probably shouldn't be using notepad for editing code files.

docwhat:
I was thinking just to view it.

attilamolnar:
Thanks!

PR #1034, opened 2015-04-16
fix whitespace host

Sheogorath-SI:
someone may can test it

### DIFF
--- a/src/user_resolver.cpp
+++ b/src/user_resolver.cpp
@@ -92,7 +92,7 @@ void UserResolver::OnLookupComplete(const std::string &result, unsigned int ttl,
 		if (rev_match)
 		{
 			std::string hostname = bound_user->stored_host;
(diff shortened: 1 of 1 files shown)


src/user_resolver.cpp
```diff
@@ -91,7 +92,7 @@ void UserResolver::OnLookupComplete(const std::string &result, unsigned int ttl,
 		if (rev_match)
 		{
 			std::string hostname = bound_user->stored_host;
-			if (hostname.length() < 65)
+			if (hostname.length() < 65 and !hostname.find(" "))
```
Heufneutje: &&, not and.

Sheogorath-SI: oh man... to many scripting thanks

Sheogorath-SI:
not needed anymore

PR #1035, opened 2015-04-16
fix whitespace host

Sheogorath-SI:
(again)

### DIFF
--- a/src/user_resolver.cpp
+++ b/src/user_resolver.cpp
@@ -1,6 +1,7 @@
 /*
  * InspIRCd -- Internet Relay Chat Daemon
  *
+ *   Copyright (C) 2015 Christoph Kern <christoph.kern@shivering-isles.com>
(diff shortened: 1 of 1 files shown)


src/user_resolver.cpp
```diff
@@ -91,7 +92,7 @@ void UserResolver::OnLookupComplete(const std::string &result, unsigned int ttl,
 		if (rev_match)
 		{
 			std::string hostname = bound_user->stored_host;
-			if (hostname.length() < 65)
+			if (hostname.length() < 65 && hostname.find_first_not_of("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-.") == std::string::npos)
```
Heufneutje: What about numeric?

attilamolnar:
Not merging as we've decided to fix this differently: validate the host in the DNS resolver so the issue cannot happen in a different part of the codebase. Thanks for your quick reaction and initial work on the fix.

PR #1037, opened 2015-04-26
Implement support for IRCv3.2 (part 1)

SaberUK:
Exactly what it says on the tin.

See #1038 for the implementation status.

### DIFF
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -963,7 +963,7 @@
 # The following block can be used to control which extensions are
 # enabled. Note that extended-join can be incompatible with delayjoin
 # and host cycling.
-#<ircv3 accountnotify="on" awaynotify="on" extendedjoin="on">
+#<ircv3 accountnotify="on" awaynotify="on" echomessage="on" extendedjoin="on">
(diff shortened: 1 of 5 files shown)


SaberUK:
@attilamolnar Can you take a look at this and merge it if you think its ok? I'll submit the other IRCv3.2 stuff in a second pull at a future date when I get around to writing it.

attilamolnar:
It's not ideal that host cycling and the chghost extension can only be enabled together. Many dislike host cycling and don't use it but that shouldn't affect the chghost extension.
Any reason you put the new caps that don't require any code from m_ircv3 into m_ircv3 instead of a separate module each?
For cap-notify: it shouldn't be implemented like this because the m_cap module and the header which is heavily used by this cap-notify implementation is undergoing a major rewrite to be able to support the new IRCv3 stuff and to be more flexible in general (e.g. support per-user cap values, make it possible to disallow turning off sasl for registered users, etc.).

attilamolnar:
It's not worth implementing cap-notify with the current cap code we have, I'll implement cap-notify along with all the new cap stuff.

PR #1039, opened 2015-04-28
[2.0] Fix not being able to message ulines when m_restrictmsg is loaded.

SaberUK:
As requested by @Adam-.

---

Whilst this is a change in behaviour I believe that it was caused by an oversight. Generally people do want their users to be able to message services even if they have messaging restricted.


attilamolnar:
Thanks!

PR #1043, opened 2015-05-12
[2.0] Fix configure overwriting the cache settings on startup.

SaberUK:
This means that the settings the admin has previously configured with are no longer ignored when upgrading.

Reported by @Cronus89 on IRC.

---

The 2.0 build system really is the gift that keeps on giving. None of this needs to be merged into master as the code there has already been completely rewritten and works like it should.


attilamolnar:
Thanks!

Issue #1036, opened 2015-04-17
Multiple-character fantasy prefixes supported in m_alias


MandL27:
Could it be possible to set prefixes in a <fantasy> tag to be longer than just a single character? According to SaberUK, the current build makes a prefix definition of "!s@me7h1nG" get interpreted as simply "!".

attilamolnar:
@MandL27 @SaberUK Can you provide use cases for this feature request?

MandL27:
@attilamolnar I've seen a few bots that respond to "!(two-letter_prefix) (command)" so if fantasy prefixes built into insp could behave similarly, it'd keep one namespace across the network.

attilamolnar:
Implemented by @SaberUK in #1044.

PR #1044, opened 2015-05-13
Add support for channel alias prefixes with multiple characters.

SaberUK:
As requested in #1036.

### DIFF
--- a/src/modules/m_alias.cpp
+++ b/src/modules/m_alias.cpp
@@ -57,7 +57,7 @@ class Alias
 
 class ModuleAlias : public Module
 {
-	char fprefix;
+	std::string fprefix;
 
 	/* We cant use a map, there may be multiple aliases with the same name.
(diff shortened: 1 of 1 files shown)


attilamolnar:
Is this OK to merge now?

src/modules/m_alias.cpp
```diff
@@ -193,19 +193,19 @@ class ModuleAlias : public Module
...
 
-		if (scommand.empty())
+		if (scommand.size() <= fprefix.size())
 		{
 			return; // wtfbbq
 		}
 
 		// we don't want to touch non-fantasy stuff
-		if (*scommand.c_str() != fprefix)
+		if (scommand.compare(0, fprefix.size()-1, fprefix) == 0)
```
attilamolnar: Second parameter should be `fprefix.size()`, as it's a length.

attilamolnar:
Thanks!

PR #1046, opened 2015-05-15
Implement support for callers having multiple handlers.

SaberUK:
This fixes the problem of having multiple modules which hook the same event colliding with each other which could potentially cause a crash if a rememberer stores a handler from another module which is unloaded first.

It does not fix the problem of the module which gets the active hook being entirely based on load (or rather `init`) order.


SaberUK:
Closing this as it seems like there is no interest in it.

PR #1048, opened 2015-05-17
Encase the IPv6 address in `STATS p` with square brackets.

SaberUK:
This is a [widely used format](http://en.wikipedia.org/wiki/IPv6_address#Literal_IPv6_addresses_in_network_resource_identifiers) as without it the port is ambiguous.

Before:

    :irc.example.com 249 NickName ::::6667 (clients, plaintext)

After:

    :irc.example.com 249 NickName :[::]:6667 (clients, plaintext)

We also use this format in `irc::sockets::sockaddrs::str()`.


attilamolnar:
I'm ok with this however the current form isn't ambiguous because the port is always the last bit in the text we send

Issue #1022, opened 2015-03-24
[Proposal] Strip colour from quit messages in cmd_quit


Renegade334:
Some ircd's such as charybdis strip colour from user-provided QUIT messages as core behaviour. There are several conceivable arguments in favour of this:
- A quitting user is not a particularly noteworthy event, and it should not draw attention away from surrounding channel activity.
- Coloured quit messages will break through user modes/channel modes/extbans that block or strip colour messages when they are displayed within an active channel or query by the client, which is an annoyance to the user.
- Quit messages cannot be modified by event handlers within InspIRCd. To implement any kind of filtering of quit messages would require direct changes to the core `cmd_quit` module.

As such, I'd be interested to see what feelings are with regards implementing this behaviour.

SaberUK:
>Quit messages cannot be modified by event handlers within InspIRCd.

There's no reason we couldn't add a prehook in UserManager::QuitUser or make the message parameter of `OnUserQuit` non-const.

>As such, I'd be interested to see what feelings are with regards implementing this behaviour.

:+1: as a module. :-1: as a core feature.

Shawn-Smith:
I think it should be stripped if the quitting user is in any channel that either blocks or strips color for it's users, that should be part of the block/strip color module though.

A new module or option could be made for stripping it no matter what.

SaberUK:
@Shawn-Smith Doing it anything other than globally seems like it would be confusing. Also, we don't have the hooks for doing variable quit messages and it would probably be too intensive to add.

PR #1053, opened 2015-05-27
Added m_nocolorquit module to implement the functionality talked about in bug #1022

Justasic:
I am not sure the config values are on par with what it really should be. Please critique :D

The module itself adds the ability to remove the message entirely from both QUIT and PART commands as well as it adds the ability to strip the color codes from the message but still allow the message to be displayed.

The module will also allow opers to override this functionality and notify the user when their PART messages are being blocked or stripped.

### DIFF
--- a/src/modules/m_nocolorquit.cpp
+++ b/src/modules/m_nocolorquit.cpp
@@ -1,0 +1,144 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2009 Daniel De Graaf <danieldg@inspircd.org>
+ *   Copyright (C) 2008 Pippijn van Steenhoven <pip88nl@gmail.com>
+ *   Copyright (C) 2007 Dennis Friis <peavey@inspircd.org>
+ *   Copyright (C) 2007 Robin Burchell <robin+git@viroteck.net>
+ *   Copyright (C) 2006-2007 Oliver Lupton <oliverlupton@gmail.com>
+ *   Copyright (C) 2015 Justin Thomas Crawford <Justasic@gmail.com>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "inspircd.h"
+
+/* $ModDesc: Strip or entirely block color coded quit and part messages. */
+
+class ModuleBlockColoredQuit : public Module
+{
+
+	// What do we block?
+	bool BlockColoredQuit;
+	bool StripColoredQuit;
+	bool BlockColoredPart;
+	bool StripColoredPart;
+
+	// Module options
+	bool NotifyUser;
+	bool AffectOpers;
+
+	bool HasColors(const std::string &text)
+	{
+			// borrowed from m_blockcolor.so
+		for (std::string::const_iterator i = text.begin(); i != text.end(); i++)
+		{
+			switch (*i)
+			{
+				case 2:
+				case 3:
+				case 15:
+				case 21:
+				case 22:
+				case 31:
+					return true;
+				break;
+			}
+		}
+		return false;
+	}
+
+public:
+	ModuleBlockColoredQuit()
+	{
+	}
+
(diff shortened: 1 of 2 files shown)


Justasic:
Github doesn't tag bugs like I remembered it so here's the link to that bug: https://github.com/inspircd/inspircd/issues/1022

Heufneutje:
It does if you put the number in the commit message as you already did. ;)

Also, I'm not too sure about the blocking part of this. Blocking the PART messages could trick clients that just assume they've left the channel after sending PART into thinking they have left when they haven't actually left (you could argue that's a badly written client, but sadly there's an alarming number of those). Blocking QUIT messages isn't too big of a deal as it would just disconnect them with a generic "connection closed", but I'm not sure if that's desirable either.

SaberUK:
This should be filed against [inspircd-extras](https://github.com/inspircd/inspircd-extras/).

src/modules/m_nocolorquit.cpp
```diff
@@ -0,0 +1,144 @@
...
+				case 31:
+					return true;
+				break;
+			}
+		}
+		return false;
+	}
+
+public:
+	ModuleBlockColoredQuit()
```
SaberUK: Empty ctors and dtors can be omitted.

src/modules/m_nocolorquit.cpp
```diff
@@ -0,0 +1,144 @@
...
+	{
+	}
+
+	virtual ~ModuleBlockColoredQuit()
+	{
+	}
+
+	virtual Version GetVersion()
+	{
+		return Version("Strip or entirely block color coded quit and part messages.",VF_VENDOR);
```
SaberUK: This isn't a `VF_VENDOR` module.

src/modules/m_nocolorquit.cpp
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	virtual void init()
+	{
+		Implementation eventlist[] = { I_OnPreCommand, I_OnRehash };
+        ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist) / sizeof(Implementation));
```
SaberUK: Indentation is broken here.

src/modules/m_nocolorquit.cpp
```diff
@@ -0,0 +1,144 @@
...
+		this->OnRehash(NULL);
+	}
+
+ 	virtual void OnRehash(User* user)
+    {
+        ConfigTag* tag = ServerInstance->Config->ConfValue("nocoloredquit");
```
SaberUK: Broken here too.

src/modules/m_nocolorquit.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+		this->BlockColoredQuit = tag->getBool("blockcoloredquitmessages", true);
+		this->StripColoredQuit = tag->getBool("stripcoloredquitmessages", false);
```
SaberUK: False is the default value for `getBool` so you don't need to specify it.

src/modules/m_nocolorquit.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    virtual ModResult OnPreCommand(std::string &command, std::vector<std::string> &parameters, LocalUser *user, bool validated, const std::string &original_line)
+	{
+		if (command == "QUIT")
+		{
+			// If AffectOpers option is enabled, ignore oper quit messages entirely.
+			if (IS_OPER(user) && !this->AffectOpers)
+				return MOD_RES_PASSTHRU;
+
+			bool hascolor = HasColors(parameters[0]);
```
SaberUK: What happens if the `parameters` array is empty?

src/modules/m_nocolorquit.cpp
```diff
@@ -0,0 +1,144 @@
...
+		this->BlockColoredPart = tag->getBool("blockcoloredpartmessages", false);
+		this->StripColoredPart = tag->getBool("stripcoloredpartmessages", false);
+
+		this->NotifyUser = tag->getBool("notifyusers", false);
+		this->AffectOpers = tag->getBool("affectopers", false);
+    }	
+
+    virtual ModResult OnPreCommand(std::string &command, std::vector<std::string> &parameters, LocalUser *user, bool validated, const std::string &original_line)
+	{
+		if (command == "QUIT")
```
SaberUK: You should probably check if the user is REG_ALL here.

attilamolnar:
2.0 is bugfix only, please submit this to inspircd-extras.

src/modules/m_nocolorquit.cpp
Justasic: if user is REG_ALL? this is something I am unfamiliar with.

src/modules/m_nocolorquit.cpp
Justasic: In testing, inspircd inserted its own default quit message which was why it didn't crash for me (rather, I assumed inspircd would always insert such message in the case that `QUIT` is sent without any parameters).

I wasn't sure that there would ever be a sane condition where inspircd would call the module without a parameter. I'll add the check anyway.

PR #1054, opened 2015-06-03
Fix modules with 2+ files not having their LinkerFlags parsed.

SaberUK:
This problem was discovered by @Adam-.


attilamolnar:
Thanks!

PR #1056, opened 2015-06-07
modules.conf.example: use full cloaking by default

Mikaela:
This was discussed on \#inspircd and \#ircv3 and @SaberUK suggested me to PR about this.

### DIFF
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -497,7 +497,7 @@
 # An optional prefix may be specified to mark cloaked hosts.          #
 #-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#
 #
-#<cloak mode="half"
+#<cloak mode="full"
 #       key="secret"
 #       prefix="net-">
 


GLolol:
I prefer half cloaking because in the case of spam and abuse, it at *least* lets the ops of a channel know roughly what IP ranges are being a nuisance.

Techman-:
You're right, it half cloaking does benefit channel operators. However, full cloaking benefits the users as a whole more. More of a network decision but perhaps this trade off should be specified above this default as well?

Mikaela:
I think people should understand the explanations of both modes and what resulted to this PR on \#ircv3 is:

```
2015-06-07 18:48:56+0300 < kaniini> SaberUK: for what its worth i am not likely to release that tool
2015-06-07 18:49:36+0300 < kaniini> SaberUK: i just hacked it up because i sit on an irc channel for lulz which has incompetent moderators (so i started DNSBL-checking what IPs I could myself to prove these people are just there to troll/crapflood/etc.)
2015-06-07 18:51:00+0300 < Mikaela> if you are referring to inspircd, it was said that many people had done it before you and someone linked to https://github.com/maugier/inspircd-decloak
2015-06-07 18:51:51+0300 < kaniini> mine does not use cloak keys, it instead guesses based on training data
2015-06-07 18:52:25+0300  * Mikaela just also learned that that is unrelated after opening the README
2015-06-07 18:52:28+0300 < kaniini> the anope 2.x method in decal's presentation was useful for feeding said training data :)
2015-06-07 18:53:06+0300 < kaniini> it could probably be improved to do "full" cloaking instead of just half, but the network in question uses half and open proxies rarely have valid rDNS
2015-06-07 18:53:08+0300 < kaniini> ;)
2015-06-07 18:55:57+0300 < Mikaela> I wonder how many networks actually use full or are there any statistics on it, I think people think full as ugly and don't use it or something
2015-06-07 18:57:01+0300 < Mikaela> alternatively half being "default" or "suggested" or "recommended" on https://github.com/inspircd/inspircd/blob/v2.0.20/docs/conf/modules.conf.example#L520
2015-06-07 19:00:14+0300 < SaberUK> Mikaela: feel free to file a pr to change it to full
```

The tool referred by @kaniini was not specified on \#ircv3, but it's a script bruteforcing the real address through cloaking.

attilamolnar:
@GLolol :+1:, I believe half is a good tradeoff because it stops script kiddies and allows more efficient moderation, which is why I don't want to change the default. We should rather update the docs telling admins that the cloaks generated this way should not be relied on.

PR #1057, opened 2015-06-07
Add support for length arguments in getString.

SaberUK:
This mirrors the min/max arguments on getInt and getFloat.


attilamolnar:
I wonder if the default behavior should be to not accept empty strings, what do you think? Maybe not right now to avoid breaking anything, but generally empty strings seem to be the exception.

SaberUK:
Closing this as it seems like there is no interest in it.

PR #1058, opened 2015-06-09
[2.0] Fix various issues with the nationalchars module.

SaberUK:
This fixes some of the more serious problems with the nationalchars module:

- Strip the directory name in the default casemapping value. This prevents the entire path from being shown in ISUPPORT.
- Error out if the casemapping value contains a space. This prevents invalid entries in ISUPPORT.
- Error out if the locale file failed to load. This fixes the case where errors could be silently ignored which is the cause of some confusion.
- Fix relative file path when building on Windows. The executable is not in the bin directory on Windows.
- Install nationalchars files on Windows.

### DIFF
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -1148,10 +1148,13 @@
 # 2) Allows using custom (national) casemapping over the network.
 #<module name="m_nationalchars.so">
 #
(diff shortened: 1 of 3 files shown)


attilamolnar:
Thanks!

PR #1060, opened 2015-06-14
[2.0] Fix system-wide installation of the data and log directories.

SaberUK:
The existing method works fine when building with the default paths but breaks when you are doing a system-wide installation.


attilamolnar:
Thanks!

PR #1061, opened 2015-06-14
[2.0] Fix mistakenly using Clang instead of GCC on older FreeBSD versions.

SaberUK:
Comparisons in Perl are sometimes really quite illogical.


attilamolnar:
Thanks!

PR #1063, opened 2015-06-15
[2.0] Show a better warning when certtool/openssl are missing.

SaberUK:
This converts this:

```
$ ./configure --generate-gnutls-cert
What is the hostname of your server?
[irc.example.com] -> 

What email address can you be contacted at?
[example@example.com] -> 

What is the name of your unit?
[Server Admins] -> 

What is the name of your organization?
[Example IRC Network] -> 

What city are you located in?
[Example City] -> 

What state are you located in?
[Example State] -> 

What is the ISO 3166-1 code for the country you are located in?
[XZ] -> 

How many days do you want your certificate to be valid for?
[365] -> 

Can't exec "certtool": No such file or directory at make/gnutlscert.pm line 142, <STDIN> line 8.

[CONFIGURE NOW CONTINUES INTO NORMAL INTERACTIVE MODE OBSCURING THE ERROR MESSAGE]
```

Into this:

```
$ ./configure --generate-gnutls-cert
Error: unable to find 'certtool' in the PATH!
```

This is needed because if you are missing the binary packages for your SSL library then it will fail to generate certificates but because `./configure` in 2.0 is awful it will continue running after encountering the error which obscures the error message from certtool/openssl. In my experience this is probably one of the most common SSL module-related problems encountered in \#inspircd.

### DIFF
--- a/make/gnutlscert.pm
+++ b/make/gnutlscert.pm
@@ -29,9 +29,16 @@ use Exporter 'import';
 use make::configure;
 our @EXPORT = qw(make_gnutls_cert);
 
+# On OS X the GnuTLS certtool is prefixed to avoid collision with the system certtool.
+my $certtool = $^O eq 'darwin' ? 'gnutls-certtool' : 'certtool';
 
 sub make_gnutls_cert()
 {
+	if (system "$certtool --version >/dev/null 2>&1")
+	{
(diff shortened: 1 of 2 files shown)


attilamolnar:
Nice, thanks!

PR #1070, opened 2015-07-04
Automatically /ACCEPT people when messaging them with callerid enabled.

### LINKED ISSUE #1064: If user with umode +g PMs someone, /accept them automatically
Charybdis (& recent forks) do this automatically as the user is probably expected to want response when they PM someone else.

```
2015-06-16 11:16:01+0300 < Mikaela> please tell me that when I PM someone with umode +g, InspIRCd /accepts them
2015-06-16 11:20:11+0300 <+Sheogorath> what?
2015-06-16 11:20:37+0300 <+Sheogorath> no, the user have to /accept you before you can msg him
2015-06-16 11:20:53+0300 <+Sheogorath> iirc there is an oper override option build in :X
2015-06-16 11:23:35+0300 < Mikaela> I mean that if I am on umode +g and /msg Sheogorath assuming you are umode -g, you are automatically accepted and can PM me back?
2015-06-16 11:23:58+0300 <+Sheogorath> mhm I guess no
2015-06-16 11:24:16+0300 < Mikaela> I am just being annoyed with one ancient IRCd and hoping InspIRCd where they arem oving to with unknown schedule behaves better
2015-06-16 11:24:16+0300 < Mikaela> :(
2015-06-16 11:24:16+0300 < Mikaela> should I open an issue?
2015-06-16 11:24:34+0300 < Mikaela> Charybdis and freenode-seven do that and it makes no sense having to /accept user when you PM them
2015-06-16 11:25:01+0300 <+Googolplexed> It does look like an oversight, looking at the module.
2015-06-16 11:25:28+0300 <+Sheogorath> well, Mikaela, let's simply test it
2015-06-16 11:25:40+0300 < Mikaela> ok
2015-06-16 11:25:54+0300 < Mikaela> 2015-06-16 11:25:38+0300 -- Sheogorath madgod@ChatSpikemm6h11.de :is messaging you, and you have umode +g. Use /ACCEPT +Sheogorath to allow.
2015-06-16 11:25:54+0300 < Mikaela> :(
2015-06-16 11:25:57+0300 < Mikaela> ==> issue
2015-06-16 11:25:57+0300 <+Sheogorath> ah * Mikaela :is in +g mode (server-side ignore).
2015-06-16 11:25:57+0300 <+Sheogorath> [10:25:38]* Mikaela :has been informed that you messaged them.
2015-06-16 11:26:49+0300 <+Sheogorath> well, should be not so difficult
2015-06-16 11:26:57+0300 <+Sheogorath> *shouldn't be
```

SaberUK:
This should fix #1064.

### DIFF
--- a/src/modules/m_callerid.cpp
+++ b/src/modules/m_callerid.cpp
@@ -379,11 +379,18 @@ public:
 
 	ModResult OnUserPreMessage(User* user, void* voiddest, int target_type, std::string& text, char status, CUList& exempt_list, MessageType msgtype) CXX11_OVERRIDE
 	{
-		if (!IS_LOCAL(user) || target_type != TYPE_USER)
+		if (!IS_LOCAL(user) || target_type != TYPE_USER || user == voiddest)
 			return MOD_RES_PASSTHRU;
 
 		User* dest = static_cast<User*>(voiddest);
-		if (!dest->IsModeSet(myumode) || (user == dest))
+		callerid_data* userdata = cmd.extInfo.get(user, true);
+		if (user->IsModeSet(myumode) && !userdata->accepting.count(dest))
+		{
+			user->WriteNotice("Automatically adding " + dest->nick + " to your accept list.");
+			cmd.AddAccept(user, dest);
+		}
+
+		if (!dest->IsModeSet(myumode))
 			return MOD_RES_PASSTHRU;
 
 		if (operoverride && user->IsOper())


Mikaela:
Could the same be also done with umode +R (this probably requires separate issue?)?

SaberUK:
Probably not. The `regdeaf` mode does not have a whitelist to add people to.

rburchell:
Maybe accept should be a more global form of whitelist, rather than just for that single module.

attilamolnar:
This doesn't handle the case when the accept list is full, it sends the auto adding message and then the list is full numeric on every PRIVMSG.

I think what this PR is trying to accomplish should be done by the client, we already give clients an interface that allows them to do this. Adding clever behavior like this to callerid will make it harder, if not impossible, for clients to understand what's going on exactly server-side with the list.

I'm not sure whether letting an accepted user bypass more restrictions such as m_dccallow, m_restrictmsg or m_silence is a good idea, considering those modules either have their own whitelist functionality or don't have one by design. It may make some sense for +R, but IMO consistency is worth far more than solving one particular corner case.

Mikaela:
> I think what this PR is trying to accomplish should be done by the client, we already give clients an interface that allows them to do this.

Which again requires IRCv3 to write specification for it, because otherwise no one is going to implement it?

attilamolnar:
Yes, having an IRCv3 spec for callerid would be nice. InspIRCd shouldn't ship this workaround and change its callerid interface (which has been stable for more than 5 years) only to provide a half-working usability improvement that clients can already do much better as they are in the position to have far more context.

kaniini:
@attilamolnar 

I am just replying here to explain why we changed CALLERID behaviour in charybdis, and hopefully this rationale will explain why an IRCv3 specification to outsource the problem to clients is unacceptable (but CALLERID's numerics *should* be standardised *anyway*).

So the rationale behind charybdis's change to CALLERID is quite simple.

A typical harassment tactic we observed on several networks was that users would harass another user and then hide behind CALLERID to avoid conversation about their harassment.  What charybdis does to solve this problem is that if you PRIVMSG a user, it will add the user to your accept list.  The reason why this is, is so that a user cannot hide behind CALLERID after saying things like "I hope you get hit by a bus."  After this change was implemented, it basically shut down that pattern of abuse, and it is my opinion that the original designer of CALLERID omitted this functionality simply because it was not a foreseen issue at that time.

I strongly suggest for anti-abuse purposes that InspIRCd at least make this behaviour a configurable option with the default being enabled.  That way networks which really want it may opt-out.

kaniini:
Mikaela asked if I could clarify some other points of charybdis' CALLERID system:

1. the CALLERID list is always there, it is not reset on `MODE +/-g`,
2. charybdis only counts manually accepted entries against the CALLERID limit, automatic entries do not get limited in this way (well, kinda/sorta, it's complicated but in practice this is how it typically works out).

The combination of this is what provides the anti-abuse aspect.

Mikaela:
And `/accept` happens even without being in umode +g if I understood correctly.

kaniini:
Correct.  You have to explicitly `/accept -nick` to work around what Charybdis is doing.  Moving to target-change tracking instead would close that loophole, as to run away from the victim you would have to cycle all of your target-change slots, which would severely gimp your client for a while anyway.

Adam-:
Plexus 4 has been using targets for about a year now for this problem. I think it is a better solution than auto /accept. Whether or not cycling the connection gets around it or not depends on the implementation of the targeting system. Charybdis stores hashes of the uids of the users, so cycling will change the hash.

Issue #263, opened 2012-07-19
[2.0] m_hideoper +H opers are still counted in /lusers


CuleX:
/lusers shows the current oper count, regardless whether the opers are +H or not. This leads to ghost opers that can't be found with /stats P.

Assume you have oper A (+o) and B (+oH). /lusers will show 2 opers online in numeric 252.

attilamolnar:
Fixed.

PR #1071, opened 2015-07-04
[2.0] Fix hidden operators being shown in LUSERS.

SaberUK:
Fixes bug #263.

### DIFF
--- a/src/modules/m_hideoper.cpp
+++ b/src/modules/m_hideoper.cpp
@@ -28,25 +28,43 @@
 class HideOper : public SimpleUserModeHandler
 {
  public:
+	size_t opercount;
+
 	HideOper(Module* Creator) : SimpleUserModeHandler(Creator, "hideoper", 'H')
+		, opercount(0)
 	{
 		oper = true;
+	}
+
+	ModeAction OnModeChange(User* source, User* dest, Channel* channel, std::string& parameter, bool adding)
+	{
+		if (SimpleUserModeHandler::OnModeChange(source, dest, channel, parameter, adding) == MODEACTION_DENY)
+			return MODEACTION_DENY;
+
+		if (adding)
+			opercount++;
+		else
+			opercount--;
+
+		return MODEACTION_ALLOW;
 	}
 };
 
 class ModuleHideOper : public Module
 {
 	HideOper hm;
+	bool active;
  public:
 	ModuleHideOper()
 		: hm(this)
+		, active(false)
 	{
 	}
 
 	void init()
 	{
 		ServerInstance->Modules->AddService(hm);
-		Implementation eventlist[] = { I_OnWhoisLine, I_OnSendWhoLine, I_OnStats };
+		Implementation eventlist[] = { I_OnWhoisLine, I_OnSendWhoLine, I_OnStats, I_OnNumeric, I_OnUserQuit };
 		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
 	}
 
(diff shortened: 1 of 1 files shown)


src/modules/m_hideoper.cpp
```diff
@@ -36,6 +36,7 @@ class HideOper : public SimpleUserModeHandler
 
 class ModuleHideOper : public Module
 {
+	bool active;
```
attilamolnar: This should be inited to false in the constructor

GLolol:
:+1:

attilamolnar:
Discussion in August with @Adam-

     <Adam> SaberUK for #1071 cant you use a modewatcher or something more intelligent, this seems ugly
     <Adam> and is frequently hit
     <SaberUK> Adam: maybe, but i copied what the code used elsewhere
     <Adam> wheres that
     <SaberUK> OnWhoisLine
     <Adam> thats not the same thing though
     <Adam> thats a yes/no to show the oper, its not counting per-call
     <Adam> look at how InvisibleWatcher works and do that?
     <SaberUK> sure

Did you look into that or what happened?

attilamolnar:
Using a modewatcher is not necessary since the mode is in the module, the modehandler could update the count

src/modules/m_hideoper.cpp
```diff
@@ -28,25 +28,37 @@
 class HideOper : public SimpleUserModeHandler
 {
  public:
+	int opercount;
```
attilamolnar: This should be inited in the constructor

src/modules/m_hideoper.cpp
```diff
@@ -60,6 +72,27 @@ class ModuleHideOper : public Module
 		return Version("Provides support for hiding oper status with user mode +H", VF_VENDOR);
 	}
 
+	void OnUserQuit(User* user, const std::string&, const std::string&)
+	{
+		if (user->IsModeSet('H'))
+			hm.opercount -= 1;
```
attilamolnar: `hm.opercount--;` is easier to read

attilamolnar:
Have you tested it yet?

src/modules/m_hideoper.cpp
```diff
@@ -60,6 +73,27 @@ class ModuleHideOper : public Module
...
+	ModResult OnNumeric(User* user, unsigned int numeric, const std::string& text)
+	{
+		if (numeric != 252 || active || user->HasPrivPermission("users/auspex"))
+			return MOD_RES_PASSTHRU;
+
+		// If there are no visible operators then we shouldn't send the numeric.
+		if (hm.opercount)
+		{
+			active = true;
+			user->WriteNumeric(252, "%s %d :operator(s) online", user->nick.c_str(), hm.opercount);
```
rburchell: RPL_LUSEROP

src/modules/m_hideoper.cpp
```diff
@@ -60,6 +73,27 @@ class ModuleHideOper : public Module
...
 
+	void OnUserQuit(User* user, const std::string&, const std::string&)
+	{
+		if (user->IsModeSet('H'))
+			hm.opercount--;
+	}
+
+	ModResult OnNumeric(User* user, unsigned int numeric, const std::string& text)
+	{
+		if (numeric != 252 || active || user->HasPrivPermission("users/auspex"))
```
rburchell: same

src/modules/m_hideoper.cpp
SaberUK: We don't have a constant for this.

SaberUK:
I'm convinced this is unfixable without reverting to my original inefficient counting method because:

1. OnOper/OnPostOper only work for local opers and is therefore useless for counting and even if they did work for remote opers theres no OnUnOper/OnPostUnOper so you couldn't decrement the count.
2. ModeWatcher on the oper mode will not work because -o only calls the mode watcher when explicitly unset (i.e. when not reopering) and +o does not call the mode watcher period.

What do you suggest, @attilamolnar?

attilamolnar:
You could count the hidden opers and subtract them from the number of opers.

SaberUK:
@attilamolnar Done and tested, hopefully for the last time.

src/modules/m_hideoper.cpp
```diff
@@ -28,25 +28,38 @@
...
+
 	HideOper(Module* Creator) : SimpleUserModeHandler(Creator, "hideoper", 'H')
+		, opercount(0)
 	{
 		oper = true;
 	}
+
+	ModeAction OnModeChange(User* source, User* dest, Channel* channel, std::string& parameter, bool adding)
+	{
+		if (dest->IsModeSet(this->GetModeChar()) != adding)
```
attilamolnar: Instead of reimplementing `SimpleUserModeHandler::OnModeChange()`, just call it and if it returns `MODEACTION_ALLOW` do the increment/decrement. This makes it so if SimpleUserModeHandler changes then this won't break.

attilamolnar:
Thanks! This breaks if +H is set on a non-oper but that's acceptable: setting oper-only modes on non-opers is not supported and in the case of +H, it's nonsensical as well.

PR #1072, opened 2015-07-04
add custom prefix and suffix for cloak by connectclass

Sheogorath-SI:


src/modules/m_cloaking.cpp
```diff
@@ -231,8 +240,8 @@ class ModuleCloaking : public Module
 			// pfx s1.s2. (xxx.xxx or s3) sfx
 			rv.reserve(prefix.length() + 15 + suffix.length());
 		}
-
-		rv.append(prefix);
+
```
SaberUK: Trailing whitespace should be removed.

attilamolnar:
What's the point in this?

SaberUK:
@attilamolnar It's useful if you have users you want channels to be able to filter e.g. having a connect class that all Tor/BNC users filter into so they can be banned/exempted easily.

attilamolnar:
@SaberUK Put a label on them and ban that using an extban, I don't see why do we need to involve m_cloaking in that.

Sheogorath-SI:
Well, on my net we have users using a meshed vpn for connecting. So we have multiple users from different private IP Ranges. Some users have a full /24 net. so we add some prifix like network/somevpn/ as prefix. so yes, we can also do it with "m_hostchange" but without an real identifier.

attilamolnar:
That's a very special setup. I believe that changing m_cloaking is not the correct approach, rather an extban is a much cleaner approach here.

Issue #920, opened 2014-09-18
GeoIP deny doesn't match


SISheogorath:
Hi,

got a little bug, geoip did not denied the IP from an User connecting via WebIRC (using cgi). The changed host is not recognized by it. 

<connect deny="*" geoip="TH,TJ,TK,TL,TM,TN,TO,TR,TT,TV,TW,TZ,UA,UG,UM,UY,UZ,VA,VC,VE,VG,VI,VN,VU,WF,WS,YE,YT,ZA,ZM,ZW" reason="You are not allowed to connect from that country!">

\* [Gast55883] is connected from Turkey (TR)

There is no <connect allow> before the deny block. and the IP is not recognized in Stats G so it simply doesn't trigger on IP change. (whois result by m_geoipban).

Greets
Sheo

PeLLw:
Hello everyone

I want to ask if they can help why does not the GEOIP.
It does not work at all. All it installed no errors, but nothing works geoip.

I hope you can answer, thank you very much.

SISheogorath:
Out dated and patch ported to master

PR #1074, opened 2015-07-18
fix cgiirc is not matching geoip

Sheogorath-SI:
reference to #920


attilamolnar:
I'll port this to master because I can't test this adequately and therefore unsure of what it may possibly break, which is why I haven't committed this earlier to 2.0. Plus it's a corner case therefore less important for 2.0.

PR #1076, opened 2015-07-19
[2.0] Fix checking whether kqueue/stdint/strlcpy are available.

SaberUK:
Reported by @Adam- who has recently joined the OS X master race.

It seems like the newer OS X versions don't put headers into the system include directories. Switch to using a test file instead.

The test file was backported from 2.2.

### DIFF
--- a/make/check_kqueue.cpp
+++ b/make/check_kqueue.cpp
@@ -1,0 +1,26 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2015 Peter Powell <petpow@saberuk.com>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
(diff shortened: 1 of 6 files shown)


attilamolnar:
Thanks, will merge after @Adam- tests it. The old code was fragile and very ugly, if this hack is done anywhere else please fix it too.

make/check_kqueue.cpp
```diff
@@ -0,0 +1,25 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
```
rburchell: author copyright information is missing

attilamolnar:
strlcpy is detected in a similar way: https://github.com/inspircd/inspircd/blob/insp20/configure#L375

make/check_kqueue.cpp
SaberUK: It was missing in the files in master too. Should I file PR to add it there too?

SaberUK:
@attilamolnar I have to go out now but i'll fix that when I get home.

make/check_kqueue.cpp
attilamolnar: @SaberUK It's fine to add it in insp20 and it'll be added to master when insp20 is merged

SaberUK:
@attilamolnar Done!

attilamolnar:
@Adam- Please confirm that it detects kqueue fine after the fix.

Adam-:
This works.

attilamolnar:
Thanks!

PR #1077, opened 2015-07-20
IRCv3.2 - Add cap-notify and invite-notify modules

AlphaT3ch:
Add cap-notify and invite-notify modules


SaberUK:
Please don't file pull requests that you haven't even tested. These modules will *not* work with 2.2.

Also, this is a duplicate of #1037.

PR #1080, opened 2015-07-22
Replace deprecated tmpnam() call

bradfirj:
From Perl 5.22 onwards, POSIX::tmpnam() has been deprecated (without the
usual 2 year deprecation cycle), using the File::Temp module instead
preserves compatibility while allowing compilation on 5.22 and later.

As it stands, Fedora Rawhide is failing like so:

Configuration failed. The following error occured:

Calling POSIX::tmpnam() is deprecated at make/utilities.pm line 407, line 32.

### DIFF
--- a/make/utilities.pm
+++ b/make/utilities.pm
@@ -29,6 +29,7 @@ use warnings FATAL => qw(all);
 
 use Exporter 'import';
 use POSIX;
+use File::Temp;
 use Getopt::Long;
 use Fcntl;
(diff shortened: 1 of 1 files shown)


SaberUK:
@bradfirj Can you confirm that File::Temp::tmpnam() existed in Perl 5.8? If so this LGTM.

bradfirj:
@SaberUK If the [documentation](http://perldoc.perl.org/5.8.8/File/Temp.html) can be trusted, the module and the function exist right back to 5.6.1

It's definitely present in 5.8.8 and 5.8.9

SaberUK:
Great. Thanks.

attilamolnar:
Thanks!

PR #1081, opened 2015-07-23
A module to command /IRCOPS that prints out a list of current online.

antoniocostacf:
This module, allow a user type a command /ircops and get a list of ircops online that's avaliable and ircops away.

### DIFF
--- a/src/modules/m_ircops.cpp
+++ b/src/modules/m_ircops.cpp
@@ -57,14 +57,14 @@ public:
 
 	}
 
-	CmdResult Handle (const std::vector<std::string> &parameters, User *user)
+	CmdResult Handle (const std::vector<std::string>&, User *user)
 	{
 		std::string ircops_str;
-		ircops_str = "Total:";
 
 		if (user)
 		{
-			user->WriteServ("============ Online IRC Operators =============");
+			ircops_str += "============ Online IRC Operators ============= \n";
+			
 			int total = 0, total_away = 0;
 			for (std::list<User*>::iterator i = ServerInstance->Users->all_opers.begin(); i != ServerInstance->Users->all_opers.end(); i++)
 			{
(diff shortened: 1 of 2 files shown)


Sheogorath-SI:
How about using alias and `/stats P`

antoniocostacf:
@Sheogorath-SI , you need be a IRCop to use this ur suggestion.

Sheogorath-SI:
No. Read: https://github.com/inspircd/inspircd/blob/master/docs/conf/inspircd.conf.example#L740-L741

SaberUK:
This duplicates existing behaviour. :-1:

antoniocostacf:
Oh, sorry i've deleted my comment that i did comment you need be a ircop to check this. 

Nice, i've not know that. But i got this not check the away status. It's just mark IDLE.

src/modules/m_ircops.cpp
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	CmdResult Handle (const std::vector<std::string> &parameters, User *user)
+	{
+		std::string ircops_str;
+		ircops_str = "Total:";
+
+		if (user)
+		{
+			user->WriteServ("============ Online IRC Operators =============");
```
SaberUK: This will be parsed by clients as having a command name of "============" and parameters of "Online", IRC", "Operators", and "=============". This breaks the protocol.

antoniocostacf: To be with accord protocol, should I use like: user->WriteServ("%s", "========");  ?

SaberUK: You need to give it a command name and prefix the final parameter with a colon.

This will probably not be merged though as it almost entirely duplicates existing core behaviour. You might have better luck over at [inspircd-extras](https://github.com/inspircd/inspircd-extras) but even then i'm not sure if it would be merged.

Sheogorath-SI:
by the way: you know that you are merging against master and not insp20?

antoniocostacf:
I got this now, do u know how i cancel or change this ?

PR #1082, opened 2015-07-23
Update m_hideoper.cpp - insp20

antoniocostacf:
I've change at code the ident@host, to check the type name of IRCop, also it's check and show  if the ircop's still have a away msg ! 
I did remove the use of User *i = oper; using just the iterator with pointers properly.


del6597:
What about the code in core_stats.cpp?

attilamolnar:
Thanks but 2.0 is bugfix only, also this only works if m_hideoper is loaded.

PR #1089, opened 2015-08-09
[2.0] Fix compilation with Microsoft Visual C++ 14.

SaberUK:
Microsoft are actually implementing standardised stuff now. Crikey.

### DIFF
--- a/win/inspircd_win32wrapper.h
+++ b/win/inspircd_win32wrapper.h
@@ -63,6 +63,7 @@
 
 /* Disable the deprecation warnings.. it spams :P */
 #define _CRT_SECURE_NO_DEPRECATE
+#define _WINSOCK_DEPRECATED_NO_WARNINGS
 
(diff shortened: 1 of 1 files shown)


win/inspircd_win32wrapper.h
```diff
@@ -63,6 +63,7 @@
 
 /* Disable the deprecation warnings.. it spams :P */
 #define _CRT_SECURE_NO_DEPRECATE
+#define _WINSOCK_DEPRECATED_NO_WARNINGS
```
SaberUK: I looked at getting rid of the deprecated code that this #define shuts up but it requires us to use features that aren't available in all compiler versions that we support.

attilamolnar:
Thanks, although I don't think compatibility with newer Visual Studio versions is particularly important for 2.0 as @Adam- uses an older version for compiling 2.0 releases.

PR #1092, opened 2015-08-26
[2.0] Fix sending the wrong numerics on join when a topic is empty.

### LINKED ISSUE #1091: Cleared/unset topic misleading behavior
When one sets a topic on a channel and clears/unsets it again, the ircd sends you an empty topic upon rejoining the channel. I found this out while testing if this was a (caching) bug in an IRC client or not.

This is misleading behavior because when you do a `/TOPIC #channel` you get a "331 No topic set." message, yet upon joining such channels you always get served this empty topic:
`:<server> 332 Robby #Test3 :` (and the numeric 333 message that goes with it).

Either this should never be sent to the client upon joining in the first place when the topic is empty, or `/TOPIC #channel` should also return this empty topic. I'm for the former, I never liked the latter implementation.

If I remember correctly it is UnrealIRCd that has (or had) the latter behavior, because in this way it was visible who had last unset the topic, when asked about this behavior.

SaberUK:
The `topicset` variable is the time the topic was changed last not whether there is a topic set.

Fixes #1091.

### DIFF
--- a/src/channels.cpp
+++ b/src/channels.cpp
@@ -412,7 +412,7 @@ Channel* Channel::ForceChan(Channel* Ptr, User* user, const std::string &privs, 
 
 	if (IS_LOCAL(user))
 	{
-		if (Ptr->topicset)
+		if (!Ptr->topic.empty())
 		{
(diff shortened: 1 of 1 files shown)


attilamolnar:
This removes a feature (users can no longer see who unset a topic) therefore should not be done in insp20.

PR #1093, opened 2015-09-02
Update another password check to use InspIRCd::TimingSafeCompare.

SaberUK:


attilamolnar:
Why is this necessary for bcrypt and pbkdf2? As I understand timing attacks for comparisons only work if you know one side of the comparison which is not the case with salts.

SaberUK:
@attilamolnar I guess so; fixed.

attilamolnar:
Thanks!

PR #1095, opened 2015-09-09
[2.0] Move RFC-compliant behaviour behind a config option.

SaberUK:
Fixes the compatibility issue mentioned by @attilamolnar in #1092.

### DIFF
--- a/docs/conf/inspircd.conf.example
+++ b/docs/conf/inspircd.conf.example
@@ -614,7 +614,12 @@
          # welcomenotice: When turned on, this sends a NOTICE to connecting users
          # with the text Welcome to <networkname>! after successful registration.
          # Defaults to yes.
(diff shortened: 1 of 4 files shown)


attilamolnar:
LGTM code-wise. Your title implies the current behavior is not RFC compliant, but I can't find the relevant part. Can you show me please what makes the original behavior incompliant?

Robby-:
The title is neither right or wrong. I've taken a look at the RFC and the situation on what to do with cleared topics (either on /TOPIC or on /JOIN) sadly is actually not documented, so that leaves this open to developer preference on how they wish to deal with cleared topics.

attilamolnar:
Closing as the original behavior was never incompliant in the first place. However, b0e64b8 is ok for master.

Issue #711, opened 2013-12-16
[2.0] m_filter: add a 'notifyuser' option, or a new flag to do this per filter


Robby-:
I'd like to request that m_filter gets a new option, similar to what m_shun has, 'notifyuser'. So it can be chosen if the user should be notified of the blocking of their message or not. Perhaps even as a flag, so this can be chosen per filter. Although I'd be happy with either solution, as long as the user notifications can be disabled.

Currently m_filter, when a message is blocked, notifies the user of the block, causing them to try changing letters, etc, in attempt to still get their message through.

PR #1096, opened 2015-09-09
[2.0] Optionally do not notify users if their messages are blocked by certain modules.

Robby-:
Optionally do not notify users if their messages are blocked by certain modules.

This fixes issue #711, although that issue only mentions m_filter, the m_chanfilter and m_muteban modules also needed this change.

### DIFF
--- a/src/modules/m_chanfilter.cpp
+++ b/src/modules/m_chanfilter.cpp
@@ -70,6 +70,7 @@ class ModuleChanFilter : public Module
 {
 	ChanFilter cf;
 	bool hidemask;
+	bool notifyuser;
 
  public:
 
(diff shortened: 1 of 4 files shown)


src/modules/m_chanfilter.cpp
```diff
@@ -110,10 +113,13 @@ class ModuleChanFilter : public Module
 			{
 				if (InspIRCd::Match(text, i->mask))
 				{
-					if (hidemask)
-						user->WriteNumeric(404, "%s %s :Cannot send to channel (your message contained a censored word)",user->nick.c_str(), chan->name.c_str());
-					else
-						user->WriteNumeric(404, "%s %s %s :Cannot send to channel (your message contained a censored word)",user->nick.c_str(), chan->name.c_str(), i->mask.c_str());
+					if (notifyuser)
```
SaberUK: In cases like this you can do the following to avoid having a massive diff:

```cpp
if (!notifyuser)
    return MOD_RES_DENY;
```

Robby-:
Good call, I didn't think about that, I'll change it.

Robby-:
Fixed in m_chanfilter and m_muteban, but due to how m_filter works, I can't do that there.

attilamolnar:
Thanks but 2.0 is bugfix only, please port this to master then I'll look into it.

PR #1098, opened 2015-09-23
Fix detection of the Apple variant of Clang.

SaberUK:
In the new Xcode release Apple removed the LLVM version from the version string. This is a problem as our version detecting code relied on it.

To solve this we split off the Apple variant of Clang and treat it if it was a different compiler. This mirrors the change implemented by other build systems like CMake 3.0.


attilamolnar:
Thanks!

PR #1103, opened 2015-10-09
Some updates to INFO.

rburchell:
Simple changes to bring things up to date with today's reality.


attilamolnar:
Probably the [messages on boot](https://github.com/inspircd/inspircd/blob/master/src/inspircd.cpp#L371) should be updated as well if you agree.

rburchell:
Ah, damn, I remember forgetting to change those in the past too I'm sure. Maybe we should just remove the list and add a "For info on authors & contributors, see /INFO"?

attilamolnar:
OK, I'll merge when you're done with that.

rburchell:
Done (and rebased, for good measure).

Sorry for the delay.

attilamolnar:
Thanks!

PR #1104, opened 2015-10-20
[2.0] Generate a 2048-bit dhparams unless explicitly specified.

SaberUK:
The reason for this change is because a 1024-bit prime is feasibly factorable in a year by attackers with a large enough budget to do so.

This is a change in build system behaviour but it should not cause any problems although I have added a backwards compatibility switch just in case. In my opinion the aforementioned risk is big enough to justify this.

GnuTLS is not affected as it defaults to 2048-bit certificates by default.

References:
- https://freedom-to-tinker.com/blog/haldermanheninger/how-is-nsa-breaking-so-much-crypto/
- https://weakdh.org/imperfect-forward-secrecy-ccs15.pdf

### DIFF
--- a/make/opensslcert.pm
+++ b/make/opensslcert.pm
@@ -46,6 +46,7 @@ sub make_openssl_cert()
 	my $state = promptstring_s('What state are you located in?', 'Example State');
 	my $country = promptstring_s('What is the ISO 3166-1 code for the country you are located in?', 'XZ');
 	my $time = promptstring_s('How many days do you want your certificate to be valid for?', '365');
+	my $use_1024 = promptstring_s('Do you want to generate less secure dhparams which are compatible with old versions of Java?', 'n');
 	print FH <<__END__;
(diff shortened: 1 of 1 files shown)


attilamolnar:
We can generate 2048 bit certs in all cases, no question required. However, > 1024 bit DH params will break Java older than Java 8, see the [Mozilla TLS wiki](https://wiki.mozilla.org/Security/Server_Side_TLS#DHE_and_Java) for more info, so the question should only affect that.

SaberUK:
@attilamolnar I have changed the question.

Robby-:
But now you will also generate 1024-bit certificates when yes is answered. The issue is that only the DH params cannot be bigger than 1024-bit. Certificates, however, can be 2048-bit with no issue.

attilamolnar:
Yes.

attilamolnar:
Thanks!

PR #1105, opened 2015-11-03
Abort rehash if included executable returns non 0

Adam-:


attilamolnar:
As discussed on IRC, this doesn't work because of https://github.com/inspircd/inspircd/blob/master/src/inspircd.cpp#L134 (SIGCHLD disposition set to SIG_IGN) causes the system to reap processes spawned by us automatically, interfering with `pclose()` getting the exit code.

PR #1106, opened 2015-11-12
Remove the PKG_CONFIG_PATH extension code from the build system.

SaberUK:
This fixes the problems Jenkins was having on OS X.

### DIFF
--- a/make/utilities.pm
+++ b/make/utilities.pm
@@ -90,23 +90,9 @@ sub make_rpath($;$)
 	return $output;
 }
 
-sub extend_pkg_path()
-{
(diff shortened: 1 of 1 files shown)


attilamolnar:
Is this still required? Jenkins seems to have no problem building it at the moment.

SaberUK:
@attilamolnar I think that fraggeln has applied the workaround I suggested which is why its working.

This code was due for removal anyway as it breaks systems with a non-"standard" default `pkg-config` path. I just hadn't got around to doing it.

attilamolnar:
Thanks!

PR #1110, opened 2015-11-26
[2.0] Fix CAP REQ to be atomic like the standard dictates.

SaberUK:
[According to the cap-3.1 specification:](http://ircv3.net/specs/core/capability-negotiation-3.1.html)

>The capability identifier set must be accepted as a whole, or rejected entirely.

Our failure to follow the standard was apparently causing problems with irssi (according to @dequis).

This is a bit of a hack because I didn't want to have to rewrite the entire CAP API.

### DIFF
--- a/src/modules/m_cap.h
+++ b/src/modules/m_cap.h
@@ -21,6 +21,8 @@
 #ifndef M_CAP_H
 #define M_CAP_H
 
+class GenericCap;
+
 class CapEvent : public Event
 {
  public:
@@ -35,6 +37,7 @@ class CapEvent : public Event
 	CapEventType type;
 	std::vector<std::string> wanted;
 	std::vector<std::string> ack;
+	std::vector<std::pair<GenericCap*, int> > changed; // HACK: clean this up before 2.2
 	User* user;
 	CapEvent(Module* sender, User* u, CapEventType capevtype) : Event(sender, "cap_request"), type(capevtype), user(u) {}
 };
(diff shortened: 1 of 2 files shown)


rburchell:
I'm curious to know what problems this was causing. I know I pointed out the same problem to you some weeks ago when I was implementing CAP support for something else, but I really had trouble imagining what would _rely_ on this behavior.

src/modules/m_cap.h
```diff
@@ -66,6 +69,7 @@ class GenericCap
 				{
 					// we can handle this, so ACK it, and remove it from the wanted list
 					data->ack.push_back(*it);
+					data->changed[this] = ext.get(data->user);
```
attilamolnar: `ext.set()` returns the previous value, should use that

src/modules/m_cap.h
```diff
@@ -35,6 +37,7 @@ class CapEvent : public Event
 	CapEventType type;
 	std::vector<std::string> wanted;
 	std::vector<std::string> ack;
+	std::map<GenericCap*, int> changed; // HACK: clean this up before 2.2
```
attilamolnar: There's no need for a map, a `std::vector<std::pair<GenericCap*, bool> >` (or int) is enough.

attilamolnar:
OK, please split this into two commits, one for adding the `trim()` and another one for what's in the title of the PR.

attilamolnar:
Thanks!

dequis:
:heart:

PR #1121, opened 2015-12-28
Replace deprecated POSIX::tmpnam() by File::Temp::tmpnam()

guikcd:
POSIX::tmpnam() seems to be deprecated since Perl 5.22, i propose to use File::Temp instead.
The only official link i've found about this deprecation is in https://metacpan.org/pod/distribution/perl/ext/POSIX/lib/POSIX.pod#FUNCTIONS.

The Debian reproducible build system catch this problem, see https://reproducible.debian.net/rbuild/unstable/amd64/inspircd_2.0.20-4.rbuild.log.

### DIFF
--- a/make/utilities.pm
+++ b/make/utilities.pm
@@ -33,7 +33,7 @@ use Exporter 'import';
 use Fcntl;
 use File::Path;
 use Getopt::Long;
-use POSIX;
+use File::Temp ();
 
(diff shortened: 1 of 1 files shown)


Adam-:
This was already fixed in #1080. The commit will be merged into master after the next release.

SaberUK:
P.S. We should really do a release soon. People on bleeding edge distros will be hitting into this when trying to use 2.0.20.

guikcd:
At least on the master branch, the POSIX module is unused on *make/utilities.pm*.

PR #1126, opened 2016-01-11
Fix warnings when building under GCC 6.

SaberUK:
This fixes the following warnings

`warning: '-pedantic' is not an option that controls warnings [-Wpragmas]` (it is on older versions)

`warning: 'template<class> class std::auto_ptr' is deprecated [-Wdeprecated-declarations]` (only in C++11 which we don't use yet, when buildtool is done this will be the default)


attilamolnar:
Thanks!

PR #1127, opened 2016-01-11
Remove the useless testnet module.

SaberUK:
This modules main use was for testing whether modules are hooked to events but since d24619c012b34d5a3d4cfb93e7bea3ff3d5721e7 modules are automatically hooked onto events which removed the main point of the module. There is still some remaining functionality but what does remain is pretty much useless.


attilamolnar:
Thanks!

PR #1128, opened 2016-01-11
Expand paths beginning in ./ in FileSystem::ExpandPath.

### LINKED ISSUE #1115: Cannot give relative path as config file
```
$ inspircd --nofork --config ./inspircd.conf
ERROR: Cannot open config file: /home/progval/.local/conf/./inspircd.conf
Exiting...
$ ls inspircd.conf 
inspircd.conf
```

Expected behavior: local `inspircd.conf` is used instead of trying (and failing) to use a file relative to the prefix.

Version: ac861f854ccee2515c191068cc2274e5d1a92fa8 (Git master at the time I am writing this)

SaberUK:
This fixes #1115 by @ProgVal.

### DIFF
--- a/src/fileutils.cpp
+++ b/src/fileutils.cpp
@@ -21,6 +21,10 @@
 
 #include <fstream>
 
+#if defined _WIN32
+# define PATH_MAX MAX_PATH
+#endif
+
 FileReader::FileReader(const std::string& filename)
 {
 	Load(filename);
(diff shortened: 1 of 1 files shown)


ProgVal:
Does it support `../` ?

SaberUK:
No, but implementing support for that would be annoying because I would need to implement all of `realpath()` because we don't want standard path expansion here (i.e. unless the user explicitly uses `./` it should be relative to the config directory.

SaberUK:
After discussing this with @attilamolnar I have decided that the best behaviour here would be to require absolute paths. If you want to use a path relative to the current directory with `--config` then use `$PWD/inspircd.conf` instead of `./inspircd.conf`.

Issue #478, opened 2013-04-08
We need more clever and sneaky ways to combat seasoned troublemakers and ban evaders


Robby-:
Every network most likely has their share of persistent troublemakers and ban evaders.

The seasoned troublemakers already know the ropes and tricks to get around things like bans (much to operators annoyance), so we must act in a more clever and sneaky way, which we already do by for example shunning.

To start, I have an idea to complement the shunning of users:
Right now, we can shun them so they can stay and not start evading and we can thus 'contain' them sort to speak. But they still can see chats and private messages, so, the more clever ones know something is up when they type something that is bound to get a reaction and they don't get any from anybody in the channel or in private. A way to make them deaf (not with umode +d, as that mode change is shown to the user as it happens) for either or both channel and private messages/notices would be good, so they can't see who is actively chatting, and about what they're talking (so lurking is ruled out), and they thus will not know who they should expect a reaction from. Of course, there are still ways for them to find out if this has been done to them, but again, that's for the more knowledgeable ones amongst them that know what to look for.

Another idea I was thinking about requires cooperation of services.
We have for example a persistent ban evader that always attempts to enter specific channels he is banned from. This user is not registered with NickServ and always changes nicks and everything, except his hostmask, which is always from the same ISP (he doesn't do proxies). A while ago I made a feature request for this which @Shawn-Smith created in the form of a 'U:' extban, it allows to set bans on hostmasks so that unregistered users who match that mask can not enter without registering first. Very nice. But now I was also thinking to complement that with optional account aging. For example, require a nick from a specific mask (or *!*@\* for everyone) to be registered for atleast N days before the user can enter. Of course, for this to work, services must send the nick registration timestamp to the server. This could be done upon identifying and stored as additional metadata maybe? Just thinking out loud here.

Yet another idea is to have gecos banning implemented at the ircd-level. Right now a bot handles this for us by looking at incoming connections. Although it is made to also look at the ident and host mask of users so we can narrow it down to for example specific IP-ranges or hostmasks if needed or desired (for example to reduce false positives).

I'll leave it at this for now.

Any ideas, suggestions or thoughts are welcome.

PR #1134, opened 2016-02-13
Optionally do not notify users if their messages are blocked by certain modules.

Robby-:
Optionally do not notify users if their messages are blocked by certain modules.

This fixes issue #711, although that issue only mentions `filter`, the `chanfilter` and `muteban` modules also needed this change.

Ported to master as requested in #1096.

### DIFF
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -402,7 +402,10 @@
 # his/her message is blocked.
 #
 # If maxlen is set then it defines the maximum length of a filter entry.
-#<chanfilter hidemask="yes" maxlen="50">
+#
+# If notifyuser is set to no, the user will not be notified when
+# his/her message is blocked.
+#<chanfilter hidemask="yes" maxlen="50" notifyuser="yes">
 
 #-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#
 # Channel history module: Displays the last 'X' lines of chat to a user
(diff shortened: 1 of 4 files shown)


attilamolnar:
This doesn't interact well with echo-message, if that's loaded the user still knows their message wasn't delivered.

Robby-:
That's disappointing, I didn't think of that. Well, that is beyond my capabilities. If you or someone else wants to make this interact well with echo-message, please do so, because we need something like this.

del6597:
I don't think this can interact well with echo-message by definition of what it provides.

User sends a message on a server with echo-message enabled.
Scenario 1: The echo-message doesn't show up. The user now knows it was blocked/wasn't sent.
Scenario 2: The echo-message shows a fake message to only the user.  The user assumes the message went through. Additionally, if the user knows any known chat triggers from bots/anope/userscripts etc, then they could just test against those firing.

Both scenarios seem ugly, but the first one makes the most sense to me if there had to be a choice.

Robby-:
Either scenario is fine with me, I would disable `echo-message` anyway to achieve the desired effect.

Scenario 3: This is similar to scenario 1, the network/server administrators do not want to use or care about `echo-message` and they disable it in the configuration, in which case this works perfectly fine.

Maybe a note should be added saying that this may be detectable by the target when `echo-message` is enabled?

I'm aware that some users might try and poke some bots to see if they respond, and some have indeed tried this and discovered they were shunned/mutebanned because the bots wouldn't respond, but against the majority of users this is effective and is a nice feature to have.

Also note that the `shun` module already has this `notifyuser` setting that does the same thing, I just extended it to a few other modules that I believe should also have this setting. This patch brings those modules more in line with each other.

This patch is related to the idea of #478.

SaberUK:
I have filed an relevant amendment to the `echo-message` over here:

https://github.com/ircv3/ircv3-specifications/pull/312

jwheare:
The `echo-message` amendment was accepted. This PR can be safely updated to send fake echos when silently filtering.

Robby-:
This PR alone was a fairly simple change, but the now needed changes to support `echo-message` is beyond my capabilities. I've enabled `Allow edits from maintainers.` if someone wants to work on it (although currently GitHub returns a red warning at the top saying .

Robby-:
This PR alone was a fairly simple change, but the now needed changes to support `echo-message` is beyond my capabilities. I've enabled `Allow edits from maintainers.` if someone wants to work on it (although currently GitHub returns a red warning at the top saying `You can't perform that action at this time.`.

SaberUK:
I've got a local branch which makes some changes to the message events which will allow for this. I'll update this when it has been committed.

PR #1135, opened 2016-02-14
Remove the STARTTLS token from ISUPPORT.

SaberUK:
This is an InspIRCd-specific token which is pointless because:

1. You can't STARTTLS after registration is complete.
2. You can already discover STARTTLS support via cap `tls`.


attilamolnar:
Thanks!

PR #1139, opened 2016-02-19
Adds <commandlist> tag with command usage to httpd_stats.

johanna-a:
This patch adds command usage to httpd_stats. It is useful for our network but not tested in production yet. This feels like it should be togglable by configuration though, rather than to be included by default. Or maybe it could be included on a substring match of the URI? (e.g. /stats/?includeCommandList)

### DIFF
--- a/src/modules/m_httpd_stats.cpp
+++ b/src/modules/m_httpd_stats.cpp
@@ -217,7 +217,15 @@ class ModuleHttpStats : public Module, public HTTPRequestEventListener
 					data << "</server>";
 				}
 
-				data << "</serverlist></inspircdstats>";
+				data << "</serverlist><commandlist>";
+
+				const CommandParser::CommandMap& commands = ServerInstance->Parser.GetCommands();
+				for (CommandParser::CommandMap::const_iterator i = commands.begin(); i != commands.end(); ++i)
+				{
+					data << "<command><name>" << i->second->name << "</name><usecount>" << i->second->use_count << "</usecount></command>";
+				}
+
+				data << "</commandlist></inspircdstats>";
 
 				/* Send the document back to m_httpd */
 				HTTPDocumentResponse response(this, *http, &data, 200);


RobertOlsson:
+1

src/modules/m_httpd_stats.cpp
```diff
@@ -217,7 +217,19 @@ class ModuleHttpStats : public Module, public HTTPRequestEventListener
...
 				}
 
-				data << "</serverlist></inspircdstats>";
+				data << "</serverlist><commandlist>";
+
+
+				const CommandParser::CommandMap& commands = ServerInstance->Parser.GetCommands();
+				for (CommandParser::CommandMap::const_iterator i = commands.begin(); i != commands.end(); ++i)
+				{
+					if (i->second->use_count)
```
attilamolnar: Any reason for not displaying all commands?

src/modules/m_httpd_stats.cpp
```diff
@@ -217,7 +217,19 @@ class ModuleHttpStats : public Module, public HTTPRequestEventListener
...
+					{
+						data << "<command><name>" << i->second->name << "</name>" << "<count>" << ConvToStr(i->second->use_count) << "</count></command>";
```
attilamolnar: `"</name>" << "<count>"` This can be together

attilamolnar: ConvToStr() is unnecessary

attilamolnar: Replace `count` with `usecount` that makes more sense

attilamolnar:
@johanna-a Hello and thanks for the contribution.

> This feels like it should be togglable by configuration though, rather than to be included by default. Or maybe it could be included on a substring match of the URI? (e.g. /stats/?includeCommandList)

Including by default is fine.

I have left some comments on the diff, once those are addressed I'll be happy to merge it.

attilamolnar:
Please write `usecount` not `use_count`, the other names don't have underscores either (`usercount`, `channelcount`, `opercount` ...) and also squash the commits so there's only one commit in the PR.

attilamolnar:
Thanks!

PR #1141, opened 2016-02-20
Purge code for Windows XP and MSVC pre-2015.

SaberUK:
I think it is safe to say that nobody should be using XP to host an IRC server, especially now that GnuTLS doesn't support XP anymore.

This pull request also bumps the minimum Visual Studio version to 2015 which removes a massive chunk of code now that Microsoft have decided to follow standards. This should not be a problem as the free community edition can be downloaded at https://www.visualstudio.com/en-us/downloads/download-visual-studio-vs.aspx

### DIFF
--- a/src/modules/m_dnsbl.cpp
+++ b/src/modules/m_dnsbl.cpp
@@ -88,7 +88,7 @@ class DNSBLResolver : public DNS::Request
 		bool match = false;
 		in_addr resultip;
 
-		inet_aton(ans_record->rdata.c_str(), &resultip);
+		inet_pton(AF_INET, ans_record->rdata.c_str(), &resultip);
 
 		switch (ConfEntry->type)
 		{
--- a/src/modules/m_spanningtree/main.cpp
+++ b/src/modules/m_spanningtree/main.cpp
@@ -209,7 +209,7 @@ void ModuleSpanningTree::ConnectServer(Link* x, Autoconnect* y)
 	else
 	{
 		in_addr n;
-		if (inet_aton(x->IPAddr.c_str(),&n) < 1)
+		if (inet_pton(AF_INET, x->IPAddr.c_str(),&n) < 1)
 			ipvalid = false;
 	}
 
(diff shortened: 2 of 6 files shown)


src/modules/m_dnsbl.cpp
```diff
@@ -78,7 +78,7 @@ class DNSBLResolver : public DNS::Request
 		bool match = false;
 		in_addr resultip;
 
-		inet_aton(ans_record.rdata.c_str(), &resultip);
+		inet_pton(AF_INET, ans_record.rdata.c_str(), &resultip);
```
SaberUK: This is fine here because inet_aton only worked with AF_INET sockets anyway. It is also deprecated for this reason.

src/modules/m_spanningtree/main.cpp
```diff
@@ -210,7 +210,7 @@ void ModuleSpanningTree::ConnectServer(Link* x, Autoconnect* y)
 	else
 	{
 		in_addr n;
-		if (inet_aton(x->IPAddr.c_str(),&n) < 1)
+		if (inet_pton(AF_INET, x->IPAddr.c_str(),&n) < 1)
```
SaberUK: The first arm of this if statement does the same as this only with an in6_addr and AF_INET6.

src/socket.cpp
```diff
@@ -146,13 +146,13 @@ std::string irc::sockets::sockaddrs::addr() const
 	char addrv[INET6_ADDRSTRLEN+1];
 	if (sa.sa_family == AF_INET)
 	{
-		if (!inet_ntop(AF_INET, &in4.sin_addr, addrv, sizeof(addrv)))
+		if (!inet_ntop(AF_INET, (void*)&in4.sin_addr, addrv, sizeof(addrv)))
```
SaberUK: Casting to void* here and elsewhere is necessary because inet_ntop takes a PVOID on Windows. This should not affect non-Windows behaviour.

attilamolnar:
I don't want to drop XP support because testing in an XP VM doesn't need gigabytes of RAM. I'm ok with bumping the minimum required VS version.

SaberUK:
Theres no point to this then as pretty much all of the compatibility code is required for XP.

SaberUK:
I'm reopening this as Adam has had to drop XP support in v2.0.24 (see the release notes for details).

PR #1144, opened 2016-02-20
Makefile restart target

johanna-a:
Having a restart target in the makefile allows for typing make -j4 restart for local testing instead of make -j4 install && ./inspircd restart.
I mostly made this to learn about the makefile building, but if it's useful then do include. :)


SaberUK:
I don't really see the point of this. Just set an alias in your shell.

PR #1147, opened 2016-02-26
[2.0] Fix GCC 6 warning about null checking this.

SaberUK:
As with 402a1bb010522a35600325c1a3084e092b40ca22 this is known to be undefined behaviour but changing it is too risky for the 2.0 branch.

### DIFF
--- a/src/configparser.cpp
+++ b/src/configparser.cpp
@@ -388,19 +388,20 @@ bool ParseStack::ParseExec(const std::string& name, int flags, const std::string
 	return ok;
 }
 
-bool ConfigTag::readString(const std::string& key, std::string& value, bool allow_lf)
-{
 #ifdef __clang__
 # pragma clang diagnostic push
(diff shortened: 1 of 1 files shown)


src/configparser.cpp
```diff
@@ -388,19 +388,20 @@ bool ParseStack::ParseExec(const std::string& name, int flags, const std::string
 	return ok;
 }
 
-bool ConfigTag::readString(const std::string& key, std::string& value, bool allow_lf)
-{
 #ifdef __clang__
```
SaberUK: I had to move the Clang checks here too because:

1. Clang also defines __GNUC__
2. Older GCCs don't allow diagnostics within functions (error: #pragma GCC diagnostic not allowed inside functions).

attilamolnar: It's ok, it's not needed in master anyway.

attilamolnar:
Thanks!

PR #1148, opened 2016-02-28
Minor spelling errors in m_spanningtree.so

guikcd:
[Lintian](https://lintian.debian.org/) Debian packaging helper tool [report](https://lintian.debian.org/full/pkg-irc-maintainers@lists.alioth.debian.org.html#inspircd) some minor spelling error for printed message in module m_spanningtree.so. It's minor, but i'll be glad if included in future 2.0.x to drop this packaging minor patches in future Debian packages. Thanks.


attilamolnar:
Thanks, the patch will be included in the next release.

PR #1149, opened 2016-02-28
Add support for SASL reauthentication.

SaberUK:


attilamolnar:
Since this PR is doing 2 things I'll comment on them separately.

Reauth:
I think only 3.2 clients should be allowed to reauth via SASL as it's not allowed per 3.1, any reason to do otherwise?

Not advertising sasl when a server is down:
There is already a config option for the target server, let's use that, not a nickname. The spanningtree events need to be improved first to fire for every server joining and splitting for that.

SaberUK:
>I think only 3.2 clients should be allowed to reauth via SASL as it's not allowed per 3.1, any reason to do otherwise?

Sure, I can change that.

>There is already a config option for the target server, let's use that, not a nickname. 

Services can be online without SASL though. Thats why Atheme has SaslServ and I think Adam said he was considering doing the same for Anope.

attilamolnar:
> Services can be online without SASL though. Thats why Atheme has SaslServ and I think Adam said he was considering doing the same for Anope.

If a services server is linked and it's configured to do sasl, it will continue doing sasl when it is online. If the services server is reconfigured at some point in time to no longer do sasl, InspIRCd's configuration should be updated at that time as well to disable sasl.

attilamolnar:
Moreover, Anope has no saslserv by default and since saslserv must be configured by adding a dummy saslserv bot manually, it can be forgotten about when disabling sasl in Anope.

SaberUK:
>>I think only 3.2 clients should be allowed to reauth via SASL as it's not allowed per 3.1, any reason to do otherwise?

>Sure, I can change that.

I just realised that theres actually no way to detect whether a client supports SASL-3.2 as you can support SASL-3.2 without CAP-3.2.

attilamolnar:
I don't think we should be concerned about clients supporting 3.2 sasl but not 3.2 cap. What do other impls do?

SaberUK:
>What do other impls do?

Charybdis and mammon (only other implementations with reauth support) behave the same as this pull request.

kaniini:
Charybdis lets any client do reauth, including 3.1 clients.

attilamolnar:
Ok, please rebase and I'll merge it like this.

SaberUK:
Done!

attilamolnar:
Thanks!

PR #1150, opened 2016-02-28
Adds classban module providing extban 'n' for connect class ban.

johanna-a:
This adds a module, connameban, that provides extban token 'n' that matches the mask against the users ->GetClass()->name.
This is not yet tested in production.

The IS_LOCAL() should work in all cases that are contained in master, since in those cases only local users are handled. Changing the signature of OnCheckBan to take a LocalUser* instead of a User* could make sense.

If not, I guess adding some sort of handling of the case where the user is not local would be necessary in this module, I'm not sure how to get that information across server nodes though.

### DIFF
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -467,6 +467,13 @@
 # This module is oper-only.
 # To use, CHGNAME must be in one of your oper class blocks.
 #<module name="chgname">
+#
+#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#
+# Connection class ban module: Adds support for extban 'n' which 
+# matches against the class name of the user's connection.
+# This module assumes that connection classes are named in a uniform
+# way on all servers of the network.
+#<module name="classban">
 
 #-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#
 # Clear chan module: Allows opers to masskick, masskill or mass-G/ZLine
--- a/src/modules/m_classban.cpp
+++ b/src/modules/m_classban.cpp
@@ -1,0 +1,47 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2016 Johanna Abrahamsson <johanna-a@mjao.org>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#include "inspircd.h"
+
+class ModuleClassBan : public Module
+{
+ public:
+	ModResult OnCheckBan(User* user, Channel* c, const std::string& mask) CXX11_OVERRIDE
+	{
+		LocalUser* localUser = IS_LOCAL(user);
+		if ((localUser) && (mask.length() > 2) && (mask[0] == 'n') && (mask[1] == ':'))
+		{
+			if (InspIRCd::Match(localUser->GetClass()->name, mask.substr(2)))
+				return MOD_RES_DENY;
+		}
+		return MOD_RES_PASSTHRU;
+	}
+
+	void On005Numeric(std::map<std::string, std::string>& tokens) CXX11_OVERRIDE
+	{
+		tokens["EXTBAN"].push_back('n');
+	}
+
+	Version GetVersion() CXX11_OVERRIDE
+	{
+		return Version("Class 'n' - Connection class ban", VF_VENDOR | VF_OPTCOMMON);
+	}
+};
+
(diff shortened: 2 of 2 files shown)


src/modules/m_connameban.cpp
```diff
@@ -0,0 +1,56 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#include "inspircd.h"
+
+class ModuleConnameBan : public Module, public Whois::EventListener
+{
+	public:
```
attilamolnar: Please only indent this with one space.

src/modules/m_connameban.cpp
```diff
@@ -0,0 +1,56 @@
...
+		Version GetVersion() CXX11_OVERRIDE
```
attilamolnar: Indent with one tab

src/modules/m_connameban.cpp
```diff
@@ -0,0 +1,56 @@
...
+			LocalUser* localUser = IS_LOCAL(user);
+			if (localUser && (mask.length() > 2) && (mask[0] == 'n') && (mask[1] == ':'))
+			{
+				if (InspIRCd::Match(localUser->GetClass()->name, mask.substr(2)))
+					return MOD_RES_DENY;
+			}
+			return MOD_RES_PASSTHRU;
+		}
+
+		void OnWhois(Whois::Context& whois) CXX11_OVERRIDE
```
attilamolnar: I don't think coupling the extban with a whois line is a great idea, let's keep this simple and only have the extban and we'll figure out how's best to show the connect class in whois.

src/modules/m_connameban.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+	public:
+		Version GetVersion() CXX11_OVERRIDE
+		{
+			return Version("Conname 'n' - Connection name (class) ban",VF_OPTCOMMON);
+		}
+
+		ModuleConnameBan() : Whois::EventListener(this) {};
+
+		ModResult OnCheckBan(User *user, Channel *c, const std::string& mask) CXX11_OVERRIDE
```
attilamolnar: The `*` (and `&`) should be next to the type, not the variable name.

src/modules/m_connameban.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+
+#include "inspircd.h"
+
+class ModuleConnameBan : public Module, public Whois::EventListener
+{
+	public:
+		Version GetVersion() CXX11_OVERRIDE
+		{
+			return Version("Conname 'n' - Connection name (class) ban",VF_OPTCOMMON);
```
attilamolnar: The flags should be `VF_VENDOR | VF_OPTCOMMON` as it's going to be part of the official distribution.

SaberUK:
This module needs to also be documented in docs/conf/modules.conf.example.

Bikeshed: I think this module would be better off named `m_classban` as that shows its intention better.

attilamolnar:
Thanks for this module, I've commented on the diff (mostly about the coding style).

I agree that the module should be renamed as most people know connect classes as connect classes.

johanna-a:
Thank you for your feedback :)
I've updated the PR and squashed the commits.
Since we're separating class info in whois from this module perhaps it should be included in core_whois? I guess it might be relevant information to opers, but probably not to regular users.

src/modules/m_classban.cpp
```diff
@@ -0,0 +1,49 @@
...
+	{
+		return Version("Class 'n' - Connection class ban", VF_VENDOR | VF_OPTCOMMON);
+	}
+
+	ModuleClassBan() {};
+
+	ModResult OnCheckBan(User* user, Channel* c, const std::string& mask) CXX11_OVERRIDE
+	{
+		LocalUser* localUser = IS_LOCAL(user);
+		if (localUser && (mask.length() > 2) && (mask[0] == 'n') && (mask[1] == ':'))
```
SaberUK: Bikeshed: 'c' might be a better extban character than 'n' here.

src/modules/m_classban.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+class ModuleClassBan : public Module
+{
+ public:
+	Version GetVersion() CXX11_OVERRIDE
+	{
+		return Version("Class 'n' - Connection class ban", VF_VENDOR | VF_OPTCOMMON);
+	}
+
+	ModuleClassBan() {};
```
attilamolnar: Empty, can remove

src/modules/m_classban.cpp
attilamolnar: Put the first check (localuser) in parenthesis as well to avoid inconsistency with the other checks in this if

src/modules/m_classban.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#include "inspircd.h"
+
+class ModuleClassBan : public Module
+{
+ public:
+	Version GetVersion() CXX11_OVERRIDE
```
attilamolnar: Please move this method to the bottom of the class to be consistent with the rest of the modules

docs/conf/modules.conf.example
```diff
@@ -464,6 +464,11 @@
 # This module is oper-only.
 # To use, CHGNAME must be in one of your oper class blocks.
 #<module name="chgname">
+#
+#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#
+# Connection class ban module: Adds support for extban 'n' which 
+# matches against the class name of the user's connection.
```
attilamolnar: I think we should mention that for this module to work properly on a network with multiple servers, all servers on the network should refer to the same class with the same name. For example, if there is a webirc class on one server and not on another, it won't work.

attilamolnar: What do you think? Feel free to suggest better wording.

attilamolnar:
Thanks!

PR #1154, opened 2016-03-08
[2.0] Purge the m_mssql module from extras.

### LINKED ISSUE #1006: [2.0] mssql module has not worked since SQLv2 was removed five years ago.
It was apparently never converted to SQLv3 in March 2010 when SQLv2 was removed. It also appears to be written against an old version of FreeTDS as neither the version in APT on Ubuntu 14.04 or in Homebrew on OSX have a `tds.h` header.

This means that **every single version of 2.0** has shipped with an uncompilable extras module. How the hell have we not had some kind of report about this in almost five years? I'm guessing nobody uses InspIRCd with Microsoft's SQL Server? Posting this so we can get feedback from Windows users.

---

FWIW I am in favour of killing the module on insp20 and migrating entirely to use ODBC for SQL in master. That way we won't have to maintain code for something which is already a solved problem.

SaberUK:
This module has not compiled since before 2.0 was released. We probably shouldn't be shipping broken modules.

I have looked into fixing this but it would pretty much require a complete rewrite with the SQLv3 API and against a newer version of FreeTDS.

Closes #1006.


attilamolnar:
See #1006.

PR #1159, opened 2016-03-22
Improvements to ./configure file generation.

SaberUK:
## Extract directory creation code to its own subroutine.

This is needed in a later commit so I extracted it to a helper method to reduce duplication.

## Fix a typo in parse_templates.

An error message was using the context variable rather than the target file variable.

## Write generated templates to the .configure directory.

Instead of polluting the root directory generated files that do not have a specific location to go in are now put in the .configure directory.

## Move the configure cache to the .configure directory too.

Same as the previous but for the configure cache file.

## Simplify the configure cache file format.

Previously the configuration file parser would break if a value contained a double quote symbol.  The format has been simplified to just `<name> [value...]` to avoid needing to escape these.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,17 +5,12 @@
 .*
 !.git*
 
+/.configure
 /BSDmakefile
 /GNUmakefile
 /build
 /docs/doxygen
-/inspircd
-/inspircd.1
-/inspircd-genssl.1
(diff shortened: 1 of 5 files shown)


attilamolnar:
Thanks!

PR #1162, opened 2016-03-25
[2.0] Delete modules cautiously when doing a deinstall.

SaberUK:
This will not affect 99% of users but its good to be sure just in case the user tries to dump module files into a global library directory.


attilamolnar:
Thanks!

PR #1167, opened 2016-03-31
Rename vague build environment variables to prevent collisions.

SaberUK:
This implements the idea I said I would implement in #1009 for the user facing variables. The rest will be changed at a later date.

Jenkins might need some updates after this, cc @tobgen.


attilamolnar:
Thanks!

PR #1169, opened 2016-04-10
[2.0] Fix extraneous quotes in makefile template.

SaberUK:
I have no idea how this happened considering that everything was green when I pushed it to Jenkins. I guess that Jenkins is just doing `make` instead of `make install`? Blah.

Thanks to @Robby- for reporting this.


attilamolnar:
Thanks, maybe it's time to integrate Travis and use that for testing next to jenkins.

SaberUK:
Good idea. I'll look into getting Jenkins set up.

PR #1170, opened 2016-04-10
Add configuration for building with Travis-CI.

SaberUK:
`travis-ci.sh` is only needed for insp20. On master the script can just use "./tools/test-build $CXX".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+compiler:
+  - "clang"
+  - "gcc"
+language: "cpp"
+notifications:
+  email: false
+os:
+  - "linux"
+  - "osx"
+script:
+  - "sh ./tools/travis-ci.sh"
+sudo: required
--- a/tools/travis-ci.sh
+++ b/tools/travis-ci.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -v
+if [ "$TRAVIS_OS_NAME" = "linux" ]
+then
+	sudo apt-get update --assume-yes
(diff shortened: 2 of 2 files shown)


SaberUK:
I can make this build modules too but i'm not sure which modules should be built. Opinions?

Sheogorath-SI:
I would say SSL/TLS modules, may m_geoip, all defaults and the regex modules.

Because those are the most widely used modules.

I would say SSL/TLS is a must openssl and gnutls.

SaberUK:
I've made it build all of the modules apart from m_regex_stdlib (which requires a C++11 compiler which I don't think Travis has without installing it manually) and m_mssql (which is completely broken, see #1154).

Sheogorath-SI:
To upgrade and use the all libs you can provide [travis_wait](https://docs.travis-ci.com/user/common-build-problems/) as prefix for the update command. I would also extract the dependencies from the build script.

May separate file for debian dependencies in a external script which can called by the build script but also by no name users to install dependencies for modules. Would be helpful for support.

SaberUK:
>To upgrade and use the all libs you can provide travis_wait as prefix for the update command.

We can't because thats only for things that don't provide output. We are apparently exceeding the time a Travis build can take overall.

>May separate file for debian dependencies in a external script which can called by the build script but also by no name users to install dependencies for modules. Would be helpful for support.

Some of the package names are different on different versions of Ubuntu so that wouldn't be very useful. I do have plans to integrate this into the build system (i.e. `./modulemanager install-dep ssl_gnutls`) in master once everything else is done though.

attilamolnar:
Thanks!

PR #1171, opened 2016-04-22
Add a method which calculates the maximum mask length.

SaberUK:
... and use it to fix topic setter truncation.


attilamolnar:
Thanks!

PR #1175, opened 2016-05-04
Stripcolor now strips from part messages

0x277F:
The stripcolor mode (+S) is designed to prevent spam of color codes inside channels. This works fine for messages, but does not affect part messages. This PR will filter color codes out of part messages on channels where +S is set.

### DIFF
--- a/src/modules/m_stripcolor.cpp
+++ b/src/modules/m_stripcolor.cpp
@@ -83,6 +83,23 @@ class ModuleStripColor : public Module
 		return MOD_RES_PASSTHRU;
 	}
 
+	void OnUserPart(Membership* memb, std::string& partmessage, CUList& except_list) CXX11_OVERRIDE
+	{
+		User* user = memb->user;
+		Channel* channel = memb->chan;
+
+		if (!IS_LOCAL(user))
+			return;
+
+		bool active = channel->GetExtBanStatus(user, 'S').check(!user->IsModeSet(csc))
+			&& ServerInstance->OnCheckExemption(user, channel, "stripcolor") != MOD_RES_ALLOW;
+
+		if (active)
+		{
+			InspIRCd::StripColor(partmessage);
+		}
+	}
+
 	Version GetVersion() CXX11_OVERRIDE
 	{
 		return Version("Provides channel +S mode (strip ansi color)", VF_VENDOR);


src/modules/m_stripcolor.cpp
```diff
@@ -83,6 +83,24 @@ class ModuleStripColor : public Module
...
 
+	void OnUserPart(Membership *memb, std::string &partmessage, CUList &except_list)
+	{
+		User *user = memb->user;
+		Channel *channel = memb->chan;
+
+		if (!IS_LOCAL(user))
+			return;
+
+		bool active = (user->IsModeSet(usc) || channel->GetExtBanStatus(user, 'S').check(!user->IsModeSet(csc)))
```
attilamolnar: Checking `user->IsModeSet()` here is undesirable. It will prevent any user with the stripcolor mode set to have colors in their part message, even on channels that doesn't have the stripcolor mode set.

src/modules/m_stripcolor.cpp
```diff
@@ -83,6 +83,24 @@ class ModuleStripColor : public Module
 		return MOD_RES_PASSTHRU;
 	}
 
+	void OnUserPart(Membership *memb, std::string &partmessage, CUList &except_list)
```
attilamolnar: The `*` and `&` should be next to the variable type.

0x277F:
Updated as per @attilamolnar's requests.

src/modules/m_stripcolor.cpp
```diff
@@ -83,6 +83,23 @@ class ModuleStripColor : public Module
...
+	void OnUserPart(Membership* memb, std::string& partmessage, CUList& except_list)
+	{
+		User* user = memb->user;
+		Channel* channel = memb->chan;
+
+		if (!IS_LOCAL(user))
+			return;
+
+		bool active = channel->GetExtBanStatus(user, 'S').check(!user->IsModeSet(csc))
+			&& !(ServerInstance->OnCheckExemption(user, channel, "stripcolor") == MOD_RES_ALLOW);
```
SaberUK: Is there any reason you can't do `!= MOD_RES_ALLOW` here?

0x277F:
Fixed @SaberUK's issue, I wasn't really thinking about it very clearly when I wrote it.

src/modules/m_stripcolor.cpp
```diff
@@ -83,6 +83,23 @@ class ModuleStripColor : public Module
 		return MOD_RES_PASSTHRU;
 	}
 
+	void OnUserPart(Membership* memb, std::string& partmessage, CUList& except_list)
```
Adam-: CXX11_OVERRIDE

0x277F:
Fixed @Adam-'s suggestion.

attilamolnar:
Thanks and sorry for the delay!

PR #1176, opened 2016-05-07
Replace snprintf usage with InspIRCd::Format.

SaberUK:
Now all formatting code goes through InspIRCd::Format and doesn't mess around with C arrays.

### DIFF
--- a/src/coremods/core_stats.cpp
+++ b/src/coremods/core_stats.cpp
@@ -232,17 +232,11 @@ void CommandStats::DoStats(Stats::Context& stats)
 			stats.AddRow(249, "Commands: "+ConvToStr(ServerInstance->Parser.GetCommands().size()));
 
 			float kbitpersec_in, kbitpersec_out, kbitpersec_total;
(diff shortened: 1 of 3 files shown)


SaberUK:
This also inadvertently fixes a build error on Windows reported by MY123 in #inspircd.

attilamolnar:
For now please only switch core_stats to `InspIRCd::Format()` because the function returns a `std::string` causing extra memory allocation/deallocation that doesn't happen with plain `snprintf`. This doesn't affect core_stats but does affect every other call site modified by this PR, those don't need the `std::string`. If we add a new `Format()` function that returns a `const char*` then the rest of the code can be switched (and the `#define` in inspircd_win32wrapper.h can be dropped).

SaberUK:
@attilamolnar I reverted some of the changes but I don't think the cloaking and map changes are allocating any more memory than they would have previously.

attilamolnar:
All calls to `Format()` cause a temporary string to be created even if that string is appended to another string immediately, but we do get simpler code. Either way, a new `Format()` function I mentioned in the previous comment will get rid of the temporary strings in those cases.

attilamolnar:
Thanks!

PR #1178, opened 2016-05-28
fix m_customprefix modes deleting RFC-defined chanmodes from 005 PREFIX

jokleinn:
This pull request will help to idiot-proof (I fell victim to this) the configuration for m_customprefix a little bit, as defining a custom mode with a rank equal to VOICE_VALUE or OP_VALUE results in +v or +o's respective entry in the 005 PREFIX response being removed altogether, which causes various bugs in some clients, ie minor annoyances.

### DIFF
--- a/src/modules/m_customprefix.cpp
+++ b/src/modules/m_customprefix.cpp
@@ -17,6 +17,7 @@
  */
 
 
+#include "mode.h"
 #include "inspircd.h"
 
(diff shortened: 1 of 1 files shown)


SaberUK:
This looks like it fixes a symptom rather than the actual bug. I suspect the actual bug lies elsewhere.

attilamolnar:
Closing in favor of #1179 which fixes the actual bug.

PR #1179, opened 2016-05-29
[2.0] Fix 005 not showing some modes with the same rank as others.

SaberUK:
If two modes with the same prefix rank are loaded (e.g. via `m_customprefix`) then only the last loaded mode would be shown in the PREFIX 005 token.

This fixes the underlying bug which #1178 worked around.

### DIFF
--- a/src/mode.cpp
+++ b/src/mode.cpp
@@ -837,11 +837,19 @@ std::string ModeParser::GiveModeList(ModeMasks m)
 	return type1 + "," + type2 + "," + type3 + "," + type4;
 }
 
+struct PrefixModeSorter
+{
+	bool operator()(ModeHandler* lhs, ModeHandler* rhs)
+	{
+		return lhs->GetPrefixRank() < rhs->GetPrefixRank();
+	}
+};
+
 std::string ModeParser::BuildPrefixes(bool lettersAndModes)
 {
 	std::string mletters;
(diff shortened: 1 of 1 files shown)


src/mode.cpp
```diff
@@ -837,27 +837,35 @@ std::string ModeParser::GiveModeList(ModeMasks m)
...
 		{
-			prefixes[modehandlers[pos]->GetPrefixRank()] = std::make_pair(
-				modehandlers[pos]->GetPrefix(), modehandlers[pos]->GetModeChar());
+			prefixes.push_back(modehandlers[pos]);
 		}
 	}
 
-	for(std::map<int,std::pair<char,char> >::reverse_iterator n = prefixes.rbegin(); n != prefixes.rend(); n++)
+	std::sort(prefixes.begin(), prefixes.end(), PrefixModeSorter());
+	for(std::vector<ModeHandler*>::reverse_iterator n = prefixes.rbegin(); n != prefixes.rend(); n++)
```
attilamolnar: Please make the following changes:
* add space between `for` and `(`
* const_reverse_iterator
* `++n`

SaberUK:
@attilamolnar I have made the changes you requested.

attilamolnar:
Thanks, nice fix!

PR #1181, opened 2016-06-06
[2.0] GCC should never assume that this can not be null.

SaberUK:
Whilst this is undefined behaviour fixing all cases of this are too invasive for 2.0.

### DIFF
--- a/configure
+++ b/configure
@@ -965,6 +965,9 @@ EOF
 	my @dotfiles = qw(main.mk inspircd);
 	push @dotfiles, 'org.inspircd.plist' if $config{OSNAME} eq 'darwin';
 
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -965,6 +965,9 @@ EOF
 	my @dotfiles = qw(main.mk inspircd);
 	push @dotfiles, 'org.inspircd.plist' if $config{OSNAME} eq 'darwin';
 
+	# HACK: we need to know if we are on GCC6 to disable null pointer checks.
```
attilamolnar: Should say
> ... to enable null checking of `this`.

or similar

attilamolnar:
Thanks!

PR #1182, opened 2016-06-22
Fix sending the wrong numeric when an empty new nick is received.

SaberUK:
This fixes InspIRCd to work like section 4.1.2 of RFC 1459 dictates.

Filed against master so we don't break anything relying on the old behaviour..


attilamolnar:
Thanks!

PR #1184, opened 2016-06-26
Improvements to flood modules

SaberUK:
- Allow use of human readable durations (e.g. 1m30) in flood modules.
- Make the duration of nickflood and joinflood configurable.

### DIFF
--- a/src/modules/m_joinflood.cpp
+++ b/src/modules/m_joinflood.cpp
@@ -23,6 +23,9 @@
 
 #include "inspircd.h"
 
+// The number of seconds the channel will be closed for.
+static unsigned int duration;
+
 /** Holds settings and state associated with channel mode +j
  */
 class joinfloodsettings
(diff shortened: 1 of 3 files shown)


src/modules/m_nickflood.cpp
```diff
@@ -20,6 +20,9 @@
 
 #include "inspircd.h"
 
+// The number of seconds nickname changing will be blocked for.
+static int duration;
```
attilamolnar: This should be an `unsigned int`

src/modules/m_joinflood.cpp
```diff
@@ -23,6 +23,9 @@
 
 #include "inspircd.h"
 
+// The number of seconds the channel will be closed for.
+static int duration;
```
attilamolnar: `unsigned int`

SaberUK:
@attilamolnar Done!

attilamolnar:
Changing the meaning of the parameter isn't possible when linking to 2.0 servers as they'll misinterpret it ("1m" -> "1" = 1 second), please revert that for now until we figure out how best to handle introducing new features that break 2.0 linking.

SaberUK:
Perhaps we should add a global `<config compat="2.0 | 3.0 | current">` switch?

SaberUK:
@attilamolnar I removed that for now. I'll look into adding the idea I suggested. ^

attilamolnar:
Thanks!

Issue #1177, opened 2016-05-10
lockserv module: ability to specify a custom message and have the commands work remotely


Robby-:
It would be nice if the `LOCKSERV` command accepted an optional custom message, for example:
`LOCKSERV Please use irc.example.tld as this server is getting decommissioned soon.`
Or even via a config option in `modules.conf`. As currently it is needed to hard-code and recompile the module with a custom message.

Another feature I think would be nice to have is if the commands `LOCKSERV` and `UNLOCKSERV` worked remotely as well, although I'm not sure how that would work if the above feature request were to be implemented. As the command syntax would need a change in that case. Maybe it could work like this for example, with a message:
`LOCKSERV server.example.tld This server is scheduled for maintenance, use irc.example.tld.`
Without a message: `LOCKSERV server.example.tld`
And subsequently: `UNLOCKSERV server.example.tld`.
The commands would have to have some logic to determine if a server name was specified or not as the first parameter.

SaberUK:
@Robby- How about this? https://github.com/inspircd/inspircd/compare/master...SaberUK:master+lockserv

Robby-:
I can't actually test it due to not having a master branch server set up yet.

attilamolnar:
Custom reason has been implemented by @SaberUK in #1185. Remote unlock can be implemented via a special rehash flag for example.

PR #1185, opened 2016-06-26
Allow the reason for locking to be specified in m_lockserv.

SaberUK:
Reference: #1177

### DIFF
--- a/src/modules/m_lockserv.cpp
+++ b/src/modules/m_lockserv.cpp
@@ -27,23 +27,24 @@
 
 class CommandLockserv : public Command
 {
-	bool& locked;
+	std::string& locked;
 
  public:
-	CommandLockserv(Module* Creator, bool& lock) : Command(Creator, "LOCKSERV", 0), locked(lock)
+	CommandLockserv(Module* Creator, std::string& lock) : Command(Creator, "LOCKSERV", 0, 1), locked(lock)
 	{
+		allow_empty_last_param = false;
 		flags_needed = 'o';
 	}
 
 	CmdResult Handle (const std::vector<std::string> &parameters, User *user)
 	{
-		if (locked)
+		if (!locked.empty())
 		{
 			user->WriteNotice("The server is already locked.");
 			return CMD_FAILURE;
 		}
 
-		locked = true;
+		locked = parameters.empty() ? "Server is temporarily closed. Please try again later." : parameters[0];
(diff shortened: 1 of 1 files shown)


src/modules/m_lockserv.cpp
```diff
@@ -86,31 +86,26 @@ class ModuleLockserv : public Module
...
+			ServerInstance->Users->QuitUser(user, locked);
 			return MOD_RES_DENY;
 		}
 		return MOD_RES_PASSTHRU;
 	}
 
 	ModResult OnCheckReady(LocalUser* user) CXX11_OVERRIDE
 	{
-		return locked ? MOD_RES_DENY : MOD_RES_PASSTHRU;
+		return locked.size() ? MOD_RES_DENY : MOD_RES_PASSTHRU;
```
attilamolnar: `!locked.empty()`

src/modules/m_lockserv.cpp
```diff
@@ -52,23 +52,23 @@ class CommandLockserv : public Command
...
 
  public:
-	CommandUnlockserv(Module* Creator, bool &lock) : Command(Creator, "UNLOCKSERV", 0), locked(lock)
+	CommandUnlockserv(Module* Creator, std::string& lock) : Command(Creator, "UNLOCKSERV", 0, 1), locked(lock)
 	{
 		flags_needed = 'o';
 	}
 
-	CmdResult Handle (const std::vector<std::string> &parameters, User *user)
+	CmdResult Handle(const std::vector<std::string>& parameters, User* user)
```
attilamolnar: Please fix the code style in a separate commit

attilamolnar: (if you fix at all)

src/modules/m_lockserv.cpp
```diff
@@ -27,23 +27,23 @@
...
 	{
-		if (locked)
+		if (!locked.empty())
 		{
 			user->WriteNotice("The server is already locked.");
 			return CMD_FAILURE;
 		}
 
-		locked = true;
+		locked = parameters.empty() || parameters[0].empty() ? "Server is temporarily closed. Please try again later." : parameters[0];
```
attilamolnar: Condition should be in ()

SaberUK: I switched to using `allow_empty_last_param = false` so the second check is unnecessary

attilamolnar:
Thanks!

PR #1186, opened 2016-06-27
Pass INSPIRCD_VERBOSE through the environment to unit-cc.

SaberUK:
Also, fix a minor cleanup bug in the makefile.


attilamolnar:
Thanks!

PR #1191, opened 2016-07-28
[2.0] Force link sqlite3 on OS X when doing Travis builds.

SaberUK:
The system version of this library does not include support for pkg-config.


attilamolnar:
Thanks!

PR #1196, opened 2016-08-05
[2.0] Fix challenge auth when using m_hash_gnutls instead of m_sha256.

SaberUK:


Techman-:
I tested this, and works as intended. :+1:

attilamolnar:
Thanks!

Issue #1198, opened 2016-08-08
SASL vhost improvements to consider for 2.2


Techman-:
As it stands in InspIRCd 2.0, services rely on CHGHOST in order to set a user's vhost during SASL authentication, compared to [other IRCds that use SVSLOGIN](https://github.com/grawity/irc-docs/blob/master/server/ts6.txt#L898-L914). It's a separate command that has to be sent from a services server, but this in itself isn't an issue. I mention SVSLOGIN as a potential solution to be discussed further below.

CHGHOST doesn't quite do the job in 2.0 for SASL purposes, for a few reasons. I'll give a case for Anope and a case for Atheme, based on how they behave as of this issue.
## Anope

In Anope, if a user has a vhost tied to their account, it will send `CHGHOST` as soon as it sees the `UID` command from the uplink. This is currently the best way to handle setting vhosts, because sending it before SASL success won't actually stick post-registration if a network has +x as an on-connect umode for that connect class.
## Atheme

In Atheme, [it will send](https://github.com/atheme/atheme/blob/36f1b1a3712058639cac34a2ef51296eb5dfbcb0/modules/protocol/inspircd.c#L721-L728) (ENCAP'd) `CHGHOST` if there's a vhost tied to the user's account. You'll see it during your SASL authentication session, but it won't stick after IRCd registration if you match a connect class that has +x by default. You'll send up with a hashed hostname anyways, which Atheme detects and has to send your vhost _again_, which is a problem because **users could already be in a channel**, which dampens the usability of vhosts as it adds extra cyclehost spam.

Here's some client logs that help show what I'm speaking about:

> 02:41:00 AM    <-->    irc: client capability, enabled: away-notify extended-join account-notify multi-prefix sasl userhost-in-names
> 02:41:00 AM    <-->    You are now logged in as Techman (Techman!Techman@clank.techtronix.net)
> 02:41:00 AM    <-->    techmansworld.com :is now your displayed host
> 02:41:00 AM    <-->    SASL authentication successful
> (...)
> 02:41:03 AM    <-->    allrednc-76oskp.techtronix.net :is now your displayed host
> 02:41:03 AM    <-->    User mode [+ixR] by Techman
> (...)
> 02:41:04 AM    <-->    User mode [-x] by Tetris.AllredNC.org
> 02:41:04 AM    <-->    techmansworld.com :is now your displayed host
## Solutions

[...]

PR #1199, opened 2016-08-12
Modify m_chgident to allow ULines to set ident of unregistered users

Techman-:
Addresses some stuff relating to #1198


Techman-:
Adam had a better WIP, whether he thinks it was a hack job or not :)

PR #1208, opened 2016-09-04
m_sasl: Send a response to the client when a malformed AUTHENTICATE is received

Techman-:


SaberUK:
Why is this needed?

Techman-:
Because otherwise InspIRCd silently rejects the message without notifying the client of anything.

SaberUK:
I know. Why is that bad? This is not something a client will ever encounter. It will only ever be encountered by an attacker.

Techman-:
Well for one, if an attacker is trying to use this I would go ahead and kill the socket anyways. The other reason is because Syzop wrote a PHP script to quickly test if an IRCd is vulnerable, and InspIRCd sending nothing at all, not even 904, isn't very reliable.

attilamolnar:
Quitting the user is unnecessary, we never do that for malformed commands only reject the command in question which is what the code currently does.

PR #1215, opened 2016-09-09
Strip carriage returns in the config parser.

SaberUK:


attilamolnar:
I don't know if this is the best solution because \r\n will turn into two spaces which might be undesirable. What do you think?

SaberUK:
@attilamolnar I don't think its too much of a problem but I could change it to not do that if you think it is.

SaberUK:
@attilamolnar I have updated the parser to skip carriage returns which should solve this in a more elegant way.

attilamolnar:
Thanks!

PR #1217, opened 2016-09-11
Undocument <power> and set the default to the server name.

SaberUK:
We could alternately make it an oper priv to die without the password. What do you think @attilamolnar?

### DIFF
--- a/docs/conf/helpop-full.conf.example
+++ b/docs/conf/helpop-full.conf.example
@@ -621,17 +621,15 @@ The duration may be specified in seconds, or in the format
 1y2w3d4h5m6s - meaning one year, two weeks, three days, 4 hours,
 5 minutes and 6 seconds. All fields in this format are optional.">
 
-<helpop key="die" value="/DIE <password>
+<helpop key="die" value="/DIE <server>
 
 This command shuts down the local server. A single parameter is
-required, which must match the password in the configuration for the
-command to function.">
+required, which must match the name of the local server.">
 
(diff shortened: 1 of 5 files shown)


attilamolnar:
I think this is fine the way it is. These commands can be disabled currently by not allowing any oper to use them (`<class:commands>`). The password nowadays is a failsafe mechanism so opers who do have access to these commands won't accidently stop the server. Having no password simply removes this failsafe mechanism for those who don't like it.

Adam-:
Many other ircds use /die server.name with an oper priv, which I think is what this should do.

SaberUK:
@attilamolnar As is you can't just use /DIE without a password. You have to explicitly specify a blank password parameter (i.e. `DIE :`). If we are going to make it intentional that opers can use /DIE without a password then we should make it so the parameter is optional.

attilamolnar:
What do you think of removing `<power>` from the 3.0 example config and passing the local server name as the default value to `getString()`? This way `<power>` blocks in 2.0 configs will behave as in 2.0 and 3.0 will have the behavior @Adam- said.

SaberUK:
@attilamolnar I can do that yeah.

attilamolnar:
:+1:

SaberUK:
@attilamolnar Like this?

attilamolnar:
Please update the helpop text as well.

SaberUK:
@attilamolnar Done!

attilamolnar:
Thanks!

PR #1219, opened 2016-09-13
Rewrite the build system directive parser.

SaberUK:
New features:

- Users will now be told to install the packages if a module with a $PackageInfo directive fails to build.
- Users who have found a bug will now be told to either contact the author of the module for a third party module or to file an issue for an official module.

Screenshot:

---

TODO: Investigate why Travis builds are failing on OSX (I can't replicate it locally, I think the Travis gnutls package is broken somehow).


SaberUK:
This should be done now. OS X Travis builds are broken because of a known Travis bug though (this does not happen locally). I have disabled OS X builds in #1223 because the infrastructure is just so unreliable.

attilamolnar:
Thanks!

PR #1221, opened 2016-09-14
Replace the deprecated MAXBANS token with MAXLIST.

SaberUK:


attilamolnar:
I'm fine with replacing the deprecated token with the new one, but I'd prefer it to be implemented in e.g. core_info because there is no reason for all this logic to be in the core.

SaberUK:
@attilamolnar I moved everything into core_info other than `GetLowerLimit()` which needs access to ListModeBase internals.

SaberUK:
Closing and reopening to force a Travis rebuild.

PR #1222, opened 2016-09-14
Fix various compiler warnings

SaberUK:
1. The latest version of libGeoIP has commas at the end of an enumerator list which causes warnings as we are still using C++03.

2. OS X Yosemite has deprecated OpenLDAP in favour of Open Directory but like with OpenSSL it still works fine and is unlikely to be removed soon.


attilamolnar:
Thanks!

PR #1223, opened 2016-09-15
Speed up Travis builds.

SaberUK:
- Switch Linux builds to Ubuntu 14.04 from 12.04
- Split dynamic and static builds into different jobs.
- Remove OS X builds as the infrastructure is slow and unreliable.

---

The Linux build time has now dropped to 7m30~ for Clang and 10m30~ for GCC. This is a time saving of just over 50%.

I have looked into precompiled headers but I don't think its possible with how flaky calcdep/unit-cc are at the moment (fixing this is work in progress).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,13 @@
 compiler:
-  - "clang"
-  - "gcc"
-language: "cpp"
+  - clang
+  - gcc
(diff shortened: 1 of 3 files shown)


attilamolnar:
Thanks!

PR #1224, opened 2016-09-15
WIP: Unicode nickname module

Googolplexed:
WIP (Particularly around max nick lengths)

### DIFF
--- a/src/modules/extra/m_utf8nick.cpp
+++ b/src/modules/extra/m_utf8nick.cpp
@@ -50,10 +50,10 @@ public:
     //Check if Nick is valid using the current codepage
     bool Call(const std::string& nick)
     {
-        ServerInstance->Logs->Log(MODNAME, LOG_DEBUG, "Checking nick %s validity",nick.c_str());
+		ServerInstance->Logs->Log(MODNAME, LOG_DEBUG, "Checking nick \"%s\" validity", nick.c_str());
 
         //Nick length should be handled as per standard ASCII ? Not sure if this should be done pre or post conversion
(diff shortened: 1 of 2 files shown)


extras/m_utf8nick.cpp
```diff
@@ -0,0 +1,160 @@
...
+        {
+            return false;
+        }
+
+        //Invalid input is replaced with U+FFFD
+        UChar unicodeString[NICKBUFFERSIZE];
+        int32_t outputLength;
+        UErrorCode err = U_ZERO_ERROR;
+        u_strFromUTF8(unicodeString,NICKBUFFERSIZE,&outputLength,nick.c_str(),-1,&err);
+        if(outputLength >= 128 || U_FAILURE(err))
```
Googolplexed: outputLength >= 128... can't even remember why I chose that constant.

src/modules/extra/m_utf8nick.cpp
```diff
@@ -0,0 +1,160 @@
...
+        {
+            return false;
+        }
+
+        //Invalid input is replaced with U+FFFD
+        UChar unicodeString[NICKBUFFERSIZE];
+        int32_t outputLength;
+        UErrorCode err = U_ZERO_ERROR;
+        u_strFromUTF8(unicodeString,NICKBUFFERSIZE,&outputLength,nick.c_str(),-1,&err);
+        if(outputLength >= 128 || U_FAILURE(err))
```
attilamolnar: Using an uninitialized `outputLength` unless `u_strFromUTF8` guarantees that it writes a value to it in all cases

PR #1226, opened 2016-09-20
Fix compatibility with the legacy <channels> tag.

SaberUK:
We should only use OperMaxChans if they have the right privilege.


attilamolnar:
Thanks!

PR #1227, opened 2016-09-20
Add a way of saying that 2.0 feature compatibility is required.

SaberUK:
This allows us to enable things without breaking linking compatibility with 2.0.

I have ported #451 which was waiting for this. If there are any more things I can port to use this then stick them in a reply and i'll add them.

### DIFF
--- a/docs/conf/inspircd.conf.example
+++ b/docs/conf/inspircd.conf.example
@@ -537,6 +537,10 @@
          # to this value.
          #fixedpart=""
 
+         # legacymode: If enabled then new features which might conflict when linking to
+         # older servers are disabled by default.
+         legacymode="no"
+
          # syntaxhints: If enabled, if a user fails to send the correct parameters
          # for a command, the ircd will give back some help text of what
          # the correct parameters are.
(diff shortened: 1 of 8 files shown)


attilamolnar:
Suppose a 2.0 network with multiple hub and leaf servers wants to upgrade to 3.0, upgrading servers one by one. Can you explain how will the upgrade work with this?

SaberUK:
With this enabled they would need to split servers off one by one and link the upgraded servers together as a temporarily independent network.  Its unfortunately very disruptive but thats the only real way I can see to do this kind of compatibility-breaking changes short of us implementing some kind of magic S2S feature matching.

rburchell:
I think that keeping compatibility at link time is something that we should aim to do, at least transitionally (that is: I don't see a point in say, 3.0 being able to link to 1.2 now. Compatibility is fine to be removed after a grace period of a few years.)

Feature negotiation would be a nice thing to consider, though I've always been a bit wary of that because it can lead to a forest of codepaths that aren't often (or always) tested, and for the most part we ended up getting around it by instead using things like ENCAP. What sort of features would warrant that sort of treatment?

PR #1228, opened 2016-09-21
Fix the directive parser using invalid environment variables.

SaberUK:
This fixes an oversight where in cases like #1224 you could not set the environment variables that the directive parser is looking for because of invalid characters.


attilamolnar:
Thanks!

PR #1233, opened 2016-10-09
m_httpd_stats: Also sanitize other server gecos

Natolumin:
The description of the local server is ``Sanitize()``'d before being printed, but those of other servers are not.
When the remote server uses a non-UTF-8 charset in its description, this causes the generated XML file to be invalid.


attilamolnar:
Thanks!

PR #1234, opened 2016-10-13
Rework the example services aliases.

SaberUK:
# Rework the example services aliases.

- Split common stuff into a generic configuration file.
- Reserve all nicknames used by services by default.
- Exempt all services clients from filtering.
- Add missing aliases for services pseudoclients.


attilamolnar:
The reworked service aliases are ok. I don't want to remove the old example until we have something else that works out of the box.

SaberUK:
Sure, i'll remove them for now and look into reworking them later.

SaberUK:
@attilamolnar Done!

docs/conf/inspircd.conf.example
```diff
@@ -864,14 +864,10 @@
 
 <badnick
          # nick: Nick to disallow. Wildcards are supported.
-         nick="ChanServ"
+         nick="Kyubey"
```
attilamolnar: Can you change this back to what it was? I understand that it'll be a duplicate because ChanServ is in generic.conf.example but duplicates are silently ignored and that way there's no chance that a perfectly okay nick will get blocked due to people being lazy to edit this.

SaberUK: Done.

attilamolnar:
Great, thanks!

PR #1236, opened 2016-10-22
Allow classes to take a port range.

SaberUK:


attilamolnar:
Can you make it so the ports are parsed on rehash and only a container lookup occurs SetClass()?

src/users.cpp
```diff
@@ -1096,14 +1096,14 @@ void LocalUser::SetClass(const std::string &explicit_name)
...
 			/* if it requires a port ... */
-			int port = c->config->getInt("port");
-			if (port)
+			if (!c->ports.empty())
 			{
-				ServerInstance->Logs->Log("CONNECTCLASS", LOG_DEBUG, "Requires port (%d)", port);
-
 				/* and our port doesn't match, fail. */
-				if (this->GetServerPort() != port)
+				if (!stdalgo::isin(c->ports, this->GetServerPort()))
```
attilamolnar: Use `.count()` to check if the value is in the set, just like with `std::set`

src/configreader.cpp
```diff
@@ -328,6 +328,16 @@ void ServerConfig::CrossCheckConnectBlocks(ServerConfig* current)
 			me->limit = tag->getInt("limit", me->limit);
 			me->resolvehostnames = tag->getBool("resolvehostnames", me->resolvehostnames);
 
+			std::string ports = tag->getString("port");
+			if (!ports.empty())
+			{
+				irc::portparser portrange(ports, false);
+				while (int port = portrange.GetToken())
+				{
```
attilamolnar: The loop body is a single line, no need for braces

attilamolnar:
Great job, thanks!

PR #1238, opened 2016-10-24
Fix OpenSSL build errors caused by BIO/BIO_METHOD now being opaque.

SaberUK:
OpenSSL is developed by incompetent people who have never heard of deprecating something before removing it. That is all.

### DIFF
--- a/src/modules/extra/m_ssl_openssl.cpp
+++ b/src/modules/extra/m_ssl_openssl.cpp
@@ -58,6 +58,15 @@
 #define INSPIRCD_OPENSSL_ENABLE_ECDH
 #endif
 
+// BIO is opaque in OpenSSL 1.1 but the access API does not exist in 1.0 and older.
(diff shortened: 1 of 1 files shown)


src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -409,6 +434,9 @@ static BIO_METHOD biomethods =
 	OpenSSL::BIOMethod::destroy, // destroy, does nothing, see function body for more info
 	NULL // callback_ctrl
 };
+#else
+static BIO_METHOD* biomethods = OpenSSL::BIOMethod::alloc();
```
attilamolnar: I'm not sure if this is correct, it gets executed before the ModuleOpenSSL's constructor which is where openssl is initialized. Move it to the constructor after library initialization, unless it's documented to work like this.

SaberUK:
I moved its initialisation to the module constructor.

attilamolnar:
Thanks!

PR #1245, opened 2016-11-08
[2.0] Workaround modern Perls not adding . to @INC.

SaberUK:
For 3.0 I will work on a better fix for this but for 2.0 this just restores the old behaviour.

Thanks to @grawity for pointing this out.


attilamolnar:
Thanks!

PR #1248, opened 2016-11-12
Allow listeners to be able to listen on unavailable addresses.

SaberUK:


attilamolnar:
Looks ok, are you sure that `IP_FREEBIND`, `IP_BINDANY` and `SO_BINDANY` work the same on these platforms and none of them have some unique behavior we need to document/be aware of? If you aren't I'll read the man pages before merging. Has this been tested?

SaberUK:
>Looks ok, are you sure that IP_FREEBIND, IP_BINDANY and SO_BINDANY work the same on these platforms and none of them have some unique behavior we need to document/be aware of?

From my understanding from reading the docs and other software which implements this they all work the same way as far as the application is concerned.


>Has this been tested?

Yes.

attilamolnar:
Thanks!

PR #1253, opened 2016-11-18
[2.0] Add a bigger warning about deprecated things being removed.

SaberUK:
This makes it clearer to server admins that they shouldn't use m_chanprotect, m_halfop and the compat- modes in m_cloaking.


attilamolnar:
Thanks!

PR #1255, opened 2016-11-22
2.0: Fixup m_timedbans

genius3000:
Noticed a few issues (or inconsistencies) in the timedbans module while fixing the statusmsg issue.

- The ban adding now sends to remote servers as well, both notices use a new string with "*** " pre-pended to the original message.
- Added the mask to the expiry message
- Copied the halfop check code from adding to expiry, made the notices use that prefix character instead of just '@'.


src/modules/m_timedbans.cpp
```diff
@@ -73,7 +73,7 @@ class CommandTban : public Command
 			user->WriteNumeric(482, "%s %s :You do not have permission to set bans on this channel",
 				user->nick.c_str(), channel->name.c_str());
 			return CMD_FAILURE;
-		}		
+		}
```
attilamolnar: Please revert this change or do whitespace related changes in a separate commit

attilamolnar:
I think the reason the ban mask is not in the NOTICE is because it can be seen in the MODE which unsets it. I'd rather not want that changed unless there is a good reason to (if so please elaborate).

genius3000:
@attilamolnar: Reverted the whitespace removal and as for adding the ban mask to the NOTICE, my thoughts were:
1) It was already in the NOTICE on addition, so simply to keep to the same "structure"
2) The added convenience to those that have all NOTICEs forced to their current or status window, saves them from having to check the channel just to see the ban mask.

genius3000:
@attilamolnar thoughts on this?
I'm not against reverting the change to the message(s), the main fixes here are the sending to remote servers on addition and checking for halfop on expiry. Would be nice to have both messages "standardized" though.

genius3000:
@SaberUK rebased, removed any modifications to the message strings, and updated the commit message.

PR #1262, opened 2016-12-02
[2.0] Backport the master Travis changes to insp20.

SaberUK:
Also, fix pure static builds which were apparently broken on GCC but nobody noticed because nobody uses them (I still think we should remove support for them).


attilamolnar:
Thanks!

PR #1264, opened 2016-12-05
Updated m_filter.cpp to add alerts on g-line/kill

ThatOneRoadie:
Added Snomask alert on same flag as block/notify, to notify on kill/g-line with clarified message for each.


SaberUK:

PR #1267, opened 2016-12-08
Store config values in a map instead of a unique vector of pairs.

SaberUK:
… and now the config is stored in a sane way make the config system case insensitive


rburchell:
👍 looks fine

PR #1269, opened 2016-12-20
m_sasl: send host/ip info

Adam-:


attilamolnar:
Is this confirmed to not break any anope and atheme versions which are currently working with inspircd?

Adam-:
Anope supports this fine.

attilamolnar:
Ok, I need a confirmation that this works on atheme too before merging

Robby-:
I've been testing this with Anope and this works fine for regular IRC clients. However, I also use m_cgiirc because I'm running a KiwiIRC-based webchat. When a webchat user connects, the IRCd will send the address of the webserver the webchat is running on, instead of sending the host/IP address provided in WEBIRC (which is the very first thing KiwiIRC sends).

grawity:
@attilamolnar Atheme has supported this since 7.2.7; older versions will just ignore it as an unrecognized subcommand.

Robby-:
This reminds me I needed to poke @attilamolnar as well, regarding the bug with m_cgiirc when using this PR, which I forgot, so here we go.

attilamolnar:
@Adam- can you check the issue mentioned by @Robby-?

Adam-:
@attilamolnar done.

Robby-:
I have been running the updated PR for about 24 hours now and can confirm this has fixed the issue. Thanks @Adam-.

attilamolnar:
Thanks!

PR #1270, opened 2016-12-22
Always append the SSL fingerprint even if EXTERNAL is not used.

SaberUK:
This allows services to implement alternate mechanisms that use the SSL fingerprint.


attilamolnar:
Is this confirmed to not break any anope and atheme versions which are currently working with inspircd?

SaberUK:
@attilamolnar I can't seem to find anything that it would break. For plain and other non-FP mechanisms the fingerprint will just be ignored.

grawity:
+1 in general (UnrealIRCd 4 does the same and nothing broke), but I don't exactly see how this is going to help with SCRAM-*-PLUS?

PR #1271, opened 2016-12-22
Move the OnCheckExemption hook out of the core.

SaberUK:
~~This needs further testing to check it doesn't have any hidden bugs. If anyone is interested in helping test then poke me in \#inspircd.dev.~~

### DIFF
--- a/include/modules/exemption.h
+++ b/include/modules/exemption.h
@@ -1,0 +1,58 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2016-2017 Peter Powell <petpow@saberuk.com>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#pragma once
+
+#include "event.h"
+
+namespace CheckExemption
+{
+	class EventListener;
+	class EventProvider;
+}
+
+class CheckExemption::EventListener
+	: public Events::ModuleEventListener
+{
+ protected:
+	EventListener(Module* mod)
+		: ModuleEventListener(mod, "event/exemption")
+	{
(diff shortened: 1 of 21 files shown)


src/coremods/core_channel/core_channel.cpp
```diff
@@ -110,6 +134,22 @@ class CoreModChannel : public Module
 		invapi.RemoveAll(chan);
 	}
 
+	ModResult OnCheckExemption(User* user, Channel* chan, const std::string& restriction)
+	{
+		if (exemptions.count(restriction))
```
genius3000: Shouldn't this be "!exemptions.count(restriction)"?

SaberUK: Yeah, thanks.

src/coremods/core_channel/cmd_topic.cpp
```diff
@@ -73,10 +74,15 @@ CmdResult CommandTopic::HandleLocal(const std::vector<std::string>& parameters,
...
 			return CMD_FAILURE;
 		}
-		if (c->IsModeSet(topiclockmode) && !ServerInstance->OnCheckExemption(user, c, "topiclock").check(c->GetPrefixValue(user) >= HALFOP_VALUE))
+		if (c->IsModeSet(topiclockmode))
 		{
-			user->WriteNumeric(ERR_CHANOPRIVSNEEDED, c->name, "You do not have access to change the topic on this channel");
-			return CMD_FAILURE;
+			ModResult MOD_RESULT;
+			FIRST_MOD_RESULT_CUSTOM(*this, CheckExemption::EventListener, OnCheckExemption, MOD_RESULT, (user, c, "topiclock"));
+			if (MOD_RESULT.check(c->GetPrefixValue(user) >= HALFOP_VALUE))
```
genius3000: This if statement needs to be inverted as well.

genius3000:
@SaberUK Other than those two inverted if statements, I ran through each mode and exemption; everything seems to work as expected. Can't guarantee I didn't miss anything, but each exemption worked at the level (and above) that I told it to, so looks good.

include/modules/exemption.h
```diff
@@ -0,0 +1,54 @@
...
+	EventListener(Module* mod)
+		: ModuleEventListener(mod, "event/exemption")
+	{
+	}
+
+	/** Called when checking if a user has exemption from something.
+	 * @param user The user to check exemption for.
+	 * @param chan The channel to check exemption on.
+	 * @param restriction The restriction to check for.
+	 * @return ...
```
attilamolnar: Documentation missing from here

attilamolnar:
I'd prefer if you used composition instead of inheritance for the event providers, that way the affected modules can provide multiple events if needed in the future and remain uniform

SaberUK:
@attilamolnar Done.

SaberUK:
@genius3000 Thanks!

PR #1273, opened 2016-12-26
Fix the clock_gettime() test on macOS Sierra

SaberUK:
macOS now supports clock_gettime() but it is part of libSystem rather than librt which does not exist on macOS.


attilamolnar:
Thanks!

PR #1274, opened 2016-12-26
Add interactive-mode support for setting the compiler binary.

SaberUK:
This is a lot more user friendly than expecting the user to set environment variables.


attilamolnar:
Thanks!

PR #1277, opened 2016-12-27
Add OnCheckMembershipVisible event and use for User::ChannelList and …

Adam-:
…in /who for get_first_visible_channel


Adam-:
It was discussed & decided to hack a fix into 2.0 and then revisit this later in master.

PR #1280, opened 2017-01-08
m_dnsbl: check returned results are in 127.0.0.0/8

Adam-:


src/modules/m_dnsbl.cpp
```diff
@@ -70,8 +70,8 @@ class DNSBLResolver : public Resolver
 			int i = countExt.get(them);
 			if (i)
 				countExt.set(them, i - 1);
-			// Now we calculate the bitmask: 256*(256*(256*a+b)+c)+d
-			if(result.length())
+			// All replies should be in 127.0.0.0/8
+			if (result.compare(0, 4, "127.") == 0)
```
SISheogorath: Should we inform the user that it's broken? Sounds useful to me ^^

SaberUK: A snotice seems appropriate.

attilamolnar:
Thanks!

PR #1284, opened 2017-01-23
m_cloaking: don't apply cloak to already cloaked users during registration

Adam-:


src/modules/m_cloaking.cpp
```diff
@@ -96,6 +96,10 @@ class CloakUser : public ModeHandler
 
 		if (adding)
 		{
+			/* assume this is more correct */
```
attilamolnar: Please fix comment style before merging

PR #1287, opened 2017-01-28
Fix some minor problems with the filter docs.

SaberUK:
- Fix tabs being used to align a config attribute instead of spaces.

- Remove warning which only applied to the old configuration format. If a user is still using this then they will have explicitly enabled it and already received a warning.


attilamolnar:
Thanks!

PR #1289, opened 2017-02-04
Add logging for OpenSSL connections.

clinew:
Specifically, log both handshake failures and certificiate
verification.

### DIFF
--- a/src/modules/extra/m_ssl_openssl.cpp
+++ b/src/modules/extra/m_ssl_openssl.cpp
@@ -517,37 +517,88 @@ class OpenSSLIOHook : public SSLIOHook
 	issl_status status;
 	bool data_to_write;
 
+	void PrintCertificate(ssl_cert* cert, int i)
+	{
+		if (!cert->exists)
+			return;
+		ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT,
+			"%2d s:%s", i, cert->GetDN().c_str());
+		ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT,
+			"   i:%s", cert->GetIssuer().c_str());
+		ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT,
+			"   fp:%s", cert->GetFingerprint().c_str());
+	}
+
+	void PrintChain()
+	{
+		PrintCertificate(certificate, 0);
+		for (unsigned i = 0; i < chain.size(); i++)
+		{
+			PrintCertificate(chain.at(i), i + 1);
+		}
+	}
+
 	// Returns 1 if handshake succeeded, 0 if it is still in progress, -1 if it failed
 	int Handshake(StreamSocket* user)
 	{
+		irc::sockets::sockaddrs peersockaddr;
+		socklen_t socklen = sizeof(peersockaddr);
+		std::string addrstr;
+		if (getpeername(user->GetFd(), &peersockaddr.sa, &socklen) == -1)
+			addrstr = "unknown";
+		else
+			addrstr = peersockaddr.addr();
+		irc::sockets::sockaddrs localsockaddr;
+		int port;
+		socklen = sizeof(localsockaddr);
+		if (getsockname(user->GetFd(), &localsockaddr.sa, &socklen) == -1)
+			port = -1;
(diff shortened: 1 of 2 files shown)


src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -463,45 +463,63 @@ class OpenSSLIOHook : public SSLIOHook
 	// Returns 1 if handshake succeeded, 0 if it is still in progress, -1 if it failed
 	int Handshake(StreamSocket* user)
 	{
+		int errerr;
```
SaberUK: Declare these where they are actually used rather than at the top.

src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -463,45 +463,63 @@ class OpenSSLIOHook : public SSLIOHook
...
+		LocalUser* luser = ((UserIOHandler*)user)->user;
 		ERR_clear_error();
 		int ret = SSL_do_handshake(sess);
 		if (ret < 0)
 		{
-			int err = SSL_get_error(sess, ret);
+			// Protocol/connection error.
+			int sslerr = SSL_get_error(sess, ret);
```
SaberUK: Is there any real need to rename this?

clinew: No, just clarity. Having 'err' and 'errerr' seems weird.

src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -463,45 +463,62 @@ class OpenSSLIOHook : public SSLIOHook
 	// Returns 1 if handshake succeeded, 0 if it is still in progress, -1 if it failed
 	int Handshake(StreamSocket* user)
 	{
+		LocalUser* luser = ((UserIOHandler*)user)->user;
```
SaberUK: What about SSL connections which aren't user connections (e.g. m_httpd) that won't use UserIOHandler?

clinew: Test

clinew: I would have assumed that a "user" would be anything that uses the connection, regardless of whether or not it is a person.  Is 'LocalUser' then not appropriate either?  What else should I use?  I need something that I can get the IP, port, Subject and Issuer Distinguished Name, and fingerprint from.  Alternatively, is there a guide that explains the abstractions within InspIRCd?  Also, could you recommend a FOSS CLI tool for perusing C++ code?  cscope only works with C :(

SaberUK: `LocalUser` represents an IRC client which is directly connected to the local server.

`SSLIOHook::IsSSL` might do what you want?

clinew:
This version of the patch takes a slightly different approach.  I couldn't find any way to get from a StreamSocket class and its associated IOHooks to the User class, which has GetIPString() and GetServerPort(), but it does have a file descriptor from which I can call getsockname() and getpeername(); that information can then be used to log the peer IP and local port number.  This works on Linux, but I have no idea about other platforms.  Thoughts?

src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -999,7 +1040,9 @@ class ModuleSSLOpenSSL : public Module
 	ModuleSSLOpenSSL()
 	{
 		// Initialize OpenSSL
-		OPENSSL_init_ssl(0, NULL);
+		if (!(OPENSSL_init_ssl(OPENSSL_INIT_LOAD_SSL_STRINGS, NULL))) {
```
SaberUK: I think that `OPENSSL_INIT_LOAD_SSL_STRINGS` is the default if you pass no options (i.e. 0). Is this not correct?

clinew: Looks like it currently is according to the online man page.

clinew:
Remove superfluous statement.  Minor cleanup.

clinew:
I'd still like to see some kind of logging here, especially because (at least when I wrote this) with SSL client cert enabled zero log messages are made if the SSL handshake fails.

PR #1292, opened 2017-02-05
Allow <log> tags to specify how often logs should be flushed.

### LINKED ISSUE #1290: Logger not flushing messages until block is filled
This results in users not being able to see a message until a large number of messages have been sent.  Culprit appears to be:
```
        if (++writeops % 20 == 0)
        {
                fflush(log);
        }
```
in 'src/logger.cpp', which flushes the log after 20 lines have been written.  Without this code, though, the log only appears to be flushed when the buffer is filled, resulting in a large delay in log messages moving from memory to disk, and partial log messages.

SaberUK:
Also use the behaviour behind this to fix #1290.

### DIFF
--- a/include/logger.h
+++ b/include/logger.h
@@ -41,14 +41,18 @@ class CoreExport FileWriter
 	 */
 	FILE* log;
 
+	/** The number of write operations after which we should flush.
+	 */
+	unsigned int flush;
+
 	/** Number of write operations that have occured
 	 */
-	int writeops;
+	unsigned int writeops;
 
  public:
 	/** The constructor takes an already opened logfile.
 	 */
-	FileWriter(FILE* logfile);
+	FileWriter(FILE* logfile, unsigned int flushcount);
 
 	/** Write one or more preformatted log lines.
 	 * If the data cannot be written immediately,
--- a/src/logger.cpp
+++ b/src/logger.cpp
@@ -120,7 +120,7 @@ void LogManager::OpenFileLogs()
 			struct tm *mytime = gmtime(&time);
 			strftime(realtarget, sizeof(realtarget), target.c_str(), mytime);
 			FILE* f = fopen(realtarget, "a");
-			fw = new FileWriter(f);
+			fw = new FileWriter(f, static_cast<unsigned int>(tag->getInt("flush", 20, 1, INT_MAX)));
 			logmap.insert(std::make_pair(target, fw));
 		}
 		else
@@ -309,8 +309,10 @@ void LogManager::Log(const std::string &type, LogLevel loglevel, const std::stri
 }
 
 
-FileWriter::FileWriter(FILE* logfile)
-: log(logfile), writeops(0)
+FileWriter::FileWriter(FILE* logfile, unsigned int flushcount)
+	: log(logfile)
+	, flush(flushcount)
+	, writeops(0)
 {
 }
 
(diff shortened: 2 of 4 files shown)


SaberUK:
@clinew Can you verify that this fixes the problem?

clinew:
Is the default value supposed to be 20?  If I don't manually set the value, this appears to be the default; manually setting it to '1' appears to work.  The code
`                        fw = new FileWriter(f, tag->getInt("flush", 20, 1));`
uses getInt, which has definition
`long ConfigTag::getInt(const std::string &key, long def, long min, long max)`
Is this supposed to be the default behaviour?

SaberUK:
Yes, the default value for file logging is 20. Weren't you using STDOUT logging with `--nofork --debug` though which now defaults to 1?

clinew:
Nope, I was using file logging with `--nofork` but without `--debug`.  FWIW, I think this is an anti-feature without periodic flushing.

SaberUK:
@clinew This will enable automatic flushing after writing every line for STDOUT logging. That would be a good idea to add at a later date for file logging though.

include/logger.h
```diff
@@ -41,14 +41,18 @@ class CoreExport FileWriter
 	 */
 	FILE* log;
 
+	/** The number of write operations after which we should flush.
+	 */
+	int flush;
```
attilamolnar: If a negative value is never okay here then please change to `unsigned int`.

include/logger.h
```diff
@@ -41,14 +41,18 @@ class CoreExport FileWriter
...
+
 	/** Number of write operations that have occured
 	 */
 	int writeops;
 
  public:
 	/** The constructor takes an already opened logfile.
 	 */
-	FileWriter(FILE* logfile);
+	FileWriter(FILE* logfile, int flushcount);
```
attilamolnar: Same

src/logger.cpp
```diff
@@ -309,8 +309,10 @@ void LogManager::Log(const std::string &type, LogLevel loglevel, const std::stri
 }
 
 
-FileWriter::FileWriter(FILE* logfile)
-: log(logfile), writeops(0)
+FileWriter::FileWriter(FILE* logfile, int flushcount)
```
attilamolnar: Same

include/logger.h
del6597: It may seem nice to use an `unsigned int` to indicate that something can never be negative, but it has consequences when doing arithmetic with a regular `int`. Such is the case further down with `if (++writeops % flush == 0)`.  I'm not sure if there is any impact in this particular scenario, but such an example can be seen [here](http://stackoverflow.com/questions/24745998/c-modulo-of-an-unsigned-integer).

SaberUK:
@attilamolnar Done.

We should add a getter for the config system that handles unsigned integers sometime btw.

attilamolnar:
@SaberUK We use `unsigned int`, not plain `unsigned`, please change newly added code to fit that style.

include/logger.h
attilamolnar: @del6597 Thank you for the input, it indeed can create problems but IMO the solution in this case is to fix the remaining code to use unsigned integers as well, not to increase the amount of code using wrong types.

SaberUK:
@attilamolnar Done.

attilamolnar:
Thanks!

PR #1294, opened 2017-02-12
[2.0] Warn when capability-providing modules are loaded without m_cap.

SaberUK:
Also add reminders to the example configs that m_cap is required.

### DIFF
--- a/src/modules/extra/m_ssl_gnutls.cpp
+++ b/src/modules/extra/m_ssl_gnutls.cpp
@@ -321,6 +321,10 @@ class ModuleSSLGnuTLS : public Module
 
 		ServerInstance->Modules->AddService(iohook);
 		ServerInstance->Modules->AddService(starttls);
+
+		if (starttls.enabled && !ServerInstance->Modules->Find("m_cap.so"))
+			ServerInstance->Logs->Log("m_ssl_gnutls", DEFAULT, "WARNING: m_cap.so is not loaded. The tls capability will not be available to clients without this!");
+
 	}
 
 	void OnRehash(User* user)
--- a/src/modules/m_ircv3.cpp
+++ b/src/modules/m_ircv3.cpp
@@ -83,6 +83,10 @@ class ModuleIRCv3 : public Module
 		OnRehash(NULL);
 		Implementation eventlist[] = { I_OnUserJoin, I_OnPostJoin, I_OnSetAway, I_OnEvent, I_OnRehash };
 		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
+
+		if (!ServerInstance->Modules->Find("m_cap.so"))
+			ServerInstance->Logs->Log("m_ircv3", DEFAULT, "WARNING: m_cap.so is not loaded. The account-notify, away-notify, and extended-join capabilities will not be available to clients without this!");
+
 	}
 
 	void OnRehash(User* user)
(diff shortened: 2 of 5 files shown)


SaberUK:
In 3.0 we should probably upgrade these warnings to `ModuleException`s.

src/modules/m_ircv3.cpp
```diff
@@ -83,6 +83,10 @@ class ModuleIRCv3 : public Module
 		OnRehash(NULL);
 		Implementation eventlist[] = { I_OnUserJoin, I_OnPostJoin, I_OnSetAway, I_OnEvent, I_OnRehash };
 		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
+
+		if (!ServerInstance->Modules->Find("m_cap.so"))
+			ServerInstance->Logs->Log("m_ircv3", DEFAULT, "WARNING: m_cap.so is not loaded. The account-notify, away-notify, and extended-join capabilities will not be available to clients without this!");
+
```
attilamolnar: Extra newline here, please remove

src/modules/extra/m_ssl_gnutls.cpp
```diff
@@ -321,6 +321,10 @@ class ModuleSSLGnuTLS : public Module
 
 		ServerInstance->Modules->AddService(iohook);
 		ServerInstance->Modules->AddService(starttls);
+
+		if (starttls.enabled && !ServerInstance->Modules->Find("m_cap.so"))
+			ServerInstance->Logs->Log("m_ssl_gnutls", DEFAULT, "WARNING: m_cap.so is not loaded. The tls capability will not be available to clients without this!");
```
attilamolnar: I think adding this would create more confusion than help anyone. The wording "The tls capability will not be available to clients without this" can be misunderstood as "no TLS available until you load m_cap". STARTTLS is enabled by default meaning anyone who loads m_ssl_gnutls but doesn't load m_cap would get this warning. The `tls` cap is mostly ignored anyway so users lose next to nothing or nothing at all, so please remove the check from here.

src/modules/m_uhnames.cpp
```diff
@@ -37,6 +37,9 @@ class ModuleUHNames : public Module
 	{
 		Implementation eventlist[] = { I_OnEvent, I_OnPreCommand, I_OnNamesListItem, I_On005Numeric };
 		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
+
+		if (!ServerInstance->Modules->Find("m_cap.so"))
+			ServerInstance->Logs->Log("m_uhnames", DEFAULT, "WARNING: m_cap.so is not loaded. The userhost-in-names capability will not be available to clients without this!");
```
attilamolnar: It is available without m_cap via `PROTOCTL UHNAMES`.

src/modules/m_namesx.cpp
```diff
@@ -37,6 +37,9 @@ class ModuleNamesX : public Module
 	{
 		Implementation eventlist[] = { I_OnPreCommand, I_OnNamesListItem, I_On005Numeric, I_OnEvent, I_OnSendWhoLine };
 		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
+
+		if (!ServerInstance->Modules->Find("m_cap.so"))
+			ServerInstance->Logs->Log("m_namesx", DEFAULT, "WARNING: m_cap.so is not loaded. The multi-prefix capability will not be available to clients without this!");
```
attilamolnar: It is available without m_cap via `PROTOCTL NAMESX`.

src/modules/m_uhnames.cpp
SaberUK: Yeah, but most clients use CAP rather than PROTOCTL nowadays.

PR #1297, opened 2017-02-26
WIP: Add support for the SHA-3 hash algorithm.

SaberUK:
This is based on the public domain reference implementation but has been adapted lightly (editing crypto code is scary so I split it out into lots of commits so it can be audited easily) for InspIRCd's hashing system. See the credits in the header comment for details.

This requires more extensive testing before it can be merged. Especially by people on big endian or non-x86 systems.

### DIFF
--- a/src/modules/m_sha3.cpp
+++ b/src/modules/m_sha3.cpp
@@ -33,10 +33,245 @@ class SHA3Provider : public HashProvider
  private:
 	int capacity;
 
+	/*
+	================================================================
+	Technicalities
+	================================================================
+	*/
+
+	typedef unsigned char UINT8;
+	typedef unsigned long long int UINT64;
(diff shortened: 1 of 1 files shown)


attilamolnar:
> (editing crypto code is scary so I split it out into lots of commits so it can be audited easily)
> This requires more extensive testing before it can be merged. 

True. Plain SHA-3 shouldn't be used for hashing passwords, i.e. sha3(password) is just as bad practice as sha256(password) is. If we want SHA-3 as a building block to offer PBKDF2-HMAC-SHA-3 then I recommend exporting it from the ssl libs. That way we have no extra crypto code, meaning no potential to mess it up, people who have a newer ssl lib will have the choice to use SHA-3 and people having an older ssl lib lacking SHA-3 support will still have bcrypt which we ship the code for already.

PR #1302, opened 2017-03-13
core_dns: add support for txt records

Adam-:


Adam-:
This isn't useful by itself. I envision the dnsbl module using it to get reasons from the blacklist.

rburchell:
Right, I can see that. But what I'd like to see is a mention of that ("this will be useful for dnsbl to get blacklist info xxx") in the commit message, so that it's clear why it was added. When you go back in time, it's nice to have context of why something changed, not just what changed, if that makes sense. A lot of the older changes we have haven't been good at that, which means that as time goes on and people's memories fade (or people leave the project), the _why_ of something is often the hardest part to figure out.

attilamolnar:
Thanks!

rolandhegedus:
@attilamolnar pls check gtalk and remove this comment :D

PR #1310, opened 2017-03-22
Improve and centralize socket engine event counters.

rburchell:
The write counters were close to useless because they were only
incremented on a write "event" which is only triggered when writing
would block.

Read handling was a little more useful in that all reads must happen
through the socket engine, so these were happening at the correct time,
but we can clean this up by doing it in the SE itself rather than each
platform port.

This means that both read and write events are now easily and usefully
defined as "a syscall of either read or write was attempted".

We also count empty read and write events as being an event, because
they still were an attempt to poll a socket in some way. This may help
to identify "bad" code which is repeatedly trying to read a socket for
some reason.

Lastly, we check for failed read/write calls, and log them as an error
event. A lot of the time, this is how sockets are determined as being
disconnected (ie. at read/write time).

While we're at it, split Update() in two to make the calls more
self-describing. This has no real impact since only one call is made at
a time anyway.

### DIFF
--- a/include/socketengine.h
+++ b/include/socketengine.h
@@ -244,11 +244,19 @@ class CoreExport SocketEngine
 		 */
 		Statistics() : lastempty(0), TotalEvents(0), ReadEvents(0), WriteEvents(0), ErrorEvents(0) { }
 
-		/** Increase the counters for bytes sent/received in this second.
-		 * @param len_in Bytes received, 0 if updating number of bytes written.
-		 * @param len_out Bytes sent, 0 if updating number of bytes read.
+		/** Update counters for network data received.
+		 * This should be called after every read-type syscall.
+		 * @param len_in Number of bytes received, or -1 for error, as typically
+		 * returned by a read-style syscall.
 		 */
-		void Update(size_t len_in, size_t len_out);
+		void UpdateReadCounters(int len_in);
+
+		/** Update counters for network data sent.
+		 * This should be called after every write-type syscall.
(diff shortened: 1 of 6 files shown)


attilamolnar:
Since the new methods now accept the return value of a read- or write-like syscall, I think the doc of `len_in`, `len_out` and/or the docs of the new methods themselves should be updated to cover this. Otherwise LGTM.

rburchell:
Seems sensible. Do you have a suggested wording?

> The number of bytes {sent,received}, or -1 for error, as typically returned by a {write,read} style syscall.

?

attilamolnar:
Yes, sounds good

rburchell:
Updated, sorry for the delay.

include/socketengine.h
```diff
@@ -244,11 +244,19 @@ class CoreExport SocketEngine
 		 */
 		Statistics() : lastempty(0), TotalEvents(0), ReadEvents(0), WriteEvents(0), ErrorEvents(0) { }
 
-		/** Increase the counters for bytes sent/received in this second.
-		 * @param len_in Bytes received, 0 if updating number of bytes written.
-		 * @param len_out Bytes sent, 0 if updating number of bytes read.
+		/** Update counters for network data recieved.
```
attilamolnar: Typo in receive

rburchell: Yargh, not a typo. I've been making that mistake most of my life. Fixed ;-)

attilamolnar:
Thanks!

PR #1315, opened 2017-04-07
[2.0] Update securelist exception list.

SaberUK:
- Add an exception for IRC-Source.
- Remove an exception for SearchIRC.
- Remove an old exception for netsplit.


docs/conf/modules.conf.example
```diff
@@ -1602,10 +1602,9 @@
 # Securelist can be harmful to some IRC search engines such as        #
 # netsplit.de and searchirc.com. To prevent securelist blocking these #
```
rburchell: Maybe remove the names from the comments since they are in the exceptions too.

SaberUK: Done, thanks.

PR #1318, opened 2017-04-14
[2.0] Fix compiler version parsing on GCC 7.

### LINKED ISSUE #1309: Fail to build with gcc-7
Hi,

inspircd 2.0.23 fail to build with gcc 7:

```
$ which g++
/usr/bin/g++
[...]
# Really launch configure
./configure --disable-interactive \
        --with-cc=g++ \
        --uid=39 \
        --enable-gnutls \
        --prefix=/usr/lib/inspircd \
        --config-dir=/etc/inspircd \
        --module-dir=/usr/lib/inspircd/modules \
        --binary-dir=/usr/sbin
`g++` was not found! A C++ compiler is required to build InspIRCd!
[...]
```

The `configure` script is looking for minor version of gcc (https://github.com/inspircd/inspircd/blob/insp20/configure#L353-L354) but gcc-7 has only a major version for now:

```
$ g++ -dumpversion
7
$ g++ --version
g++ (Debian 7-20170316-1) 7.0.1 20170316 (experimental) [trunk revision 246203]
[...]
```

There is a new option introduced with gcc-7:

```
$ g++ -dumpfullversion
7.0.1
```

I think the variable `$config{GCCMINOR}` seems not quite useful... maybe it is better to not rely on minor version of gcc.

Sources:
https://gcc.gnu.org/ml/gcc-patches/2017-01/msg00851.html
https://gcc.gnu.org/ml/gcc-patches/2017-01/msg00593.html

SaberUK:
GCC 7 may optionally omit the MINOR version number in dumpversion depending on the compiler flags it was built with.

This also fixes a potential bug which will appear when GCCVER/GCCMINOR is greater than 9.

Fixes #1309.

### DIFF
--- a/configure
+++ b/configure
@@ -269,10 +269,9 @@ if (defined $opt_cc)
 {
 	$config{CC} = $opt_cc;
 }
-our $exec = $config{CC} . " -dumpversion | cut -c 1";
-chomp($config{GCCVER}		= `$exec`);				# Major GCC Version
-$exec = $config{CC} . " -dumpversion | cut -c 3";
-chomp($config{GCCMINOR}		= `$exec`);
+`$config{CC} -dumpversion` =~ /^(\d+)(?:\.(\d+))?/;
+$config{GCCVER} = defined $1 ? $1 : '';
+$config{GCCMINOR} = defined $2 ? $2 : '0';
 $config{MAXBUF}			= "512";				# Max buffer size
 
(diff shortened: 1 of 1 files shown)


rburchell:
Do you have a reference on the GCC behaviour change? It seems strange that they'd continue with a breaking behaviour like this

SaberUK:
@rburchell It is mentioned in the linked issue.

attilamolnar:
Thanks!

PR #1319, opened 2017-04-14
Switch compiler detection to use a more reliable method.

SaberUK:
Its clear that parsing version output is not reliable enough (see #1309) so switch to using a method which is less likely to break.

### DIFF
--- a/make/test/compiler_info.cpp
+++ b/make/test/compiler_info.cpp
@@ -1,0 +1,41 @@
+/*
+ * InspIRCd -- Internet Relay Chat Daemon
+ *
+ *   Copyright (C) 2017 Peter Powell <petpow@saberuk.com>
+ *
+ * This file is part of InspIRCd.  InspIRCd is free software: you can
+ * redistribute it and/or modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation, version 2.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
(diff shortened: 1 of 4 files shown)


make/test/compiler_info.cpp
```diff
@@ -0,0 +1,41 @@
...
+# if defined __apple_build_version__
+#  define INSPIRCD_COMPILER_NAME "AppleClang"
+# else
+#  define INSPIRCD_COMPILER_NAME "Clang"
+# endif
+# define INSPIRCD_COMPILER_VERSION __clang_major__ << '.' << __clang_minor__
+#elif defined __GNUC__
+# define INSPIRCD_COMPILER_NAME "GCC"
+# define INSPIRCD_COMPILER_VERSION __GNUC__ << '.' << __GNUC_MINOR__
+#endif
```
rburchell:
```
#else
# error ...
#endif
```

... paranoia, but probably a good idea regardless?

make/test/compiler_info.cpp
```diff
@@ -0,0 +1,41 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#include <iostream>
+
+#if defined __INTEL_COMPILER // Also defines __clang__ and __GNUC__
+# define INSPIRCD_COMPILER_NAME "Intel"
+# define INSPIRCD_COMPILER_VERSION (__INTEL_COMPILER / 100) << '.' << (__INTEL_COMPILER % 100)
```
rburchell: Is anyone actively testing ICC?

make/test/compiler_info.cpp
SaberUK: Unknown compilers will just get the defaults (NAME => Unknown, VERSION => 0.0).

make/test/compiler_info.cpp
SaberUK: I do every now and again.

make/test/compiler_info.cpp
rburchell: Right, true enough.

attilamolnar:
Thanks!

Issue #1317, opened 2017-04-11
Permchannels should automatically set +P when loading the database if not already set


SaberUK:
Otherwise when the channel empties it will be deleted until the next boot.

I'm not sure if this is a bug we can fix in 2.0 or if it should be deferred to 3.0 for compat reasons.

---

Regardless, i've written a patch for master to fix it: https://github.com/inspircd/inspircd/compare/master...SaberUK:master+autoperm

attilamolnar:
Under what circumstances is it possible that a channel is in the database and not have +P?

SaberUK:
@attilamolnar When people define a permanent channel via defining their own `<permchannels>` tags (we document how to do this in modules.conf.example so I guess that means we support it) rather than just from setting `+P` and loading the generated file.

attilamolnar:
@SaberUK makes sense, I think a doc update + one time log message if the situation is detected is fine for 2.0. For 3.0, push the permchannels mode to the mode change list after all modes. If it's duplicate, it will be dropped harmlessly.

attilamolnar:
For 3.0, calling our mode handler after the other modes should be fine in all cases, even if the permchannels mode is already set.

SaberUK:
@attilamolnar I didn't notice earlier but there is already a warning in the config in 2.0, i'll just add a log message for it.

attilamolnar:
This has been implemented on both insp20 and master, closing.

PR #1321, opened 2017-04-14
[2.0] Warn when a user configures a permanent channel without +P.

SaberUK:
Ref: #1317

### DIFF
--- a/src/modules/m_permchannels.cpp
+++ b/src/modules/m_permchannels.cpp
@@ -298,6 +298,12 @@ public:
 
 				ServerInstance->Logs->Log("m_permchannels", DEBUG, "Added %s with topic %s", channel.c_str(), topic.c_str());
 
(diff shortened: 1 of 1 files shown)


src/modules/m_permchannels.cpp
```diff
@@ -298,6 +298,10 @@ class ModulePermanentChannels : public Module
 
 				ServerInstance->Logs->Log("m_permchannels", DEBUG, "Added %s with topic %s", channel.c_str(), topic.c_str());
 
+				if (modes.find('P') == std::string::npos)
```
rburchell: I'd prefer to see braces around this since it's such a long body

SaberUK: Added.

attilamolnar:
Thanks!

PR #1324, opened 2017-04-27
[2.0] Fix WEBIRC not rejecting invalid IP addresses.

SaberUK:
If an IP address which is invalid is given then the IP address given to the user `webirc_ip` will contain junk. This is not as dangerous as it looks as the core will reject the invalid IP when SetClientIP is called. It should be rejected earlier than that though to prevent incorrect notices being given.

Some examples of things this patch prevents:

```
WEBIRC password * ###################################################################### :
WEBIRC password * ###################################################################### :invalid ipaddr here
```

---

Also penalise clients that use WEBIRC incorrectly (mostly useful for preventing trolls from spamming opers with invalid WEBIRC host messages).

### DIFF
--- a/src/modules/m_cgiirc.cpp
+++ b/src/modules/m_cgiirc.cpp
@@ -85,6 +85,7 @@ class CommandWebirc : public Command
 			irc::sockets::sockaddrs ipaddr;
 			if (!irc::sockets::aptosa(parameters[3], 0, ipaddr))
 			{
+				IS_LOCAL(user)->CommandFloodPenalty += 5000;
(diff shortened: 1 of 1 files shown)


src/modules/m_cgiirc.cpp
```diff
@@ -81,6 +82,14 @@ class CommandWebirc : public Command
 			if(user->registered == REG_ALL)
 				return CMD_FAILURE;
 
+			irc::sockets::sockaddrs ipaddr;
+			if (!irc::sockets::aptosa(parameters[3], 0, ipaddr))
+			{
+				IS_LOCAL(user)->CommandFloodPenalty += 5000;
```
SaberUK: This is not harmful as the command is local-only. We should probably migrate to SplitCommand but I didn't want to make such an invasive change in 2.0.

PR #1325, opened 2017-05-01
Add SHA-512 support to SQL authenticator

Cyanodioxine:
Requires the extra module m_hash_gnutls, otherwise $sha512pass won't populate.


SaberUK:
You could alternately do this with the built in hashing functions of your SQL server. 

e.g. MySQL: ` SHA2("$pass", 512)`C

PR #1330, opened 2017-05-16
[2.0] Fix ModuleManager not being able to find make::configure.

SaberUK:
Apparently not only is the Perl documentation on BEGIN block execution order wrong but in a BEGIN block any updates to the include paths are not applied until the end of the block. Yay!


attilamolnar:
Thanks!

Issue #1332, opened 2017-05-19
make problem OpenBSD


layer3:
Hi,

./configure runs fine getting the error when trying to execute make (make: no target to make.)

OpenBSD: 6.1

SaberUK:
What version of InspIRCd are you using?

What is the output of `make -v`?

layer3:
V2.0.23 

# make -v
make: unknown option -- v

SaberUK:
Can you pastebin the full output of running configure and then make?

layer3:
https://pastebin.com/9gJ5ZQB1

SaberUK:
Can you try updating to v2.0.24? I'm building an OpenBSD VM now to test this.

layer3:
yeah, I can try 2.0.24 no problem! let you know then!

layer3:
same error

Signing certificate...

Certificate generation complete, copying to config directory... Done.

Symlinking src/modules/m_ssl_openssl.cpp from extra/
SSL certificates found, skipping.

Detecting modules ..................................................................................................................................................
Ok, 146 modules.
Locating include directory for package gnutls for module m_ssl_gnutls.cpp... -I/usr/local/include -I/usr/local/include/p11-kit-1 -I/usr/include -DVERSION_GNUTLS="3.5.10" (version 3.5.10)
Locating library directory for package gnutls for module m_ssl_gnutls.cpp... -L/usr/local/lib -lgnutls (version 3.5.10)
Adding runtime library path to m_ssl_gnutls.cpp ... /usr/local/lib
Checking version of package openssl is >= 0.9.7... Yes (version 1.0.0)
Locating include directory for package openssl for module m_ssl_openssl.cpp... -I/usr/include -DVERSION_OPENSSL="1.0.0" (version 1.0.0)
Locating library directory for package openssl for module m_ssl_openssl.cpp... -L/usr/lib -lssl -lcrypto (version 1.0.0)
Adding runtime library path to m_ssl_openssl.cpp ... /usr/lib
Writing inspircd_config.h
Writing GNUmakefile ...
Writing BSDmakefile ...
Writing inspircd ...
Writing cache file for future ./configures ...


To build your server with these settings, please run 'make' now.
Please note: for SSL support you will need to load required
modules in your config. This configure script has added those modules to the
build process. For more info, please refer to:
http://wiki.inspircd.org/Installation_From_Tarball
*** Remember to edit your configuration files!!! ***


# make
make: no target to make.
#

[...]

PR #1333, opened 2017-05-19
[2.0] Tell OpenBSD users that they need to use a different make command.

SaberUK:
OpenBSD decided to remove `BSDmakefile` support because it is non standard (heaven forbid!) so our automagic makefile stuff no longer works. We can't do a lot about this other than warn OpenBSD users to run a different command instead.

Ref: https://github.com/openbsd/src/commit/7c8e039b953a21c37b44a481b0a7b62a845164dc

See also: #1332

### DIFF
--- a/configure
+++ b/configure
@@ -733,6 +733,14 @@ to a newer 3.x or 4.x (or whatever is available currently) version.
 FOO2
 }
 
+if ($^O eq 'openbsd') {
+	print <<__OPENBSD_WARNING__;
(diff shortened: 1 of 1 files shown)


SaberUK:
I'm wondering about whether we should just remove the makefile shim and just support GNU make. Any comments on that?

attilamolnar:
Thanks!

> I'm wondering about whether we should just remove the makefile shim and just support GNU make. Any comments on that?

I agree, it simplies the build related code.

PR #1334, opened 2017-05-19
Remove some dead code from m_md5.

SaberUK:
This is left over from the old hashing system that allowed you to customise these values.


attilamolnar:
Thanks!

PR #1335, opened 2017-05-19
Tidy up the socket engine selection code.

SaberUK:
Just some low hanging fruit left over from before we had `print_error` and from when you could change the socket engine in interactive mode.


attilamolnar:
Thanks!

PR #1338, opened 2017-05-20
Warn the user about clock drifting when running on OpenVZ.

SaberUK:
Apparently even though its 2017 and there are much better options for virtualisation available this is still a problem that people need to be warned about.


attilamolnar:
Thanks!

PR #1341, opened 2017-05-23
Insp20

carlosferry:
Ability to merge two channels. Please see https://github.com/carlosferry/inspircd/blob/master/m_merge.cpp for example.


Adam-:
This should be filed against https://github.com/inspircd/inspircd-extras, not here.

PR #1342, opened 2017-05-23
Allow for multiple fingerprints in an oper block

linuxdaemon:
Allows a user to specify a space separated list of SSL fingerprints for an oper block.

### DIFF
--- a/src/modules/m_sslinfo.cpp
+++ b/src/modules/m_sslinfo.cpp
@@ -172,6 +172,20 @@ class ModuleSSLInfo : public Module
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


SaberUK:
This should be filed against master. insp20 is not receiving big changes.

linuxdaemon:
I can do that, I was more hoping to get his in 2.0 since it doesn't break current configs or anything and it fixes an inconsistency.

attilamolnar:
Thanks for this, please port this to the master branch. As mentioned by @SaberUK 2.0 is bugfix only.

PR #1343, opened 2017-05-30
[2.0] Update the securelist exemption for IRC-Source's crawler.

SaberUK:
Literally a few days after we released they decided to change their crawling domain. Argh.

https://twitter.com/theircsource/status/866890251423371267


attilamolnar:
Thanks!

PR #1347, opened 2017-06-16
[2.0] Fix exempting CTCP ACTIONs in m_blockcaps and m_noctcp.

SaberUK:
Previously we assumed that CTCP ACTIONs matched "\1ACTION ". This is incorrect because "\1ACTION\1" and "\1ACTION" are valid CTCPs.

### DIFF
--- a/src/modules/m_blockcaps.cpp
+++ b/src/modules/m_blockcaps.cpp
@@ -67,7 +67,7 @@ public:
 	{
 		if (target_type == TYPE_CHANNEL)
 		{
-			if ((!IS_LOCAL(user)) || (text.length() < minlen))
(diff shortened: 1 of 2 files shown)


genius3000:
The change to m_blockcaps, won't that just allow any ACTION through now; instead of ignoring 'ACTION' in the capitals count?

SaberUK:
>The change to m_blockcaps, won't that just allow any ACTION through now; instead of ignoring 'ACTION' in the capitals count?

No because its comparing to the entire message. An empty action contains no characters and therefore no capital letters.

genius3000:
Ah, I totally missed that lol, thanks.

Issue #1247, opened 2016-11-10
module m_filter - adding filtered message


Gavin-X:
i think it would be more useful to ircops to see the msg that's currently being block than the reason why it was blocked.

[09:20:19] -aurora.jhb.za.zairc.net- *** ANNOUNCEMENT: FILTER: Cobra had their message filtered, target was #AdultChat: **Spam/Advertising (See: /RULES (#4)) [ID: 2016061602] [OID: GP]**

The reason should be displayed to the user instead of the ircops. 

Maybe have a setting in the conf to choose to show either reason or msg

Gavin-X:
Suggestion - ServerInstance->SNO->WriteGlobalSno('a', "FILTER: "+user->nick+" had their message filtered, target was "+target+": "+text+" ["+f->reason+"]");

SaberUK:
I'm closing this for the reason mentioned in #1348.

Robby-:
@SaberUK This is still open, it seems you didn't click the right button.

SaberUK:
Oops, fixed~

PR #1348, opened 2017-06-19
Blocked Text from user in notice to opers

Gavin-X:
logged in https://github.com/inspircd/inspircd/issues/1247

### DIFF
--- a/src/modules/m_filter.cpp
+++ b/src/modules/m_filter.cpp
@@ -341,7 +341,7 @@ ModResult ModuleFilter::OnUserPreMessage(User* user, void* dest, int target_type
 		}
 		if (f->action == FA_BLOCK)
 		{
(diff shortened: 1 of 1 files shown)


SISheogorath:
I'm not a real fan of this patch because it maybe exposes sensitive or secret data.

It can for example expose passwords which contain newly censored words. If you send them to nickserv and they are blocked, it's exposed to the oper. For sure it's possible by modify the IRCds sources or use very verbose logging to get this but I would try to avoid it make it easier than it already is.

A possible attack vector is opened. If someone make it to get oper access on the network and is capable to set filters he can filter the word `identify` and all passwords are exposed until someone notice this problem.

SaberUK:
I agree with @SISheogorath. The risk of exposing passwords makes this undesirable.

SaberUK:
As you haven't replied in five~ months i'm going to close this. Please reopen if you have any more comments.

Issue #1336, opened 2017-05-19
Cloaking fails to cloak partial host names when host name has "." within it


bubthegreat:
Straightforward issue - cloaking breaks for any hostname that has a "." in it - so if my hostname is username@host.name.something.com, it will only cloak the username@host. and still exposes the .name.something.com.

from m_cloaking:

```
	std::string LastTwoDomainParts(const std::string &host)
	{
		int dots = 0;
		std::string::size_type splitdot = host.length();

		for (std::string::size_type x = host.length() - 1; x; --x)
		{
			if (host[x] == '.')
			{
				splitdot = x;
				dots++;
			}
			if (dots >= 3)
				break;
		}

		if (splitdot == host.length())
			return "";
		else
			return host.substr(splitdot);
	}
```
Seems like modifying this to if (dots >= 2) would be enough to still cloak appropriately, and would just "over cloak" the network, which doesn't seem to be an issue unless there are other modules that depend on the fully cloaked name.

The logic seems straightforward in that it looks for the last "." from the right, and then cloaks everything to the left of that location, when it should only do so for the first two so you cloak anything before domain.extension, but I may be missing something that this would break dependencies for, so bugging it for wiser minds to review.

SISheogorath:
Just as a hint: The word `hostname` describes the host part of an FQDN(Full Qualified Domain Name) or to be more correct: FQHN (Full Qualified Host Name). The delimiter for a Domain name is always a `.`. in other words a Hostname can't contain a `.` by definition.

Anyways are right that the current way does show the 3rd domain part. So the function or the description of it is wrong. I'm pretty sure we can't fix this for `insp20` but for `master` aka InspIRCd 3.0 it should be possible :)

If you want to fix it yourself feel free to file a PR. Otherwise we'll take care of it :+1:

SaberUK:
>I'm pretty sure we can't fix this for insp20 but for master aka InspIRCd 3.0 it should be possible :)

We need to retain link compat with 2.0 so in order to fix in master #1227 or something similar would be required.

SISheogorath:
You're right. We can add a new config parameter to configure how many parts of the domain left as cleartext in mode "half". In full mode, it's ignored.

SaberUK:
@bubthegreat This has been implemented in #1353.

PR #1353, opened 2017-06-27
Make domainparts configureable

SISheogorath:
Basic modification to solve #1336 

Compiled but untested.

Maybe @bubthegreat  can test, as he requested this change.

### DIFF
--- a/src/modules/m_cloaking.cpp
+++ b/src/modules/m_cloaking.cpp
@@ -143,6 +143,7 @@ class ModuleCloaking : public Module
 	std::string prefix;
 	std::string suffix;
 	std::string key;
+	unsigned int domainparts;
 	dynamic_reference<HashProvider> Hash;
 
 	ModuleCloaking() : cu(this), mode(MODE_OPAQUE), ck(this), Hash(this, "hash/md5")
@@ -160,7 +161,7 @@ class ModuleCloaking : public Module
 	 */
 	std::string LastTwoDomainParts(const std::string &host)
 	{
-		int dots = 0;
+		unsigned int dots = 0;
 		std::string::size_type splitdot = host.length();
 
 		for (std::string::size_type x = host.length() - 1; x; --x)
@@ -170,7 +171,7 @@ class ModuleCloaking : public Module
 				splitdot = x;
 				dots++;
 			}
-			if (dots >= 3)
+			if (dots >= domainparts)
 				break;
 		}
 
@@ -314,7 +315,15 @@ class ModuleCloaking : public Module
 			switch (mode)
 			{
 				case MODE_HALF_CLOAK:
-					testcloak = prefix + SegmentCloak("*", 3, 8) + suffix;
+					// Use old cloaking verification to stay compatible with 2.0
+					// But verify domainparts when use 3.0-only features
+					if (domainparts == 3)
+						testcloak = prefix + SegmentCloak("*", 3, 8) + suffix;
+					else
+					{
+						irc::sockets::sockaddrs sa;
+						testcloak = GenCloak(sa, "", testcloak + ConvToStr(domainparts));
+					}
 					break;
 				case MODE_OPAQUE:
 					testcloak = prefix + SegmentCloak("*", 4, 8) + suffix;
@@ -331,7 +340,10 @@ class ModuleCloaking : public Module
 
 		std::string modestr = tag->getString("mode");
 		if (modestr == "half")
+		{
 			mode = MODE_HALF_CLOAK;
+			domainparts = tag->getInt("domainparts", 3, 1, 10);
+		}
 		else if (modestr == "full")
 			mode = MODE_OPAQUE;
 		else
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -481,9 +481,12 @@
 #                                                                     #
 # There are two methods of cloaking:                                  #
 #                                                                     #
-#   half           Cloak only the "unique" portion of a host; show    #
-#                  the last 2 parts of the domain, /16 subnet of IPv4 #
-#                  or /48 subnet of the IPv6 address.                 #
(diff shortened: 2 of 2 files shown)


bubthegreat:
Will probably be a day or two before I can test on the server, but will do.  I was looking at some code for other IRC clients, and they took an approach of checking things from right to left, rather than left to right.

I guess I don't understand the necessity to leave *any* of it uncloaked, rather than masking the entire thing, domain and all - is there some dependency that would break if that were an option?

SISheogorath:
For everything cloaked you have cloaking mode `full`. It wouldn't need this modification

Heufneutje:
> I guess I don't understand the necessity to leave any of it uncloaked, rather than masking the entire thing, 

Mainly for having a recognizable part of the hostname while still hiding the parts that would let one figure out the IP. It makes it easier for channel operators to manage bans and such. If every cloak is a full hash it makes it hard to distinguish.

src/modules/m_cloaking.cpp
```diff
@@ -327,8 +328,11 @@ class ModuleCloaking : public Module
 		suffix = tag->getString("suffix", ".IP");
 
 		std::string modestr = tag->getString("mode");
-		if (modestr == "half")
+		if (modestr == "half") 
+		{
 			mode = MODE_HALF_CLOAK;
+			domainparts = tag->getInt("domainparts", 3);
```
SaberUK: You should also add some validation to this using the min/max parameters.

SISheogorath: Mhm currently thinking what's the better min. `0` or `1` :thinking:

In the case of `0` full cloaking would be the better choice.

And there is no max by RFC. So I'm no sure if provide one is really perfect. In general, a max of 5 or 10 should be fine.

SaberUK: 1-10 sounds reasonable to me.

SaberUK:
Don't forget to document this in `modules.conf.example`.

SaberUK:
LGTM.

SISheogorath:
Still not tested. I would like to test it before we merge it.

src/modules/m_cloaking.cpp
```diff
@@ -327,8 +328,11 @@ class ModuleCloaking : public Module
 		suffix = tag->getString("suffix", ".IP");
 
 		std::string modestr = tag->getString("mode");
-		if (modestr == "half")
+		if (modestr == "half")
```
attilamolnar: Extra whitespace at the end of the line

SISheogorath:
Requested changes are done.

src/modules/m_cloaking.cpp
```diff
@@ -309,15 +310,16 @@ class ModuleCloaking : public Module
...
 	{
 		std::string testcloak = "broken";
+		irc::sockets::sockaddrs sa;
 		if (Hash)
 		{
 			switch (mode)
 			{
 				case MODE_HALF_CLOAK:
-					testcloak = prefix + SegmentCloak("*", 3, 8) + suffix;
+					testcloak = GenCloak(sa, "", testcloak + std::to_string(domainparts));
```
SaberUK: Does this break linking compat with 2.0? I think that it might...

Also, std::to_string is a C++11 addition. Use ConvToStr instead.

SISheogorath: Yes, it breaks. But I'm not sure what's the alternative to sync `domainparts` then. I can only think about some complicated way using `OnSyncNetwork` which I would like to avoid.

SaberUK: You could do something to the effect of this:

````
if domainparts is what it was set to in 2.0:
   set testcloak to the 2.0 format
else
  set testcloak to the new format
end
````

SaberUK:
LGTM.

attilamolnar:
Can you please describe how was this tested?

Specifically, I'm interested in testing linking. There are many possible scenarios, e.g.
- the two sides are both 3.0, `domainparts` = 3 on both sides, cloaking key matches
- the two sides are both 3.0, `domainparts` = 3 on both sides, cloaking key does not match
- the two sides are both 3.0, `domainparts` = 2 on both sides, cloaking key matches
- the two sides are both 3.0, `domainparts` = 2 on both sides, cloaking key does not match
- mixed 2.0/3.0, 3.0 has `domainparts` = 3 and cloaking key matches
- mixed 2.0/3.0, 3.0 has `domainparts` = 3 and cloaking key does not match
- mixed 2.0/3.0, 3.0 has `domainparts` != 3 and cloaking key matches
- mixed 2.0/3.0, 3.0 has `domainparts` != 3 and cloaking key does not match

Which ones were verified?

SISheogorath:
Test results:

* the two sides are both 3.0, domainparts = 3 on both sides, cloaking key matches - > `SUCCESSED`
* the two sides are both 3.0, domainparts = 3 on both sides, cloaking key does not match -> `FAILED`
* the two sides are both 3.0, domainparts = 2 on both sides, cloaking key matches -> `SUCCESSED`
* the two sides are both 3.0, domainparts = 2 on both sides, cloaking key does not match -> `FAILED`
* the two sides are both 3.0, domainparts = 2 on one side, domainparts = 3 on the other side, cloaking key matches -> `FAILED`
* the two sides are both 3.0, domainparts = 2 on one side, domainparts = 3 on the other side, cloaking key does not match --> `FAILED`
* mixed 2.0/3.0, 3.0 has domainparts = 3 and cloaking key matches -> `SUCCESS`
* mixed 2.0/3.0, 3.0 has domainparts = 3 and cloaking key does not match -> `FAILED`
* mixed 2.0/3.0, 3.0 has domainparts != 3 and cloaking key matches -> `FAILED`
* mixed 2.0/3.0, 3.0 has domainparts != 3 and cloaking key does not match -> `FAILED`

PR #1358, opened 2017-07-25
2.0: [Fix?] Make RLines match against IP as well as host

genius3000:
RLines currently only check for a match including the User's host and not the IP. This may have been intentional but I believe to be an oversight.
One may wish to set an RLine using a portion of the IP and it stands the chance of not matching if their IP resolves to any host.


SaberUK:
This is a behaviour change which is not suitable for 2.0, can you retarget it to master? Thanks.~

genius3000:
Replaced this PR with #1361

PR #1359, opened 2017-07-27
2.0: Inform the client when a SASL message cannot be sent

genius3000:
Back when the SASL `target` variable was introduced (commit 8cb1935360087b4e38802b837981e5f41e9b87d7), the example config wasn't updated and it's not a very well known config variable. This variable allows the ENCAP message to return false if the server isn't available (ie, Services are offline). Though currently, even when used it just sends a SASLFallback event that (afaik) is not caught anywhere.

The issue here is that if Services are down and a client is connecting and sees SASL available (through CAP), it requests it, gets acknowledged and tries to use it. The message either sends to all servers (default target of '*') or fails to send to the target server. The client sits there waiting for a response, until registration timeout hits. We can't do much with sending to all servers and receiving no reply (afaik), but we can let the client know when a targeted message fails. This is a fairly small change (imo) and won't affect anything unless a target is specified and Services are down.
I know some clients do a SASL timeout, but that can still be longer than the server's registration timeout.
For reference, I know that Quassel and some (if not all) versions of ZNC have this timeout issue with SASL and Services being down. This just happened to me yesterday on ChatSpike.

3.0 (master) handles this differently by not sending the SASL CAP when the target server is offline. That's too much of a change for 2.0 though.

This has been tested to work on a 2-server insp20 + Anope network.

### DIFF
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -1591,6 +1591,13 @@
 # Layer via AUTHENTICATE. Note: You also need to have m_cap.so loaded
 # for SASL to work.
 #<module name="m_sasl.so">
+# Define the following to your services server name to improve security
+# by ensuring the SASL messages are only sent to the services server
+# and not to all connected servers. This prevents a rogue server from
+# capturing SASL messages. Having this defined can also improve client
+# connections when your services are down, as the client will be told
+# that SASL failed rather than just timing out on registration.
+#<sasl target="services.mynetwork.com">
 
 #-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#
 # Secure list module: Prevent /LIST in the first minute of connection,
--- a/src/modules/m_sasl.cpp
+++ b/src/modules/m_sasl.cpp
@@ -35,6 +35,10 @@ static void SendSASL(const parameterlist& params)
 {
 	if (!ServerInstance->PI->SendEncapsulatedData(params))
 	{
+		User* u = ServerInstance->FindUUID(params[2]);
+		if (u)
+			u->WriteNumeric(904, "%s :SASL authentication failed", u->nick.c_str());
+
 		SASLFallback(NULL, params);
 	}
 }


SaberUK:
Is that the right numeric to send? From the IRCv3.1 SASL spec:

>906 aka ERR_SASLABORTED is sent when the SASL authentication is aborted because the client sent an AUTHENTICATE command with * as the parameter.

I think 904 might be more appropriate?

>904 aka ERR_SASLFAIL is sent when the SASL authentication fails because of invalid credentials or other errors not explicitly mentioned by other numerics.

Robby-:
904 seems the most appropriate for this indeed.

In my opinion it would have been nicer if the spec had a dedicated numeric for SASL failures due to incorrect credentials, and another numeric for all other errors not explicitly mentioned by other numerics.

genius3000:
Yeah it does appear that 904 would be the one to use.

src/modules/m_sasl.cpp
```diff
@@ -35,6 +35,10 @@ static void SendSASL(const parameterlist& params)
 {
 	if (!ServerInstance->PI->SendEncapsulatedData(params))
 	{
+		User* u = ServerInstance->FindUUID(params[2]);
+		if (u)
+			u->WriteNumeric(904, "%s :SASL authentication failed", u->nick.c_str());
+
```
SaberUK: Have you checked whether `authExt` needs to be unset here?

docs/conf/modules.conf.example
```diff
@@ -1591,6 +1591,8 @@
 # Layer via AUTHENTICATE. Note: You also need to have m_cap.so loaded
 # for SASL to work.
 #<module name="m_sasl.so">
+# Define the following to your Services server name to be more secure.
```
SaberUK: Services should be lower case here as its a common noun not a proper noun.

Also, you should say /why/ using this setting is more secure rather than just saying that it is more secure.

src/modules/m_sasl.cpp
genius3000: As far as I can tell it's unset in the OnUserConnect catch by checking if the user has an active SaslAuthenticator, then aborting and unsetting it. I have not attempted a deeper "debug" level check of this, but clients like (iirc)Weechat do a SASL timeout and continue connecting if they don't get a registration timeout, which would have the same effect on the set SaslAuthenticator.

docs/conf/modules.conf.example
genius3000: Fixed/updated

src/modules/m_sasl.cpp
genius3000: I placed a simple debug snomask message inside the if block at line 346 (if user has a SaslAuthenticator) and it is catching the continued connect after attempting a SASL auth, therefore aborting and unsetting it.

SaberUK:
Merged, apologies for the delay and thanks! 👍

PR #1361, opened 2017-08-05
Make RLines match against IP as well as host

genius3000:
RLines currently only check for a match including the User's host and not the IP. This may have been intentional but I believe to be an oversight.
One may wish to set an RLine using a portion of the IP and it stands the chance of not matching if their IP resolves to any host.

This PR replaces #1358

### DIFF
--- a/src/modules/m_rline.cpp
+++ b/src/modules/m_rline.cpp
@@ -62,8 +62,9 @@ class RLine : public XLine
 		if (lu && lu->exempt)
 			return false;
 
-		std::string compare = u->nick + "!" + u->ident + "@" + u->host + " " + u->fullname;
-		return regex->Matches(compare);
+		const std::string host = u->nick + "!" + u->ident + "@" + u->host + " " + u->fullname;
+		const std::string ip = u->nick + "!" + u->ident + "@" + u->GetIPString() + " " + u->fullname;
+		return (regex->Matches(host) || regex->Matches(ip));
 	}
 
 	bool Matches(const std::string &compare)


attilamolnar:
> This may have been intentional but I believe to be an oversight.
I believe this is intentional because the helpop states:
> `Sets or removes an r-line (regex line) on a `n!u@h\\sgecos mask.`

> One may wish to set an RLine using a portion of the IP and it stands the chance of not matching if their IP resolves to any host.
If their IP does not resolve to any host then the `host` field of `User` will contain the IP string and an rline for the IP will work with the current code. If the goal is to ban an IP or IP range cidr zlines can be used. Do you need to match an IP and the realname?

genius3000:
> I believe this is intentional because the helpop states:
`Sets or removes an r-line (regex line) on a n!u@h\\sgecos mask.`

On that note, K/G/E-line all say similar:
`Sets or removes a g-line (host based ban) on host mask.`
Yet they match to both host and IP string. Knowing this, one would assume an RLine would follow the same rule of matching.

> If their IP does not resolve to any host then the host field of User will contain the IP string and an rline for the IP will work with the current code. Is this not the case?

Yes and that is why my issue is only with when a host does resolve and doesn't match an IP RLine mask.

> If the goal is to ban an IP or IP range cidr zlines can be used. Do you need to match an IP and the realname?

I agree, ZLines are likely to be the better option for most things banning IP. I see the need for RLine to match against IP as well as host because it is the only regex server ban. As often as one may need to use a regex ban with a hostmask, you could also be using an IP mask. I personally don't see the reason to not match against both, like the rest do.

PR #1370, opened 2017-09-16
Add OpenSSL CRLs.

clinew:
The 'crlfile' argument can point to a file that contains valid CRLs.
The 'crlpath' argument can point to a directory which contains CRLs,
albeit in OpenSSL's special hashed/symlink format.  The 'crlmode' option
'chain' checks all certificates in the chain while the option 'leaf'
checks only the leaf certificate in a chain.

### DIFF
--- a/src/modules/extra/m_ssl_openssl.cpp
+++ b/src/modules/extra/m_ssl_openssl.cpp
@@ -200,6 +200,45 @@ namespace OpenSSL
 			return SSL_CTX_load_verify_locations(ctx, filename.c_str(), 0);
 		}
 
+		void SetCRL(const std::string& crlfile, const std::string& crlpath, const std::string& crlmode)
+		{
+			if (crlfile.empty() && crlpath.empty())
+				return;
+
+			/* Set CRL mode */
+			unsigned long crlflags = X509_V_FLAG_CRL_CHECK;
+			if (crlmode == "chain")
+			{
+				crlflags |= X509_V_FLAG_CRL_CHECK_ALL;
+			}
+			else if (crlmode != "leaf")
+			{
+				throw ModuleException("Unknown mode '" + crlmode + "'; expected either 'chain' (default) or 'leaf'");
+			}
+
+			/* Load CRL files */
+			X509_STORE* store = SSL_CTX_get_cert_store(ctx);
+			if (!store)
+			{
+				throw ModuleException("Unable to get X509_STORE from SSL context; this should never happen");
+			}
+			ERR_clear_error();
+			if (!X509_STORE_load_locations(store,
+				crlfile.empty() ? NULL : crlfile.c_str(),
+				crlpath.empty() ? NULL : crlpath.c_str()))
+			{
+				int err = ERR_get_error();
+				throw ModuleException("Unable to load CRL file '" + crlfile + "' or CRL path '" + crlpath + "': '" + (err ? ERR_error_string(err, NULL) : "unknown") + "'");
+			}
+
+			/* Set CRL mode */
+			if (X509_STORE_set_flags(store, crlflags) != 1)
+			{
+				throw ModuleException("Unable to set X509 CRL flags");
+			}
+		}
+
+
 		long GetDefaultContextOptions() const
 		{
 			return ctx_options;
@@ -358,6 +397,12 @@ namespace OpenSSL
 				ERR_print_errors_cb(error_callback, this);
 				ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT, "Can't read CA list from %s. This is only a problem if you want to verify client certificates, otherwise it's safe to ignore this message. Error: %s", filename.c_str(), lasterr.c_str());
 			}
+
+			// Load the CRLs.
+			std::string crlfile  = tag->getString("crlfile");
+			std::string crlpath  = tag->getString("crlpath");
+			std::string crlmode  = tag->getString("crlmode", "chain");
+			ctx.SetCRL(crlfile, crlpath, crlmode);
 
 			clictx.SetVerifyCert();
 			if (tag->getBool("requestclientcert", true))


clinew:
Tested with https://github.com/clinew/inspircdtests/blob/master/ssl_crl.sh (WARNING: Test is tailored for my system, do not run on non-recoverable instance).

clinew:
I also have a 2.0 branch with this, the OpenSSL logging, and SaberUK's SHA-256 fingerprint patch: https://github.com/clinew/inspircd/commits/insp20+openssl+

clinew:
Updated version checks for NULL X509_STORE*, which should never happen, but who trusts OpenSSL?  ;)
Also calls the ERR_* family of functions around the X509_STORE_load_location() call and supplies an additional error string to the ModuleException message, which successfully identifies at least EPERM and ENOENT errors.

src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -204,6 +204,47 @@ namespace OpenSSL
 			return SSL_CTX_load_verify_locations(ctx, filename.c_str(), 0);
 		}
 
+		void SetCRL(const std::string& crlfile, const std::string& crlpath, const std::string& crlmode)
+		{
+			unsigned long crlflags;
+			X509_STORE* store;
```
SaberUK: This should be declared where it is actually used (just before 223?) rather than at the top.

src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -204,6 +204,47 @@ namespace OpenSSL
...
+			if (crlmode == "chain")
+			{
+				crlflags = X509_V_FLAG_CRL_CHECK | X509_V_FLAG_CRL_CHECK_ALL;
+			}
+			else if (crlmode == "leaf")
+			{
+				crlflags = X509_V_FLAG_CRL_CHECK;
```
SaberUK: This can be the default value and you can `|= X509_V_FLAG_CRL_CHECK_ALL` in the "chain" block and then you can do != "leaf" instead of using an else block.

src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -364,6 +405,12 @@ namespace OpenSSL
 				ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT, "Can't read CA list from %s. This is only a problem if you want to verify client certificates, otherwise it's safe to ignore this message. Error: %s", filename.c_str(), lasterr.c_str());
 			}
 
+			// Load the CRLs.
+			std::string crlfile  = tag->getString("crlfile", "");
```
SaberUK: Empty strings are the default for `getString` so there is no need to pass them.

clinew:
Rebased on master, updated, and tested.

SaberUK:
Which OpenSSL versions has this been tested against?

src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -200,6 +200,45 @@ namespace OpenSSL
...
+			else if (crlmode != "leaf")
+			{
+				throw ModuleException("Unknown mode '" + crlmode + "'; expected either 'chain' (default) or 'leaf'");
+			}
+
+			/* Load CRL files */
+			X509_STORE* store = SSL_CTX_get_cert_store(ctx);
+			if (!store)
+			{
+				throw ModuleException("Unable to get X509_STORE from SSL context; this should never happen");
```
SaberUK: Does OpenSSL give an error in this case that you can recover using `ERR_get_error`?

src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -200,6 +200,45 @@ namespace OpenSSL
...
+				crlpath.empty() ? NULL : crlpath.c_str()))
+			{
+				int err = ERR_get_error();
+				throw ModuleException("Unable to load CRL file '" + crlfile + "' or CRL path '" + crlpath + "': '" + (err ? ERR_error_string(err, NULL) : "unknown") + "'");
+			}
+
+			/* Set CRL mode */
+			if (X509_STORE_set_flags(store, crlflags) != 1)
+			{
+				throw ModuleException("Unable to set X509 CRL flags");
```
SaberUK: The same as above for here.

src/modules/extra/m_ssl_openssl.cpp
clinew: If it does, it isn't documented in `SSL_CTX_get_cert_store(3)`.  Want me to dig into the source?

src/modules/extra/m_ssl_openssl.cpp
clinew: Likewise.

clinew:
openssl-1.0.2m.  If you'd like I could see about testing with the 1.1.x series.

src/modules/extra/m_ssl_openssl.cpp
SaberUK: If its not documented then it probably doesn't exist.

SaberUK:
Testing that it at least builds with 1.1 would be useful, thanks!

B00mX0r:
Confirmed that this build worked on a test machine.

```~/inspircd/build$ openssl version
OpenSSL 1.1.0g  2 Nov 2017
```

```~/inspircd/build$ ldd ./modules/m_ssl_openssl.so
	linux-vdso.so.1 (0x00007fff64fa6000)
	libssl.so.1.1 => /usr/lib/x86_64-linux-gnu/libssl.so.1.1 (0x00007f135e35c000)
	libcrypto.so.1.1 => /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1 (0x00007f135decb000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f135db4c000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f135d839000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f135d622000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f135d27f000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f135d07b000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f135ce5d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f135e7d8000)```

**Comment by SaberUK** [comment]
Thanks. I'll merge this when i've had a chance to look into adding support for this to the GnuTLS and mbedTLS backends.

**Comment by clinew** [comment]
Sadly the 1.1.x story turns out to be a bit more complex.  My initial attempts to build it met with the following error:
```
/var/tmp/portage/net-irc/inspircd-3.0.0/work/inspircd-3.0.0/src/modules/m_ssl_openssl.cpp:108:14: error: 'DH_free' was not declared in this scope
    DH_free(dh);
```
I was able to hack around it by including `<openssl/dh.h>` in the header for `m_ssl_openssl.cpp`, but another error regarding `X509_get_not_after()` came up next, and the errors seemed off anyways.  Turns out that looking at `/usr/include/openssl/ssl.h` (version 1.1.x) has the following snippet:
```
# if OPENSSL_API_COMPAT < 0x10100000L
#  include <openssl/x509.h>
#  include <openssl/crypto.h>
#  include <openssl/lhash.h>
#  include <openssl/buffer.h>
# endif
```
Setting the value to `0x10000000L` during InspIRCd compilation didn't help, though.  Eventually I managed to find this webpage https://wiki.openssl.org/index.php/OpenSSL_1.1.0_Changes .  It turns out that this must be set during *OpenSSL* compile time, via `--enable-deprecated`.  My distro was building OpenSSL with `--disable-deprecated`, thus causing the errors; switching it to `--enable-deprecated` fixed the issue; note that I did not need to define `OPENSSL_API_COMPAT` when I later built InspIRCd.

The CRL tests did work once I got it built, though, and it appears the `X509_STORE_load_locations()` is *not* deprecated, so that's a plus.

TL;DR: The InspIRCd module won't compile against the OpenSSL 1.1.x library unless it was built with `--enable-deprecated`.

clinew:
Oh and I rebased on top of 3.0.0-a4 while I was at it; no conflicts occurred.

SaberUK:
That error is unrelated to this. Can you open another issue for it so I can look into it later? Thanks.

PR #1407, opened 2017-10-22
Helpop Updates

genius3000:
#### Extbans
* Add character 'n' - connect class ban (m_classban) [a54497e]
* Move character 'U' from the Acting list to the Matching list as it matches and can be used with other acting bans. Also reword it to read "Matches" and have "module" at the end like the rest.


SaberUK:
Thanks!

PR #1410, opened 2017-11-04
Fix CMD_RESTART with relative paths as argv[0].

jcjordyn130:


SaberUK:
Thanks for the patch!

On what platforms and configurations (i.e. home directory, system-wide, etc) have you tested this?

jcjordyn130:
You're welcome for the patch, I've tested it on Void Linux (x86_64, glibc) with an in-tree install.

SaberUK:
Merged, thanks again! 👍

jcjordyn130:
You're welcome again 👍

PR #1413, opened 2017-11-15
Fix referring to sqloper and sqlauth as extras in modules.conf.example.

B00mX0r:
This fixes a docs issue that refers to m_sqloper and m_sqlauth both as extras modules, despite the fact that they are not.


SaberUK:
Merged, thanks!

Issue #1416, opened 2017-11-25
Setting CloakMode on a remote user does not show the mode remotely


B00mX0r:
Tested on 3.0.0a3. Not tested on 2.0.

When setting CloakMode (umode +x) on a remote user (i.e. /SAMODE Remote1 +x), only Remote1's server will see the user as having umode +x.

Example:
3 linked servers (A is hub): A, B, and C.
3 users: UserA, UserB, and UserC, on each server respectively. UserA has umode +o.
1) UserA: /SAMODE UserB +x
2) UserA: /whois UserB: will see their cloaked DisplayHost, but will not see them as having umode +x.
3) UserB: /whois UserB: will see their cloaked DisplayHost, and will see them as having umode +x.
4) UserC: /whois UserB: will see their cloaked DisplayHost, but will not see them as having umode +x.

The issue does not arise when the setter is local, regardless of whether it occurred through SAMODE via a local operator or from the user enabling cloaking on themselves.

Theory: In m_cloaking, we unset +x whenever there is a host change (https://github.com/inspircd/inspircd/blob/master/src/modules/m_cloaking.cpp#L305). We then expect that if the host change is because we are setting the cloaked host, +x will be set again on line 107 (https://github.com/inspircd/inspircd/blob/master/src/modules/m_cloaking.cpp#L107). However, the bug arises because we only reset +x locally on line 107. Line 66 (https://github.com/inspircd/inspircd/blob/master/src/modules/m_cloaking.cpp#L66) fails to resolve this for remote users, as the host change happens after the /samode broadcast. Therefore, line 66 will set the host properly, but then it will immediately be unset again by line 305.

PR #1417, opened 2017-11-25
Fix m_cloaking umode desync

B00mX0r:
Servers will now correctly show whether a remote user has +x or not.

This fixes https://github.com/inspircd/inspircd/issues/1416.

### DIFF
--- a/src/modules/m_cloaking.cpp
+++ b/src/modules/m_cloaking.cpp
@@ -42,14 +42,18 @@ static const char base32[] = "0123456789abcdefghijklmnopqrstuv";
 class CloakUser : public ModeHandler
 {
  public:
+	bool active;
 	LocalStringExt ext;
 	std::string debounce_uid;
 	time_t debounce_ts;
 	int debounce_count;
 
 	CloakUser(Module* source)
-		: ModeHandler(source, "cloak", 'x', PARAM_NONE, MODETYPE_USER),
-		ext("cloaked_host", ExtensionItem::EXT_USER, source), debounce_ts(0), debounce_count(0)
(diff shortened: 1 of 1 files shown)


src/modules/m_cloaking.cpp
```diff
@@ -63,6 +64,7 @@ class CloakUser : public ModeHandler
 		 */
 		if (!user)
 		{
+			active = IS_LOCAL(source) ? adding : !adding; // Remote setters broadcast mode before host while local setters do the opposite, so this takes that into account
```
SaberUK: Can you put the comment on the previous line please.

B00mX0r:
@SaberUK fixed the line comment. Anything else need changing?

src/modules/m_cloaking.cpp
```diff
@@ -42,6 +42,7 @@ static const char base32[] = "0123456789abcdefghijklmnopqrstuv";
 class CloakUser : public ModeHandler
 {
  public:
+	bool active;
```
SaberUK: Can you initialise this to false in the constructor as non-local primitives do not have a guaranteed default value. Also, while you are altering the constructor can you alter it to be in the correct style, i.e.

```
ModuleCloaking()
	: active(false)
	, cu(this)
	// rest of initialisers here in the same style
{
}
````

rburchell: (not related to this change)

It might be nice to pick up initialization of members in the headers instead. That would require being C++11-only, but that ought to be acceptable by now, anyway.

B00mX0r:
@SaberUK Added active bool to m_cloaking's constructors and fixed their style as requested.

SaberUK:

Issue #425, opened 2013-02-15
[2.0] m_override reports non-overrides


satmd:
m_override will report a /mode #channel b regardless of permissions required.

The issue stems from m_override, line 108 being overly strict and not taking into account individual modes' levelrequired.

Suggestion: Before line 108, filter out modes that could have been done without override correctly.

B00mX0r:
It appears to be for any listmode (+g, +e, etc). I noticed that you didn't specify this.

SaberUK:
I'm having trouble replicating this against the current 2.0 git. Is it still an issue for you?

genius3000:
I just updated my net to current 2.0 git a few days ago and still see this. Joining user's client requests ban list (allowed) but as override capable oper, it's shown as override.  
I believe this is what #1418 is for.

Robby-:
I just ran into this issue as well. One network logged an override, while the other didn't. It turns out the difference was in the `<type:override>` setting. On one network its set as `<type:override="*">` and on the other its set as `<type:override="INVITE KEY LIMIT BANWALK">` The first one with `*` logs the override, while the latter one with the list of overrides does not log an override when doing a `/MODE #channel b`.

PR #1418, opened 2017-11-26
Fix oper override message when viewing a list mode

B00mX0r:
Oper override would execute for MODE every time a mode in the command had an "adding" permission higher than the user's current permission. This would result in an oper override announcement showing when an oper simply did /mode #chan +b.

Because hideopers gives channels/auspex permission to bypass it, this patch does not use oper override to override the viewing of list modes with hideopers in place. 

This fixes https://github.com/inspircd/inspircd/issues/425.


B00mX0r:
@SaberUK My sincere apologies. The PR description was wrong, as I accidentally referred to the incorrect module. I have rewritten the description to properly reflect the changes that this PR implements.

include/modules/hidelist.h
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#pragma once
+
+#include "event.h"
+
+class HidelistEventListener : public Events::ModuleEventListener
```
SaberUK: HideListEventListwnwe.

include/modules/hidelist.h
```diff
@@ -0,0 +1,37 @@
...
+{
+ public:
+	HidelistEventListener(Module* mod)
+		: ModuleEventListener(mod, "event/hidelist")
+	{
+	}
+
+	/** Called whenever a /WHOIS is performed by a local user.
```
SaberUK: Update the code documentation too.

include/modules/hidelist.h
```diff
@@ -0,0 +1,37 @@
...
+	 * @param whois Whois context, can be used to send numerics
+	 */
+	virtual ModResult OnListDenied(User* user, Channel* chan, const std::string& modename) = 0;
```
SaberUK: Pass a mode handler (or maybe a mode reference?) here.

B00mX0r: We don't have this information in the ModeWatcher class. Also, the mode name just needs to be passed for the m_override snomask announcement.

B00mX0r:
I renamed OnListDenied to OnListDeny to better reflect that the denial is dependent on the MOD_RESULT from this call.

B00mX0r:
Targeting v3 instead of v2 because it's in RC stage now

Issue #480, opened 2013-04-09
m_cloaking should normalize the case of FQDNs


ruok5:
`m_cloaking` produces different results for FQDNs with differences only in capitalization:

`1-2-3-4-SFBA.hfc.isp.tld` --> `network-e4d099e1.hfc.isp.tld`
`1-2-3-4-sfba.hfc.isp.tld` --> `network-d7bfc48a.hfc.isp.tld`

Which, rare as it may be, of course breaks things.

It is my opinion that the solution should be to _lowercase_ the FQDN before generating the cloaked version, as it is by far most common for FQDNs to be exclusively lowercase letters. Anything that relies on matching cloaked hostnames would be completely broken if uppercase were used. I also believe that this change in behavior should be made optional via the module's configuration.

PR #1419, opened 2017-11-26
Add support for case-insensitive host cloaking (on by default)

B00mX0r:
This PR allows for all hosts to be lowercased before m_cloaking runs on them. Case-sensitivity can be re-enabled via a config option for the cloak tag.

This resolves https://github.com/inspircd/inspircd/issues/480.

### DIFF
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -511,12 +511,14 @@
 #                                                                     #
 # The methods use a single key that can be any length of text.        #
 # An optional prefix may be specified to mark cloaked hosts.          #
+# Casesensitive controls whether hosts are case-sensitive.            #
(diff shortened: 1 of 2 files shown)


SaberUK:
Can you enable this in the config but disable it in the code? That way people who are upgrading from 2.0 won't have their configs be broken.

Also, can you only lowercase the parts of the host which are being hashed? That way people can get their fancy hostname casing whilst preserving cloaking. See SegmentCloak for details.

I'm also concerned about potential desyncs. I've got some upcoming changes to module link data which might be of use with this.

B00mX0r:
I implemented both parts of this. Also, I'm confused about your concern towards desyncs. The cloaked host is broadcasted over spanningtree, and therefore each server will correctly have the same host for each cloaked user. It is true that the same user from the same IP on different servers can have a different cloaked host if one does enables case sensitivity, but there won't be a desync.

SaberUK:
>It is true that the same user from the same IP on different servers can have a different cloaked host if one does enables case sensitivity

Yeah this is what i meant. Apologies for not being clear.

Issue #1161, opened 2016-03-24
m_chanfilter: Oper override


Brainscrewer:
Currently when chanmode +g is set for certain words, for example `*#*`, IRC Operators are also affected. 

In my opinion IRC Operators should be able to override this if they have enough access as defined in the configuration.

B00mX0r:
This is already implemented in v3.0, as m_chanfilter calls to OnCheckExemption for "filter." Use m_exemptchanops.

SadieCat:
Since commit e2b50c3551 server operators with the `channels/ignore-chanfilter` oper privilege are exempt from channel filters.

PR #1422, opened 2017-11-27
Add CHANMSG oper override

B00mX0r:
This allows for opers to override loaded modules blocking them from sending a message to a channel, such as m_chanfilter or m_muteban.

This implements and resolves https://github.com/inspircd/inspircd/issues/1161.

### DIFF
--- a/src/modules/m_override.cpp
+++ b/src/modules/m_override.cpp
@@ -40,6 +40,7 @@ class Override : public SimpleUserModeHandler
 
 class ModuleOverride : public Module
 {
+	bool active;
 	bool RequireKey;
 	bool NoisyOverride;
 	bool UmodeEnabled;
@@ -78,7 +79,8 @@ class ModuleOverride : public Module
 
  public:
 	ModuleOverride()
-		: UmodeEnabled(false)
+		: active(false)
+		, UmodeEnabled(false)
 		, ou(this)
 		, topiclock(this, "topiclock")
 		, inviteonly(this, "inviteonly")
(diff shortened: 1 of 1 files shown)


src/modules/m_override.cpp
```diff
@@ -199,6 +201,46 @@ class ModuleOverride : public Module
...
+			active = false;
+
+			// If there is something stopping us from sending our privmsg to the channel then we override
+			if (MOD_RESULT == MOD_RES_DENY)
+			{
+				Channel* channel = static_cast<Channel*>(dest);
+				ServerInstance->SNO->WriteGlobalSno('v',user->nick+" used oper override to send a message on "+channel->name);
+				return MOD_RES_ALLOW;
+			}
+			return MOD_RES_ALLOW; // We already ran FIRST_MOD_RESULT, so if there was no MOD_RES_DENy then they were all passthrus, so there's no point in re-running it anyway
```
B00mX0r: I intentionally kept this line instead of removing it (which would work because of line 225) to clarify the line comment on 225. I can remove this if people want me to.

genius3000:
Just to note it:
> override anything blocking them from sending a message to a channel  

isn't entirely true as there are still 3 checks in core_privmsg prior to calling OnUserPreMessage; See [core_privmsg.cpp:L129-L151](https://github.com/inspircd/inspircd/blob/master/src/coremods/core_privmsg.cpp#L129-L151)
The checks are only ran when the user's prefix value is less than the voice value, but that leaves 'noextmsgmode' (+n) without any override.

B00mX0r:
Updated commit message, config file, and PR description to reflect that this only applies to privmsg restrictions from external modules. As agreed to in IRC, moving the definitions of +n, +t, and +b are not in the scope of this PR, though are probably good discussion points for a different issue.

SaberUK:
I'm not sure about this. I think that it would be better to extend support for exemptchanops to more modules (which needs to happen anyway) and then add exemptchanops privileges to oper blocks.

SaberUK:
Can you resolve the merge conflicts in this? Thanks.

B00mX0r:
@SaberUK conflicts resolved as requested.

B00mX0r:
Just a note that as I'm working on moving this to OnCheckExemption, I noticed that the events class doesn't currently have a way to set hook priorities that I can see.

SaberUK:
Marked as blocked until I manage to extract the priority logic from the cross-module event system.

B00mX0r:
@SaberUK Is this slated for 4.0?

Issue #363, opened 2012-11-14
User mode to enable override


CuleX:
Currently, m_override just enables overrides for all opers with the privs to override what they're allowed to.

My proposal is that a new user mode (+O?) is added. If and only if an oper has that mode _and_ override privileges in his/her O:line, overrides should be possible.

Rationale: Currently the override module is spammy (reports overrides for actions a user can be validly taking, like fetching a ban list) and it's very easy to produce an accidental override. Having an user mode would completely prevent accidents.

Additionally, it might be worth a thought to have a timer to remove the override-enabling usermode after a configurable amount of time (+ disabling the timer via configuration file).

PR #1423, opened 2017-11-27
Add optional oper override umode +O

B00mX0r:
This adds the optional user mode "O". If the user mode is enabled, then only opers who have given it to themselves may use oper override. If the user mode is disabled (default if config option left blank), then override works normally.

Requiring a usermode for override to work requires opers to actively enable overriding in order to prevent accidental overrides. This allows opers to continue to act as normal users when not doing oper-related tasks. In my opinion, this was worth it because override, unlike SA*, is focused solely on the operator rather than other users (SAJOIN can be used on others).

This resolves https://github.com/inspircd/inspircd/issues/363.

### DIFF
--- a/src/modules/m_override.cpp
+++ b/src/modules/m_override.cpp
@@ -27,10 +27,23 @@
 #include "inspircd.h"
 #include "modules/invite.h"
 
+class Override : public SimpleUserModeHandler
+{
+ public:
+	Override(Module* Creator) : SimpleUserModeHandler(Creator, "override", 'O')
+	{
+		oper = true;
+		if (!ServerInstance->Config->ConfValue("override")->getBool("enableumode"))
+			DisableAutoRegister();
+	}
+};
+
 class ModuleOverride : public Module
 {
 	bool RequireKey;
 	bool NoisyOverride;
+	bool UmodeEnabled;
+	Override ou;
 	ChanModeReference topiclock;
 	ChanModeReference inviteonly;
 	ChanModeReference key;
@@ -65,7 +78,9 @@ class ModuleOverride : public Module
 
  public:
 	ModuleOverride()
-		: topiclock(this, "topiclock")
+		: UmodeEnabled(false)
+		, ou(this)
+		, topiclock(this, "topiclock")
 		, inviteonly(this, "inviteonly")
 		, key(this, "key")
 		, limit(this, "limit")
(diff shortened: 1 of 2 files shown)


src/modules/m_override.cpp
```diff
@@ -84,6 +98,17 @@ class ModuleOverride : public Module
 		ConfigTag* tag = ServerInstance->Config->ConfValue("override");
 		NoisyOverride = tag->getBool("noisy");
 		RequireKey = tag->getBool("requirekey");
+		bool ShouldEnableUmode = tag->getBool("enableumode", false);
```
SaberUK: `false` is the default value for getBool so you don't need to pass it.

src/modules/m_override.cpp
```diff
@@ -84,6 +98,17 @@ class ModuleOverride : public Module
...
+		if (ShouldEnableUmode && !UmodeEnabled)
+		{
+			ServerInstance->Modules->AddService(ou);
```
SaberUK: Changing this on rehash is problematic because it may cause desyncs if one server has the mode and another does not. You should enable it in the mode constructor at load time like m_redirect does.

B00mX0r:
Resolved both comments. Note that a user can still cause a desync by reloading the module and changing their enableumode config option, which is also true for m_redirect. This may be a better issue to bring up with m_spanningtree for an OnServiceAdd-type hook.

src/modules/m_override.cpp
```diff
@@ -27,10 +27,23 @@
 #include "inspircd.h"
 #include "modules/invite.h"
 
+class OverrideMode : public SimpleUserModeHandler
+{
+ public:
+	OverrideMode(Module* Creator) : SimpleUserModeHandler(Creator, "overridemode", 'O')
```
SaberUK: Just `override` is fine for the mode name.

src/modules/m_override.cpp
```diff
@@ -94,6 +110,9 @@ class ModuleOverride : public Module
 	bool CanOverride(User* source, const char* token)
 	{
 		std::string tokenlist = source->oper->getConfig("override");
+		// If we require oper override umode (+O) but it is not set
```
SaberUK: This should be before the getConfig call.

src/modules/m_override.cpp
```diff
@@ -27,10 +27,23 @@
 #include "inspircd.h"
 #include "modules/invite.h"
 
+class Override : public SimpleUserModeHandler
+{
+ public:
+	Override(Module* Creator) : SimpleUserModeHandler(Creator, "Override", 'O')
```
SaberUK: Mode names should be all lower case, i.e. `override`.

src/modules/m_override.cpp
```diff
@@ -27,10 +27,23 @@
...
+	{
+		oper = true;
+		if (ServerInstance->Config->ConfValue("override")->getBool("enableumode"))
```
SaberUK: I think this is inverted? Surely you should only call `DisableAutoRegister()` if the mode is disabled?

B00mX0r:
@SaberUK Thanks for catching those two issues. I fixed them both.

PR #1428, opened 2017-12-02
Remove extra double quote (") from modules.conf

Chewsterchew:
It's something small but noteworthy nonetheless


SaberUK:

Issue #1431, opened 2017-12-03
Modules with optional modes cause segfault on config change + reloadmodule


B00mX0r:
If a previously-loaded and set mode is unloaded as a result of a ReloadModule (i.e. https://github.com/inspircd/inspircd/pull/1423 or m_redirect before https://github.com/inspircd/inspircd/pull/466), InspIRCd will segfault. It never checks if the mode that it sets exists, so it calls the NeedsParam function, which then attempts to run a switch statement on an invalid pointer.

Excerpt of the problem using the currently-unmerged m_override with umode:
```Sun Dec 03 2017 21:54:44 USERINPUT: C[464AAAAAA] I RELOADMODULE m_override.so
Sun Dec 03 2017 21:54:44 CULLLIST: classbase::+ @0x5555559831c0
Sun Dec 03 2017 21:54:44 m_cap: OnReloadModuleSave()
Sun Dec 03 2017 21:54:44 core_reloadmodule: Saved data about 1 users 0 chans 2 modules
Sun Dec 03 2017 21:54:44 m_spanningtree: S[19] O :464 METADATA * modules :-m_override.so
Sun Dec 03 2017 21:54:44 m_spanningtree: S[21] O :464 METADATA * modules :-m_override.so
Sun Dec 03 2017 21:54:44 USEROUTPUT: C[464AAAAAA] O :vpn.priv.succ.in MODE B00mX0r -O
Sun Dec 03 2017 21:54:44 MODULE: Module m_override.so unloaded
Sun Dec 03 2017 21:54:44 CULLLIST: Deleting 14ModuleOverride @0x55555592d4c0
Sun Dec 03 2017 21:54:44 CULLLIST: classbase::-14ModuleOverride @0x55555592d4c0
Sun Dec 03 2017 21:54:44 CULLLIST: classbase::~ @0x55555592d500
Sun Dec 03 2017 21:54:44 CULLLIST: classbase::~ @0x55555592d4c0
Sun Dec 03 2017 21:54:44 CULLLIST: classbase::~ @0x555555991b00
Sun Dec 03 2017 21:54:44 CULLLIST: classbase::+ @0x555555991b00
Sun Dec 03 2017 21:54:44 CULLLIST: classbase::+ @0x55555592d4c0
Sun Dec 03 2017 21:54:44 CULLLIST: classbase::+ @0x55555592d500
[...]

PR #1432, opened 2017-12-03
Ensure ModeHandler doesn't try to process a non-existent mode

B00mX0r:
This resolves the segfault described in #1431.

### DIFF
--- a/src/mode.cpp
+++ b/src/mode.cpp
@@ -449,6 +449,9 @@ unsigned int ModeParser::ProcessSingle(User* user, Channel* targetchannel, User*
 
 		Modes::Change& item = *i;
 		ModeHandler* mh = item.mh;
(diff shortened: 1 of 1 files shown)


src/mode.cpp
```diff
@@ -449,6 +449,8 @@ unsigned int ModeParser::ProcessSingle(User* user, Channel* targetchannel, User*
 
 		Modes::Change& item = *i;
 		ModeHandler* mh = item.mh;
+		if (mh == NULL)
```
SaberUK: Can you add a comment before this describing the contexts in which this might be null? Thanks.

B00mX0r:
Added the comment. Like it says, part of this patch is also to serve as a defensive gesture against any errors that may come up in the future as the result of someone attempting to refactor a module in a way that falsely assumes a mode exists.

SaberUK:
Whilst this is good would it be also possible for you to look into exactly why the mode change is being processed by core_reloadmodule?

PR #1442, opened 2017-12-17
Show server versions in /MAP for opers

### LINKED ISSUE #1203: Show server version when doing /map
As an oper, I want to be able to running version of the servers listed when doing /map.
Useful to verify that you are running the correct version without doing /quote version on all servers.

B00mX0r:
For opers, /MAP will now show the server versions for leafs next to their UIDs. This resolves #1203.

Example output:
```
[12/16/2017 11:54:54 PM] 1.test.net (464 InspIRCd-3.0.0-b30230d39)       1 [33.33%] [Up: 29s]
[12/16/2017 11:54:54 PM] |-2.test.net (789 InspIRCd-3.0.0-94b39fd3e)    1 [33.33%] [Up: 23s]
[12/16/2017 11:54:54 PM] `-3.test.net (799 InspIRCd-3.0.0-94b39fd3e)     1 [33.33%] [Up: 18s]
[12/16/2017 11:54:54 PM] 3 servers and 3 users, average 1.00 users per server
[12/16/2017 11:54:54 PM] End of /MAP
```


SaberUK:
This is kinda icky. Can we introduce a new piece of SINFO data which just contains the version number instead?

src/modules/m_spanningtree/treeserver.cpp
```diff
@@ -37,6 +37,7 @@ TreeServer::TreeServer()
 	, Parent(NULL), Route(NULL)
 	, VersionString(ServerInstance->GetVersionString())
 	, fullversion(ServerInstance->GetVersionString(true))
+	, rawversion(INSPIRCD_VERSION)
```
SaberUK: This will make servers which didn't send this show the local version. This seems like it would be undesirable behaviour. Maybe default to something else instead?

PR #1444, opened 2017-12-19
Add snomask alert on m_filter quit/part G-Line

B00mX0r:
There was an oversight in https://github.com/inspircd/inspircd/pull/1264, where G-Lines as a result from a PART/QUIT message filter were not broadcast to opers via snomask.

### DIFF
--- a/src/modules/m_filter.cpp
+++ b/src/modules/m_filter.cpp
@@ -440,6 +440,7 @@ ModResult ModuleFilter::OnPreCommand(std::string &command, std::vector<std::stri
 			{
 				/* Note: We gline *@IP so that if their host doesnt resolve the gline still applies. */
 				GLine* gl = new GLine(ServerInstance->Time(), f->gline_time, ServerInstance->Config->ServerName.c_str(), f->reason.c_str(), "*", user->GetIPString());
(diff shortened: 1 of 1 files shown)


src/modules/m_filter.cpp
```diff
@@ -440,6 +440,7 @@ ModResult ModuleFilter::OnPreCommand(std::string &command, std::vector<std::stri
 			{
 				/* Note: We gline *@IP so that if their host doesnt resolve the gline still applies. */
 				GLine* gl = new GLine(ServerInstance->Time(), f->gline_time, ServerInstance->Config->ServerName.c_str(), f->reason.c_str(), "*", user->GetIPString());
+				ServerInstance->SNO->WriteGlobalSno('a', "FILTER: " + user->nick + " had their " + command + " message filtered and was G-Lined: " + f->reason);
```
SaberUK: "Wibble had their PRIVMSG message filtered" sounds a bit weird to me?

B00mX0r: Will only happen for QUIT and PART.

SaberUK: Yes, I misread. Thanks for clarifying.

Issue #1122, opened 2015-12-28
ERR_NOSUCHNICK used for both nicks and channels


attilamolnar:
See #1117

PR #1446, opened 2017-12-19
Fixed misc. instances of ERR_NOSUCHNICK instead of channel numerics

B00mX0r:
Per #1122, this PR replaces multiple instances of ERR_NOSUCHNICK with the proper channel numerics. This includes mostly ERR_NOSUCHCHANNEL, but also a couple of ERR_NOTONCHANNEL instances.

### DIFF
--- a/include/numeric.h
+++ b/include/numeric.h
@@ -88,8 +88,7 @@ class Numeric::Numeric
 
 namespace Numerics
 {
-	/** ERR_NOSUCHNICK numeric
-	 */
+	/** Builder for the ERR_NOSUCHNICK numeric. */
 	class NoSuchNick : public Numeric::Numeric
 	{
 	 public:
@@ -97,7 +96,19 @@ namespace Numerics
 			: Numeric(ERR_NOSUCHNICK)
 		{
 			push(nick);
-			push("No such nick/channel");
+			push("No such nick");
+		}
+	};
+
+	/** Builder for the ERR_NOSUCHCHANNEL numeric. */
+	class NoSuchChannel : public Numeric::Numeric
+	{
+	 public:
+		NoSuchChannel(const std::string& chan)
+			: Numeric(ERR_NOSUCHCHANNEL)
+		{
+			push(chan);
+			push("No such channel");
 		}
 	};
 }
--- a/src/coremods/core_channel/cmd_invite.cpp
+++ b/src/coremods/core_channel/cmd_invite.cpp
@@ -56,9 +56,14 @@ CmdResult CommandInvite::Handle (const std::vector<std::string>& parameters, Use
 				timeout = ConvToInt(parameters[3]);
 		}
 
-		if ((!c) || (!u) || (u->registered != REG_ALL))
+		if (!c)
 		{
-			user->WriteNumeric(Numerics::NoSuchNick(c ? parameters[0] : parameters[1]));
+			user->WriteNumeric(Numerics::NoSuchChannel(parameters[1]));
+			return CMD_FAILURE;
+		}
+		if ((!u) || (u->registered != REG_ALL))
+		{
+			user->WriteNumeric(Numerics::NoSuchNick(parameters[0]));
 			return CMD_FAILURE;
 		}
 
--- a/src/coremods/core_channel/cmd_kick.cpp
+++ b/src/coremods/core_channel/cmd_kick.cpp
@@ -42,9 +42,14 @@ CmdResult CommandKick::Handle (const std::vector<std::string>& parameters, User 
 	else
 		u = ServerInstance->FindNick(parameters[1]);
 
-	if ((!u) || (!c) || (u->registered != REG_ALL))
+	if (!c)
 	{
-		user->WriteNumeric(Numerics::NoSuchNick(c ? parameters[1] : parameters[0]));
+		user->WriteNumeric(Numerics::NoSuchChannel(parameters[0]));
+		return CMD_FAILURE;
+	}
+	if ((!u) || (u->registered != REG_ALL))
+	{
+		user->WriteNumeric(Numerics::NoSuchNick(parameters[1]));
 		return CMD_FAILURE;
 	}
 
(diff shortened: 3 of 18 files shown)


src/coremods/core_channel/cmd_kick.cpp
```diff
@@ -42,9 +42,14 @@ CmdResult CommandKick::Handle (const std::vector<std::string>& parameters, User
 	else
 		u = ServerInstance->FindNick(parameters[1]);
 
-	if ((!u) || (!c) || (u->registered != REG_ALL))
+	if (!c)
 	{
-		user->WriteNumeric(Numerics::NoSuchNick(c ? parameters[1] : parameters[0]));
+		user->WriteNumeric(ERR_NOSUCHCHANNEL, parameters[0], "No such channel");
```
SaberUK: Can you add something like `Numerics::NoSuchNick` but for channels and use that instead?

src/coremods/core_channel/cmd_topic.cpp
```diff
@@ -38,15 +38,15 @@ CmdResult CommandTopic::HandleLocal(const std::vector<std::string>& parameters,
...
+		user->WriteNumeric(ERR_NOSUCHCHANNEL, parameters[0], "No such channel");
 		return CMD_FAILURE;
 	}
 
 	if (parameters.size() == 1)
 	{
 		if ((c->IsModeSet(secretmode)) && (!c->HasUser(user)))
 		{
-			user->WriteNumeric(Numerics::NoSuchNick(c->name));
+			user->WriteNumeric(ERR_NOTONCHANNEL, c->name, "You're not on that channel");
```
SaberUK: This should be 'no such channel' to avoid revealing the existence of secret channels..

SaberUK:
Now that its just being used for channels the trailing parameter of `Numerics::NoSuchNick` needs to be updated.

SaberUK:
(Oops, wrong button)

src/coremods/core_user/cmd_mode.cpp
```diff
@@ -45,7 +45,10 @@ CmdResult CommandMode::Handle(const std::vector<std::string>& parameters, User*
 
 	if ((!targetchannel) && (!targetuser))
 	{
-		user->WriteNumeric(Numerics::NoSuchNick(target));
+		if (target[0] == '#')
```
SaberUK: Can you use IsChannel here? It will be more future proof for when we have multiple channel types.

include/numeric.h
```diff
@@ -97,7 +97,18 @@ namespace Numerics
 			: Numeric(ERR_NOSUCHNICK)
 		{
 			push(nick);
-			push("No such nick/channel");
+			push("No such nick");
+		}
+	};
+
+	class NoSuchChannel : public Numeric::Numeric
```
SaberUK: Can you add documentation for this class? Something like this:

```
/** Builder for the ERR_NOSUCHNICK numeric. */
```

src/coremods/core_channel/cmd_invite.cpp
```diff
@@ -56,9 +56,14 @@ CmdResult CommandInvite::Handle (const std::vector<std::string>& parameters, Use
 				timeout = ConvToInt(parameters[3]);
 		}
 
-		if ((!c) || (!u) || (u->registered != REG_ALL))
+		if (!c)
 		{
-			user->WriteNumeric(Numerics::NoSuchNick(c ? parameters[0] : parameters[1]));
+			user->WriteNumeric(Numerics::NoSuchChannel(parameters[1]));
```
SaberUK: This will always result in an error as a target cannot be both a user and a channel.

src/coremods/core_channel/cmd_kick.cpp
```diff
@@ -42,9 +42,14 @@ CmdResult CommandKick::Handle (const std::vector<std::string>& parameters, User
...
 		u = ServerInstance->FindNick(parameters[1]);
 
-	if ((!u) || (!c) || (u->registered != REG_ALL))
+	if (!c)
 	{
-		user->WriteNumeric(Numerics::NoSuchNick(c ? parameters[1] : parameters[0]));
+		user->WriteNumeric(Numerics::NoSuchChannel(parameters[0]));
+		return CMD_FAILURE;
+	}
+	if ((!u) || (u->registered != REG_ALL))
```
SaberUK: This will always result in an error as a target cannot be both a user and a channel.

src/coremods/core_channel/cmd_invite.cpp
B00mX0r: The code that this is replacing is: ```if ((!c) || (!u) || (u->registered != REG_ALL))```

Git diffs isn't showing https://github.com/B00mX0r/inspircd/blob/1e6d366930fde3c21f0ebf3485983905ba62491a/src/coremods/core_channel/cmd_invite.cpp#L44,#L49, which is the relevant part that shows how both should be defined.

src/coremods/core_channel/cmd_kick.cpp
B00mX0r: The code that this is replacing is ```if ((!u) || (!c) || (u->registered != REG_ALL))```

Git diffs isn't showing https://github.com/B00mX0r/inspircd/blob/1e6d366930fde3c21f0ebf3485983905ba62491a/src/coremods/core_channel/cmd_kick.cpp#L34,#L43, which is the relevant part that shows how both of them should be defined (you need a valid user to kick and a channel to kick them from).

include/numeric.h
B00mX0r: Added that comment for both NoSuchNick and NoSuchChannel. Thanks!

src/modules/m_remove.cpp
```diff
@@ -74,9 +74,14 @@ class RemoveBase : public Command
...
-		if ((!target) || (target->registered != REG_ALL) || (!channel))
+		if (!channel)
 		{
-			user->WriteNumeric(Numerics::NoSuchNick(channel ? username.c_str() : channame.c_str()));
+			user->WriteNumeric(Numerics::NoSuchChannel(channame.c_str()));
+			return CMD_FAILURE;
+		}
+		if ((!target) || (target->registered != REG_ALL))
+		{
+			user->WriteNumeric(Numerics::NoSuchNick(username.c_str()));
```
SaberUK: There's no need to use c_str() here + above I think?

B00mX0r: Good catch, thanks!

PR #1447, opened 2017-12-21
Separate secret and private channels on whois for non-opers

### LINKED ISSUE #969: All users should have private channels listed separately like what opers see in /whois
When an oper does a /whois on a user, they get the list of channels that user is on separated into two blocks, the second of which contains channels which are private or secret, and is preceded by:

: is on private/secret channels:

When you do a /whois on yourself, or on someone else and you're not an oper, you see all of those channels mixed together, whether they are private or not.

I've seen said by several users recently that it would be convenient to know at a glance which channels they are on that are private.  I think it would be useful to list the channels this same way for all users and not just opers.  Of course, still preserving that you only see channels listed that you can actually see (like you don't get any private channels that another user is on unless you are also on that channel, but it still shows up in the private list if you are on that channel with them).

B00mX0r:
This separates secret channels from non-secret channels when users WHOIS themselves.

This resolves #969.

### DIFF
--- a/src/coremods/core_whois.cpp
+++ b/src/coremods/core_whois.cpp
@@ -37,6 +37,18 @@ enum
 	RPL_CHANNELSMSG = 651
 };
 
+enum SplitWhoisState
+{
+	// Don't split private/secret channels into a separate RPL_WHOISCHANNELS numeric.
+	SPLITWHOIS_NONE,
+
+	// Split private/secret channels into a separate RPL_WHOISCHANNELS numeric.
+	SPLITWHOIS_SPLIT,
+
+	// Split private/secret channels into a separate RPL_WHOISCHANNELS numeric with RPL_CHANNELSMSG to explain the split.
+	SPLITWHOIS_SPLITMSG
+};
+
 class WhoisContextImpl : public Whois::Context
 {
 	Events::ModuleEventProvider& lineevprov;
@@ -75,6 +87,8 @@ class CommandWhois : public SplitCommand
 	void SendChanList(WhoisContextImpl& whois);
 
  public:
+	SplitWhoisState splitwhois;
+
 	/** Constructor for whois.
 	 */
 	CommandWhois(Module* parent)
@@ -125,8 +139,9 @@ class WhoisChanListNumericBuilder : public Numeric::GenericBuilder<' ', false, W
 
 class WhoisChanList
 {
+	const SplitWhoisState& splitwhois;
 	WhoisChanListNumericBuilder num;
-	WhoisChanListNumericBuilder spynum;
+	WhoisChanListNumericBuilder secretnum;
 	std::string prefixstr;
 
 	void AddMember(Membership* memb, WhoisChanListNumericBuilder& out)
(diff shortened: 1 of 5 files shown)


src/coremods/core_whois.cpp
```diff
@@ -177,10 +177,11 @@ void CommandWhois::SendChanList(WhoisContextImpl& whois)
...
 		Membership* memb = *i;
 		Channel* c = memb->chan;
-		/* If the target is the sender, neither +p nor +s is set, or
-		 * the channel contains the user, it is not a spy channel
+		/* If neither +p nor +s is set, or the channel contains the user
+		 * and the user is not running whois on themselves, it is not
+		 # a spy channel
 		 */
-		if ((whois.IsSelfWhois()) || ((!c->IsModeSet(privatemode)) && (!c->IsModeSet(secretmode))) || (c->HasUser(whois.GetSource())))
+		if ((!c->IsModeSet(privatemode) && !c->IsModeSet(secretmode)) || (c->HasUser(whois.GetSource()) && !whois.IsSelfWhois()))
```
SaberUK: It might also be desirable to split private/secret channels when the source is not the target user but is also on the channel. What do you think?

B00mX0r:
Since I couldn't see any harm and could only see benefits with making it so private and secret chans are also separated for other users whom you are in the same channel with, I went ahead and implemented that change.

Before merging, I'd also like to talk a bit about the RPL_CHANNELSMSG numeric to send the "is on private/secret channels" message. Where did this numeric come from? I cannot find any record of it anywhere else, and this is the only place in which it is used. My IRC client does not properly support it (unsurprising since it doesn't seem to originate from anywhere), and so the "is on private/secret channels" message is improperly separated from the rest of the WHOIS.

B00mX0r:
What if we do RPL_WHOISSPECIAL? I think that would make sense given that this is a special WHOIS line, and that this line is already supported by many clients.

SaberUK:
I've made some tweaks to this so that normal users can get split channel lists too. Can you see any issues with the changes i've made? I'm pretty sure they're okay but there might be some bugs I haven't spotted.

B00mX0r:
@SaberUK thanks for the changes. I changed the numeric to RPL_WHOISSPECIAL so clients would not break up the WHOIS reply, optimized an if loop, and removed an extra whitespace.

Other than that, looks good to me. I tested different config variations and oper/user scenarios.

SaberUK:
> I changed the numeric to RPL_WHOISSPECIAL so clients would not break up the WHOIS reply

This numeric was explicitly changed to an unused numeric to avoid collisions like this. Clients that don't currently support it can implement support for it if they want.

src/coremods/core_whois.cpp
```diff
@@ -201,13 +216,13 @@ void CommandWhois::DoWhois(LocalUser* user, User* dest, time_t signon, unsigned
 
 	SendChanList(whois);
 
-	if (!whois.IsSelfWhois() && !ServerInstance->Config->HideServer.empty() && !user->HasPrivPermission("servers/auspex"))
+	if (whois.IsSelfWhois() || ServerInstance->Config->HideServer.empty() || user->HasPrivPermission("servers/auspex"))
```
SaberUK: This is unrelated to this PR so can you revert it for now?

PR #1449, opened 2017-12-21
Fix two minor docs issues

### LINKED ISSUE #1448: [2.0] Some Docs Issues
There are a few docs issues in 2.0 to be fixed.

1. SPLITMSG and YES for operspywhois in inspircd.conf.example have the same description, while "yes" does not split the message.
2. The <channames> tag is uncommented by default in modules.conf.example

B00mX0r:
This fixes the copypasta in the description for operspywhois and properly comments the <channames> and <chanprotect> tags in modules.conf.example. This resolves #1448.

### DIFF
--- a/docs/conf/inspircd.conf.example
+++ b/docs/conf/inspircd.conf.example
@@ -724,10 +724,10 @@
           # the ircd. This may be set for security reasons or vanity reasons.
           customversion=""
 
-          # operspywhois: show opers (users/auspex) the +s channels a user is in. Values:
-          #  splitmsg  Split with an explanatory message
-          #  yes       Split with no explanatory message
-          #  no        Do not show
+          # operspywhois: show opers (users/auspex) the +s and +p channels a user is in. Values:
+          #  splitmsg  Split secret/private from public channels with an explanatory message
+          #  yes       Show secret/private channels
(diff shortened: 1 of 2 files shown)


docs/conf/inspircd.conf.example
```diff
@@ -725,8 +725,8 @@
           customversion=""
 
           # operspywhois: show opers (users/auspex) the +s channels a user is in. Values:
-          #  splitmsg  Split with an explanatory message
-          #  yes       Split with no explanatory message
+          #  splitmsg  Split secret and non-secret with an explanatory message
+          #  yes       Show secret channels
```
SaberUK: Private channels are also split.

docs/conf/modules.conf.example
```diff
@@ -402,14 +402,14 @@
...
-<channames
-	# denyrange: characters or range of characters to deny in channel
-	# names.
-	denyrange="2,3"
-
-	# allowrange: characters or range of characters to specifically allow
-	# in channel names.
-	allowrange="">
+#<channames
+#	# denyrange: characters or range of characters to deny in channel
```
SaberUK: There's no need to comment things out which are already commented out.

PR #1450, opened 2017-12-30
m_chanhistory: add setting to optionally exempt bots from receiving channel history

Techman-:


SISheogorath:
I think this can only go to master. It's a new feature and insp20 is under feature-freeze for a long time already.

Techman-:
@SISheogorath: It adds an optional feature that doesn't affect already-existant behavior, I think it's fine and I even made it forward-compatible with 3.0 at the request of @SaberUK

SaberUK:
@SISheogorath The insp20 feature freeze is not 100% strict. If a change is is extremely small and/or fixes a problem (like this one, see the commit I pushed for a description) then insp20 is fine.

PR #1451, opened 2018-01-02
Performance and code style improvements of OnDataReady

systocrat:
This patch improves performance and code style of OnDataReady by:
- Moving buffer allocation and sizing outside of a loop to properly reuse memory already created
- Re-using variables instead of reinitializing/recalculating them during the loop
- Const-ifying variables that do not change during the loop
- Replacing a goto with a much more appropriate flag inside the loop

### DIFF
--- a/src/users.cpp
+++ b/src/users.cpp
@@ -227,49 +227,73 @@ void UserIOHandler::OnDataReady()
 			user->nick.c_str(), (unsigned long)recvq.length(), user->MyClass->GetRecvqMax());
 		return;
 	}
+
 	unsigned long sendqmax = ULONG_MAX;
 	if (!user->HasPrivPermission("users/flood/increased-buffers"))
 		sendqmax = user->MyClass->GetSendqSoftMax();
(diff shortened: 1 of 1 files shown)


src/users.cpp
```diff
@@ -227,49 +227,72 @@ void UserIOHandler::OnDataReady()
...
-			if (line.length() < ServerInstance->Config->Limits.MaxLine - 2)
+
+			if (eol_found)
+				break;
+
+			if (line.length() < linemax_crlf)
 				line.push_back(c);
 		}
+
 		// if we got here, the recvq ran out before we found a newline
```
Adam-: This comment should be updated

systocrat: I've updated this comment to something more descriptive in my latest update, please check it out when you have the time and inclination.

SaberUK:
Merged in 3398ce7e50f1c269e8221df04e1eefb52d54c820 with some minor amendments. Thanks!

PR #1458, opened 2018-01-25
[2.0] Make sure banredirect metadata can not be duplicated

### LINKED ISSUE #1367: Banned from channel with no bans set in +b list
Unknown

When joining a channel with no bans set in +b(verified with /check output), the channel responds with a "you are banned" reply and redirects you to a redirect channel.

To join the channel.

**Additional information you deem important (e.g. issue happens only occasionally):**
Only happened to one channel. After deleting the channel from inspircd(by having everyone /part) and recreating, the issue disappeared.

2.0.23

linuxdaemon:
During a netburst, the banredirect `ModeWatcher::BeforeMode` is called with as if the ban was being set, resulting in the metadata being duplicated for the ban. This second metadata object will not be removed when the ban is unset, resulting in a "ghost ban" where there is no ban matching a user, but they are still redirected to the redirect channel.

### Steps to Reproduce
1. Have 2 servers, `SA` and `SB`.
2. Have 2 users, `UA` and `UB` on `SA` and `SB` respectively.
3. Both users join `#channel` and `UA` gets op.
4. `UA` joins `#otherchannel`.
5. `UA` sets `MODE #channel +b UB!*@*#otherchannel`.
6. `UB` parts `#channel` and attempts to rejoin, leading to them being forwarded to `#otherchannel`.
7. `SA` and `SB` split.
8. `SA` and `SB` reconnect.
9. `UB` is still banned from `#channel` and forwarded to `#otherchannel`.
10. `UA` sets `MODE #channel -b UB!*@*#otherchannel`.
11. `UB` is still unable to join `#channel` and is still forwarded to `#otherchannel`.

### Explanation
What appears to be happening is the `ModeWatcher::BeforeMode` method is being called as if the ban was being set during the netburst, resulting in the metadata being duplicated. The obvious fix to me was changing the `BanRedirectList` to a `set::set` to make sure only one copy of each ban appears in the list. Since the channel ban list can only have one matching entry anyways, this seemed to be the easiest fix.

This fixes #1367

### DIFF
--- a/src/modules/m_banredirect.cpp
+++ b/src/modules/m_banredirect.cpp
@@ -40,17 +40,9 @@ class BanRedirectEntry
 	: targetchan(target), banmask(mask)
 	{
 	}
-
-	bool operator<(const BanRedirectEntry& other) const
-	{
-		if (targetchan != other.targetchan)
-			return targetchan < other.targetchan;
-
-		return banmask < other.banmask;
-	}
 };
 
-typedef std::set<BanRedirectEntry> BanRedirectList;
+typedef std::vector<BanRedirectEntry> BanRedirectList;
 typedef std::deque<std::string> StringDeque;
 
 class BanRedirect : public ModeWatcher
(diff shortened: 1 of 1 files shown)


src/modules/m_banredirect.cpp
```diff
@@ -40,9 +40,17 @@ class BanRedirectEntry
 	: targetchan(target), banmask(mask)
 	{
 	}
+
+	bool operator<(const BanRedirectEntry& other) const
+	{
+		if (this->targetchan < other.targetchan)
```
Adam-: The correct way to implement this is:

```c++
if (targetchan != other.targetchan) return targetchan < other.targetchan;
return banmask < other.banmask;
```

src/modules/m_banredirect.cpp
```diff
@@ -259,7 +267,7 @@ class ModuleBanRedirect : public Module
 
 				for(BanRedirectList::iterator i = redirects->begin(); i != redirects->end(); i++)
 				{
-					modestack.Push('b', i->targetchan.insert(0, i->banmask));
+					modestack.Push('b', i->banmask + i->targetchan);
```
Adam-: Refrain from making superfluous changes in 2.0

linuxdaemon: This change is necessary as the `iterator` in `std::set` is `const_iterator` so `i->targetchan` can not be modified, and `std::string::insert` modifies the string.

linuxdaemon:
A duplicate check would generally be slower on insert, since it would have to iterate the whole vector, whereas inserting in to a set would only iterate part of the way, due to a set being sorted.

Adam-:
Of course. But it would be less change, and we try to be careful with 2.0 to not accidentally break it. Ban lists are generally very small (<250).

SaberUK:

PR #1460, opened 2018-02-12
[2.0]: Helpop: Move extban 'U' from acting to matching

genius3000:
Already fixed in master in 8587df2


SaberUK:

PR #1461, opened 2018-02-16
Add Haiku support

puckipedia:
Compiled on hrev51795, with the modern GCC. on gcc2h, run `setarch x86` first.
I also tried compiling `m_regex_posix`, `m_regex_stdlib`, and `m_ssl_openssl` because those have devel libs by default.

cc @SaberUK

### DIFF
--- a/make/template/main.mk
+++ b/make/template/main.mk
@@ -81,6 +81,11 @@ ifeq ($(SYSTEM), darwin)
   CORELDFLAGS = -dynamic -bind_at_load -L.
   PICLDFLAGS = -fPIC -shared -twolevel_namespace -undefined dynamic_lookup
 endif
+ifeq ($(SYSTEM), haiku)
+  LDLIBS = -lnetwork -lstdc++
+  CORELDFLAGS = -L.
+  PICLDFLAGS = -fPIC -shared
+endif
 
 ifndef INSPIRCD_DEBUG
   INSPIRCD_DEBUG=0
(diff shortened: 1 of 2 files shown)


make/template/main.mk
```diff
@@ -59,9 +59,17 @@ ifneq ($(SYSTEM), openbsd)
...
+ifneq ($(SYSTEM), haiku)
+  LDFLAGS += -rdynamic
+
 ifneq ($(SYSTEM), darwin)
   LDLIBS += -pthread
 endif
+endif
+
+ifeq ($(SYSTEM), haiku)
+  CORELDFLAGS += -lnetwork
```
SaberUK: This should probably be in LDLIBS rather than CORELDFLAGS.

make/template/main.mk
```diff
@@ -36,8 +36,8 @@ BUILDPATH ?= $(PWD)/build
 SOCKETENGINE = @SOCKETENGINE@
 CORECXXFLAGS = -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -pipe -Iinclude -Wall -Wextra -Wfatal-errors -Wno-unused-parameter -Wshadow
 LDLIBS = -lstdc++
-CORELDFLAGS = -rdynamic -L. $(LDFLAGS)
-PICLDFLAGS = -fPIC -shared -rdynamic $(LDFLAGS)
```
SaberUK: You probably want to just override both of these with Haiku specific variants like we do on Darwin (see below).

SaberUK:

PR #1469, opened 2018-03-25
Do not send unset topics on channel join per RFC

### LINKED ISSUE #1091: Cleared/unset topic misleading behavior

B00mX0r:
Fixes #1091. It looks like this was already fixed, and then somehow reverted in the move to modularity.

PRing against master because of the discussion in aforementioned issue.

### DIFF
--- a/src/coremods/core_channel/core_channel.cpp
+++ b/src/coremods/core_channel/core_channel.cpp
@@ -147,7 +147,7 @@ class CoreModChannel : public Module, public CheckExemption::EventListener
 			// Remove existing invite, if any
 			invapi.Remove(localuser, chan);
 
-			if (chan->topicset)
+			if (chan->topic.length())
 				Topic::ShowTopic(localuser, chan);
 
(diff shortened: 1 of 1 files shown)


SaberUK:
LGTM. Can you check if there are any other places that topicts is checked before I merge this though.

SaberUK:

PR #1471, opened 2018-03-30
[2.0] Update the GeoIP ExtItem when CGI:IRC sets the real IP

### LINKED ISSUE #1470: /check command in combination with m_geoip only able to fetch the geoip country code of a webirc, not the real conneciton behind it
**Description**
When using the /check command on a user connecting via a cgi:irc/webirc connection, then the geoip location is not fetched from the actual real host (despite /check finding it) but only from the cgi:irc host. 
This issue is confirmed to occur on InspIRCd 2.0.25

1. Be /oper'ed up on an InspIRCd network (tested on version 2.0.25)
2. Have a user with a properly configured webirc (e.g. KiwiIRC) connect
3. /check the nickname and look at the CHECK meta:geoip_cc - it will be the originiation of the webirc
IP/host NOT of the actual real IP/host of the connected entity.

This appears unintuitive as the /check command can fetch the real IP/host but the issue might be at which point m_geoip tries to fetch the location during the registration process.

Thank you for your time and effort in advance.

Regards,

Koragg

genius3000:
I've tested this for both normal user and webirc connections using netcat; limiting the update to after NICK/USER have been sent keeps the GeoIP check to 1 during a normal user and to just 2 during webirc (as intended). Untested with pass/ident cgi:irc methods but they run off `OnUserRegister()` which is called right after `NICK` or `USER` (whichever is sent second).

Fixes #1470

### DIFF
--- a/src/modules/extra/m_geoip.cpp
+++ b/src/modules/extra/m_geoip.cpp
@@ -58,7 +58,7 @@ class ModuleGeoIP : public Module
 				throw ModuleException("Unable to initialize geoip, are you missing GeoIP.dat?");
 
 		ServerInstance->Modules->AddService(ext);
-		Implementation eventlist[] = { I_OnSetConnectClass, I_OnStats };
+		Implementation eventlist[] = { I_OnSetConnectClass, I_OnSetUserIP, I_OnStats };
 		ServerInstance->Modules->Attach(eventlist, this, sizeof(eventlist)/sizeof(Implementation));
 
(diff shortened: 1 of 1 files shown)


KoraggKnightWolf:
Dear genius3000,

Thank you once more for a swift response and also for making the necessary modifications. I have tested them on my own network and can confirm that /check now gives out the geoip location of the real host/IP of webirc users as it should.

Regards,

Koragg

PR #1472, opened 2018-03-30
[2.0] Properly check E-Lines in general and specifically from CGI:IRC

### LINKED ISSUE #1465: E-lines don't work for WebIRC users
1. Set an E-line on an IPv6 range.
2. Set a G-line on the parent IPv6 range.
3. User cannot connect with WebIRC due to G-line.

User is G-lined when connecting with WebIRC.

User should be able to connect just fine because of the E-line.

InspIRCd-2.0.25 rv2.0.25-11-g758cf50

genius3000:
This PR accomplishes two things:
1. Fixes CGI:IRC checking E-Lines on a user while their host is still that of the CGI:IRC server. Which allows a false match to happen and remain.
2. Fixes the ELine::Matches() function to not return 'false' when the user is already marked exempt (previously matched an E-Line). Looking in the history, this was fixed once already and reverted as it caused issue #989.  
I believe that the root cause of this issue was the user's host still being the CGI:IRC server which was E-Lined, as it had to match an active E-Line in order to return 'true' again. This is fixed as commit 1 here.  
To the best of my searching, the `user->exempt` flag is set:
    * upon initial connection by checking for E-Lines matching the user (if a matched E-Line is returned, set true; else false)
    * whenever User::SetClientIP() is called (unless specified not to recheck e-lines) and this does the same true/false setting as above
    * within the full server userlist re-check of E-Lines; this marks each user's exempt flag false prior to checking and marks it true if a match was found. This will not be affected by this.

I've tested this with debug SNOTICEs showing user nick, IP, host, and exempt status throughout the 'OnCheckReady()' of m_cgiirc (using WEBIRC syntax). The findings of #989 were not seen and could only be seen if the first commit was reverted (user host being incorrect). I would still appreciate if @B00mX0r could confirm that this doesn't bring the issue back.  
The issue as noted in #1465 was tested, with positive results. The E-Line on the user's IP was correctly matched even with an E-Line originally matched on the CGI:IRC server.

Fixes #1465

### DIFF
--- a/src/modules/m_cgiirc.cpp
+++ b/src/modules/m_cgiirc.cpp
@@ -201,8 +201,8 @@ class ModuleCgiIRC : public Module
 	{
 		cmd.realhost.set(user, user->host);
 		cmd.realip.set(user, user->GetIPString());
-		ChangeIP(user, newip);
 		user->host = user->dhost = user->GetIPString();
+		ChangeIP(user, newip);
 		user->InvalidateCache();
 		RecheckClass(user);
 		// Don't create the resolver if the core couldn't put the user in a connect class or when dns is disabled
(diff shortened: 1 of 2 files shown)


SaberUK:
@Robby- can you test this please?

Robby-:
I've been testing this for a few days now, and all looks good, no issues.
Thank you @genius3000 and @SaberUK.

PR #1475, opened 2018-04-01
Improve settings validation in m_repeat

genius3000:
* Validate the settings in the same order as the parameter syntax
* Always validate `Lines` and `Secs`
    * previously you needed to set a `Diff` value for `Lines` and `Secs` to be validated
* Check `Backlog` for greater than _Max_
    * previously `Backlog` was only checked against being disabled


SaberUK:

PR #1476, opened 2018-04-02
Optimize multiple redundant function calls in core

B00mX0r:
This PR optimizes loops inside of InspIRCd's core to remove redundant function calls and unnecessary memory reallocation. For most of the files affected here, that means STL optimization in the for loops. For users.cpp, a few redundant calls to mathematical operations in the write function (very commonly used) were optimized.

### DIFF
--- a/src/coremods/core_channel/cmd_names.cpp
+++ b/src/coremods/core_channel/cmd_names.cpp
@@ -36,7 +36,7 @@ CmdResult CommandNames::HandleLocal(const std::vector<std::string>& parameters, 
 {
 	Channel* c;
 
-	if (!parameters.size())
+	if (parameters.empty())
 	{
 		user->WriteNumeric(RPL_ENDOFNAMES, '*', "End of /NAMES list.");
 		return CMD_SUCCESS;
--- a/src/inspsocket.cpp
+++ b/src/inspsocket.cpp
@@ -293,7 +293,7 @@ void StreamSocket::FlushSendQ(SendQueue& sq)
 					const SendQueue::Element& elem = *i;
 					iovecs[j].iov_base = const_cast<char*>(elem.data());
 					iovecs[j].iov_len = elem.length();
-					rv_max += elem.length();
+					rv_max += iovecs[j].iov_len;
 				}
 				rv = SocketEngine::WriteV(this, iovecs, bufcount);
 			}
--- a/src/users.cpp
+++ b/src/users.cpp
@@ -645,8 +645,8 @@ bool User::ChangeNick(const std::string& newnick, time_t newts)
 	}
 
 	if (this->registered == REG_ALL)
-		this->WriteCommon("NICK %s",newnick.c_str());
-	std::string oldnick = nick;
+		this->WriteCommon("NICK %s", newnick.c_str());
+	const std::string oldnick = nick;
 	nick = newnick;
 
 	InvalidateCache();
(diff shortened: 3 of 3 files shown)


rburchell:
I'm not sure I see the point of at least some parts of this change as-is.

Don't get me wrong: performance improvements are always a good thing, but without measurement, there might not be improvement at all (and in the majority of the cases I see here, there likely won't be, outside of debug builds[1]).

That leaves us with a change which makes things a little more convoluted to read for no run time performance change - and additionally, changes code that is (thus far) known to work everywhere to something new and untested -- a very minor risk admittedly, but still.

If you are genuinely interested in performance, then I'd suggest first measuring to see what might be interesting to look into further. Tools like perf (perf record / report, and top), valgrind's callgrind, heaptrack, etc, etc, are all worth looking into. The added advantage here is that once you have found something that is worth improving, you can also provide metrics on _how_ your change improves it precisely -- see [this commit message](https://github.com/qt/qtdeclarative/commit/1a3dd2031a8151bf9d50d070840fb664f7fd5d44) for what I'd consider to be a good example of that.

[1]: compilers can already optimize away calls to end() for instance, see https://godbolt.org/g/pLtWMw and note that the assembly is virtually identical at -O1 already..

---

disclaimer: I'm not really interested in really reviewing this, this is just a drive-by comment from a notification. my opinion also doesn't really matter in the grand scheme of things.

SISheogorath:
After checking a bit back on this changes, I think they may introduce more trouble than they help out with. So `end()` calls are already O(1) but they are still dynamic. Why is that important? Because we run module code. For example this PR "optimizes" the ban check loop, but that's actually not the best idea as `CheckBan()` calls module code. Module code is dynamic and indeterministic as we don't know the code. This could actually result in changes in the banlist itself (yes, stupid idea, but theoretically possible) and this way cause `end()`  to cause an different result than at the beginning of the iteration. This way this change can actually result into segfaults when there is module code that does bad, but valid things.

Then again, we have `it++` vs `++it`. Yes, `it++` is never faster than `++it` and while these changes makes sense on this level and with old compilers, as iterators are part of the CXX11 standards, they are already optimized by the compiler. So no need for these changes either.

There are a few changes remaining, which actually makes sense, but in general, when you do micro optimizations, be careful, done wrong they hurt way more than they help.

Just to mention: I don't want to offend anyone. There are just various problems I see and I wanted to point them out very clear.

In general :+1: nice to see people care about micro optimization :)

SaberUK:
@B00mX0r please can you revert the `.end()` changes (for the reason w00t mentioned above) and the changes from post increment to pre increment (there is no performance changes in this context from doing so, we can update the style of these as they are edited).

Thanks!

B00mX0r:
Hi all,

Thanks for the feedback. I have implemented the changes. For reference, as I do not have incredible familiarity with STL, I was working off of https://wiki.inspircd.org/STL_FAQ. My apologies, as this page was evidently outdated by improvements in gcc.

src/cull_list.cpp
```diff
@@ -33,10 +33,9 @@ void CullList::Apply()
 		working.swap(SQlist);
 		for(std::vector<LocalUser *>::iterator a = working.begin(); a != working.end(); a++)
 		{
-			LocalUser *u = *a;
```
SaberUK: This should be kept as a local variable for readability reasons.

PR #1480, opened 2018-04-07
Fix an invalid mode parameters numeric upon attempting to view listmodes

B00mX0r:
This fixes an issue where after https://github.com/inspircd/inspircd/commit/d5a6054948502625d7f0c235f6faaeea58734de5, attempting to view the list for a listmode will cause a "You must specify a parameter" error under https://github.com/inspircd/inspircd/commit/d5a6054948502625d7f0c235f6faaeea58734de5#diff-34903a46f53c58683aa794a80721ce8eR92.

### DIFF
--- a/include/listmode.h
+++ b/include/listmode.h
@@ -178,6 +178,15 @@ class CoreExport ListModeBase : public ModeHandler
 	 */
 	virtual bool ValidateParam(User* user, Channel* channel, std::string& parameter);
 
+	/** In the event that the mode should be given a parameter, and no parameter was provided, this method is called.
+	 * This allows you to give special information to the user, or handle this any way you like.
+	 * @param user The user issuing the mode change
+	 * @param dest For user mode changes, the target of the mode. For channel mode changes, NULL.
+	 * @param channel For channel mode changes, the target of the mode. For user mode changes, NULL.
+	 * See mode.h
+	 */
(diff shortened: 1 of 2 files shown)


src/listmode.cpp
```diff
@@ -221,6 +221,11 @@ bool ListModeBase::ValidateParam(User*, Channel*, std::string&)
 	return true;
 }
 
+void ListModeBase::OnParameterMissing(User* user, User* dest, Channel* channel)
+{
+	return;
```
SaberUK: There's no need for a return statement here.

src/listmode.cpp
```diff
@@ -221,6 +221,11 @@ bool ListModeBase::ValidateParam(User*, Channel*, std::string&)
 	return true;
 }
 
+void ListModeBase::OnParameterMissing(User* user, User* dest, Channel* channel)
```
SaberUK: You can omit the names for these parameters as they are not used.

include/listmode.h
```diff
@@ -178,6 +178,11 @@ class CoreExport ListModeBase : public ModeHandler
 	 */
 	virtual bool ValidateParam(User* user, Channel* channel, std::string& parameter);
 
+	/** Handle no mode parameter given
+	 * See mode.h
+	 */
```
SaberUK: Please copy the documentation from the base class.

B00mX0r:
@SaberUK Thanks for the feedback. Implemented the suggested changes.

PR #1482, opened 2018-04-09
Fix blank instead of empty PART message when a user is shunned

B00mX0r:
This fixes an issue where, if a shunned user parted a channel, their PART message would be truncated to nothing rather than deleted. This resulted in users who leave with a message leaving with "" as their message, whereas shunned users executing PART with no message would cause expected behavior.

### DIFF
--- a/src/modules/m_shun.cpp
+++ b/src/modules/m_shun.cpp
@@ -280,7 +280,7 @@ class ModuleShun : public Module
 		else if ((command == "PART") && (parameters.size() > 1))
 		{
 			/* same for PART */
-			parameters[1].clear();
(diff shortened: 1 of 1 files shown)


SaberUK:
Should this also be applied to QUIT?

B00mX0r:
@SaberUK No, as QUIT clears parameters rather than parameters[0]. The reason why PART cleared parameters[1] is because, unlike QUIT, the first parameter is the channel being parted, and that parameter must remain.

SaberUK:
Is there any reason you didn't use `.clear()` like for QUIT?

PR #1483, opened 2018-04-09
Improve some documentation

### LINKED ISSUE #1481: TLINE helpop docs do not specify that it matches against a user@host mask.
The Helpop information for TLINE is out of date as it can be performed using user@host as well as just host/ip address.  Perhaps someone could do it as that would be a nice thing to know.

genius3000:
* Improve exemptchanops in inspircd.conf.example (`<options>`) and modules.conf.example; along with both helpop examples.
* Add "the" to all "requires X module" in both helpop examples.
* Add SNOMASK 'r' and 'R' (OPERLOG) to both helpop examples.
* Clean up unnecessarily split SNOMASK lines in both helpop examples.
* Move chmode 'g' to be in alphabetical order in both helpops examples.
* Specify the needed mask formats for TLINE in helpop-full example.
* Remove extra `\` in RLINE mask format in helpop-full example.

Resolves #1481

### DIFF
--- a/docs/conf/helpop-full.conf.example
+++ b/docs/conf/helpop-full.conf.example
@@ -2,6 +2,8 @@
 #  Helpop Standard  #
 #####################
 
+<config format="xml">
+
 <alias text="HELP" replace="HELPOP $2-">
 
 <helpop key="start" value="InspIRCd Help System
@@ -406,11 +408,12 @@ LOCKSERV       UNLOCKSERV">
 
 Returns the ip and nickname of the given users.">
 
-<helpop key="tline" value="/TLINE <host/IP mask>
+<helpop key="tline" value="/TLINE <mask>
 
 This command returns the number of local and global clients matched,
 and the percentage of clients matched, plus how they were matched
-(by IP address or by hostname).">
+(by IP address or by hostname). Mask should be given as either a
+nick!user@host or user@IP (wildcards acceptable).">
 
 <helpop key="lockserv" value="/LOCKSERV
 
@@ -471,7 +474,7 @@ n    Block private and channel notices
 P    Block part messages
 q    Block quit messages
 o    Don't match against opers
-c    Strip all color codes from the message before matching
+c    Strip all formatting codes from the message before matching
 *    Represents all of the above flags
 -    Does nothing, a non-op for when you do not want to specify any
      flags
@@ -752,7 +755,7 @@ Sends a message to all +w users.">
 
 <helpop key="rline" value="/RLINE <regex> [<duration> :<reason>]
 
-Sets or removes an r-line (regex line) on a n!u@h\\sgecos mask. You
+Sets or removes an r-line (regex line) on a n!u@h\sgecos mask. You
 must specify all three parameters to add an rline, and one parameter
 to remove an rline (just the regex).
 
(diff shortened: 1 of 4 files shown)


docs/conf/helpop-full.conf.example
```diff
@@ -927,11 +930,39 @@ using their cloak when they quit.">
...
- g <mask>           Blocks messages matching the given glob mask
-                    (requires chanfilter module).
- X <mode>           Makes channel operators immune to the specified
-                    restrictive mode (requires exemptchanops module).
+ X <mode>:<status>  Makes users of <status> or higher immune to the
+                    specified restrictive mode. For example: flood:h
+                    (requires exemptchanops module).
+Possible restrictive modes for +X are:
+
+auditorium-see (requires the auditorium module)
```
SaberUK: The description should probably be after the module requirement.

SaberUK: Also, the helpop file mostly uses uses "requires X module" whereas this uses "requires the X module". We should probably just decide on one. Probably the latter? it seems more grammatically correct to me.

genius3000: Did you mean "before" the module requirement, ie. like the others currently do? Which I now see would make more sense.  
I do agree, "requires the x module" sounds more correct as it's a specific module being referred to. I can go through and change the rest.

docs/conf/helpop-full.conf.example
```diff
@@ -927,11 +930,39 @@ using their cloak when they quit.">
                     channel (requires stripcolor module).
  T                  Blocks /NOTICEs to the channel from users who are
                     not at least halfop (requires nonotice module).
-
- g <mask>           Blocks messages matching the given glob mask
-                    (requires chanfilter module).
- X <mode>           Makes channel operators immune to the specified
-                    restrictive mode (requires exemptchanops module).
+ X <mode>:<status>  Makes users of <status> or higher immune to the
```
SaberUK: Mode means something different in IRC parlance. Can we pick a better word here?

docs/conf/helpop-full.conf.example
```diff
@@ -927,11 +930,39 @@ using their cloak when they quit.">
...
+                    specified restrictive mode. For example: flood:h
+                    (requires exemptchanops module).
+Possible restrictive modes for +X are:
```
SaberUK: This should have a space before it and should probably read

>Valid exemptions ???s for +X are:

(See the comment above for what to put instead of ???.)

docs/conf/helpop-full.conf.example
SaberUK: Ideally it should be something like this:

```
wibble      Does something really important that is described here
                 (requires the wibbling module)
```

docs/conf/helpop-full.conf.example
```diff
@@ -927,11 +930,39 @@ using their cloak when they quit.">
...
+
+auditorium-see (requires the auditorium module)
+               Permission required to see the full user list of a +u channel
+auditorium-vis (requires the auditorium module)
+               Permission required to be visible in a +u channel
+blockcaps      (requires the blockcaps module)
+               Channel mode +B - Blocks messages with too many capital letters
```
SaberUK: There's no need for a capital letter after a dash.

SaberUK: Also, these descriptions describe what the mode does but the behaviour of the extban is generally the opposite of the mode (i.e. exempting matching users rather than punishing matching users).

genius3000: Maybe it's even better to just list the channel mode character they are, as the descriptions were just listed prior and we already say it's an exemption.

docs/conf/helpop-full.conf.example
SaberUK: That would work, but then you'd have to realign all of the columns. How about "type"?

genius3000: Yeah, "type" would work; then I can specify them as "restriction types" in the description to be clear.  
Though they way it's done in master for m_repeat (with an off-set into the description text) doesn't seem terrible, in a compromise between space and ease.

docs/conf/helpop-full.conf.example
```diff
@@ -752,7 +753,7 @@ Sends a message to all +w users.">
 
 <helpop key="rline" value="/RLINE <regex> [<duration> :<reason>]
 
-Sets or removes an r-line (regex line) on a n!u@h\\sgecos mask. You
+Sets or removes an r-line (regex line) on a n!u@h\sgecos mask. You
```
genius3000: I have tested that this shows correctly when called via `/helpop rline` (in Quassel). So unless other clients parse it incorrectly, I assume it was a typo and can confuse people as to whether a `\s` or a `\\s` is required in the mask.

docs/conf/helpop-full.conf.example
```diff
@@ -992,30 +1011,25 @@ Note that all /STATS use is broadcast to online IRC operators.">
...
+ O      Allows receipt of remote oper-up, oper-down, and oper-failure messages.
  q      Allows receipt of local quit messages.
  Q      Allows receipt of remote quit messages.
+ r      Allows receipt of local oper commands (requires the operlog module).
+ R      Allows receipt of remote oper commands (requires the operlog module).
  t      Allows receipt of attempts to use /STATS (local and remote).
- v      Allows receipt of oper-override notices (requires override module).
+ v      Allows receipt of oper-override notices (requires the override module).
  x      Allows receipt of local Xline notices (g/Z/q/k/e/R/shuns).
  X      Allows receipt of remote Xline notices (g/Z/q/k/e/R/shuns).">
```
genius3000: Keeping these all single line matches up with the `STATS` help above, with the `l` line being a few characters shorter than `J` here.

docs/conf/inspircd.conf.example
```diff
@@ -613,8 +613,38 @@
...
+         #  - nonotice        Channel mode +T - blocks /NOTICEs to the channel
+         #                    (requires the nonotice module).
+         #  - regmoderated    Channel mode +M - blocks unregistered users from
+         #                    speaking (requires the services account module).
+         #  - stripcolor      Channel mode +S - strips mIRC color codes from
+         #                    messages (requires the stripcolor module).
+         #  - topiclock       Channel mode +t - limits changing the topic to (half)ops
+         # You can also configure this on a per-channel basis with a channel mode.
+         # See m_exemptchanops in modules.conf.example for more details.
+         exemptchanops="censor:o filter:o nickflood:o nonick:v regmoderated:o"
```
genius3000: Kept the addition of the restriction (mode) descriptions here to keep it easy when configuring.

docs/conf/helpop-full.conf.example
SaberUK: This is an escape sequence when using `<config format="compat">`. We should probably force XML mode in these example files.

genius3000: Ah compat format! Well the examples do ship as `xml` with `prefixquit="""` and a note about it. Could add a comment note before the rline key.

docs/conf/helpop-full.conf.example
```diff
@@ -833,32 +834,34 @@ using their cloak when they quit.">
                     be disabled).
  o <nickname>       Gives op status to <nickname>.
  a <nickname>       Gives protected status to <nickname>, preventing
-                    them from them from being kicked (+q only,
-                    requires chanprotect module).
+                    them from being kicked (+q only, requires the
+                    chanprotect module).
```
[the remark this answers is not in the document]
genius3000: Definitely!

docs/conf/helpop-full.conf.example
```diff
@@ -750,9 +751,11 @@ can be negated by any eline that matches.">
 
 Sends a message to all +w users.">
 
+# NOTE: Use '\\s' instead of '\s' (in the mask example below)
```
SaberUK: Can't we just force `<config format="xml">` for this file and then let people who want to change that fix it themself?

docs/conf/helpop-full.conf.example
```diff
@@ -1040,42 +1056,42 @@ setting +I <extban>.
...
+               matching users (requires the blockcaps module).
+ C:<banmask>   Blocks CTCPs from matching users (requires the noctcp
                module).
  N:<banmask>   Blocks nick changes from matching users (requires
-               nonicks module).
- Q:<banmask>   Blocks kicks by matching users (requires nokicks
+               the nonicks module).
+ Q:<banmask>   Blocks kicks by matching users (requires the nokicks
                module).
  S:<banmask>   Strips color/bold/underline from messages from matching
```
SaberUK: This should probably say.

>Strips formatting codes from messages ...

docs/conf/helpop-full.conf.example
```diff
@@ -1040,42 +1056,42 @@ setting +I <extban>.
...
  R:<account>   Matches users logged into a matching account (requires
-               services account module).
+               the services account module).
  U:<banmask>   Matches unregistered users matching the given banmask.
-               (requires services account module).
+               (requires the services account module).
 
 Acting extbans:
 
  c:<banmask>   Blocks any messages that contain color codes from
```
SaberUK: This should probably say.

>Blocks any messages that contain formatting codes from

docs/conf/helpop-full.conf.example
genius3000: Sure, I'll check/update the same elsewhere for the mode too.

docs/conf/helpop-full.conf.example
genius3000: Ditto

docs/conf/helpop-full.conf.example
genius3000: I didn't think that was per file. If it is then sure that would be acceptable.

PR #1485, opened 2018-04-09
Reduce some unnecessary or confusing behavior

B00mX0r:
This PR attempts to streamline behavior through adding more vector reservations, changing some confusing loops, and removing a few unnecessary function calls.

These are all minor changes to parts of the code that I found confusing to read through and understand.

### DIFF
--- a/src/coremods/core_channel/core_channel.cpp
+++ b/src/coremods/core_channel/core_channel.cpp
@@ -132,6 +132,7 @@ class CoreModChannel : public Module, public CheckExemption::EventListener
 		// Build a map of limits to their mode character.
 		insp::flat_map<int, std::string> limits;
 		const ModeParser::ListModeList& listmodes = ServerInstance->Modes->GetListModes();
+		limits.reserve(listmodes.size());
 		for (ModeParser::ListModeList::const_iterator iter = listmodes.begin(); iter != listmodes.end(); ++iter)
 		{
 			const unsigned int limit = (*iter)->GetLowerLimit();
--- a/src/hashcomp.cpp
+++ b/src/hashcomp.cpp
@@ -125,7 +125,7 @@ bool irc::equals(const std::string& s1, const std::string& s2)
 	for (; *n1 && *n2; n1++, n2++)
 		if (national_case_insensitive_map[*n1] != national_case_insensitive_map[*n2])
 			return false;
-	return (national_case_insensitive_map[*n1] == national_case_insensitive_map[*n2]);
+	return true;
 }
 
 bool irc::insensitive_swo::operator()(const std::string& a, const std::string& b) const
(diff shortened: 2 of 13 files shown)


src/hashcomp.cpp
```diff
@@ -255,9 +255,19 @@ bool irc::tokenstream::GetToken(long &token)
...
+
 irc::sepstream::sepstream(const std::string& source, char separator, bool allowempty)
-	: tokens(source), sep(separator), pos(0), allow_empty(allowempty)
+	: tokens(source)
+	, sep(separator)
+	, pos(0)
+	, allow_empty(allowempty)
 {
+	this->tokenlen = this->tokens.length();
+	this->numparts = std::count(this->tokens.begin(), this->tokens.end(), this->sep);
```
SaberUK: This is not valid because a separator may exist one or more times when allow_empty is false.

e.g. "foo  bar          baz" is a space delimited string with three tokens; "foo", "bar", and "baz".

src/hashcomp.cpp
```diff
@@ -125,7 +125,7 @@ bool irc::equals(const std::string& s1, const std::string& s2)
 	for (; *n1 && *n2; n1++, n2++)
 		if (national_case_insensitive_map[*n1] != national_case_insensitive_map[*n2])
 			return false;
-	return (national_case_insensitive_map[*n1] == national_case_insensitive_map[*n2]);
+	return true;
```
SaberUK: I think this breaks the comparison of strings of an inequal length?

src/hashcomp.cpp
```diff
@@ -273,15 +283,15 @@ bool irc::sepstream::GetToken(std::string &token)
...
-			this->pos = this->tokens.length() + 1;
+			this->pos = this->tokenlen + 1;
 			token.clear();
 			return false;
 		}
 	}
 
 	size_t p = this->tokens.find(this->sep, this->pos);
 	if (p == std::string::npos)
-		p = this->tokens.length();
```
SaberUK: std::string::length() is usually (the standard recommends it and its done by all of the compilers we support) O(1) so caching a string length is unnecessary.

src/coremods/core_channel/core_channel.cpp
```diff
@@ -132,6 +132,7 @@ class CoreModChannel : public Module, public CheckExemption::EventListener
 		// Build a map of limits to their mode character.
 		insp::flat_map<int, std::string> limits;
 		const ModeParser::ListModeList& listmodes = ServerInstance->Modes->GetListModes();
+		limits.reserve(listmodes.size());
```
SaberUK: This is probably going to reserve way too much space as its indexed by limit and most people will have half a dozen limits or so.

This is also a reasonably cold codepath so a extra few allocations won't hurt here.

SaberUK:
As you have not responded to any of my messages in three-ish months i've manually committed the part of this which I think is okay (5585654df265bc37d547fa7738e35cc7ae7dacbb). Thanks for contributing! 👍

PR #1488, opened 2018-04-11
Use "real name" to refer to the real name field.

genius3000:
Currently, "gecos" is used in the `ServerInfo` class for "description" and in the `User` class for "real name", named (in `User`) "fullname".  
This PR is:
* Changing the ServerInfo "gecos" field to "description", only a few small changes here.
* Changing documentation and server sent messages that use "gecos" to mean "real name" to use a more clear form of "real name", "realname", "name" (whichever fits best with the current adjacent form).

The module m_gecosban example in modules.conf.example remains with "Gecos" as the primary name, as it fits the alphabetical order of the file and needs to remain the module name (2.0 compat).

Edit: more changes regarding "realname" to come at a later time.
~~With staying clear of compat issues for now, I would still like to accomplish in this PR:~~
~~- [ ] Changing the `User` "fullname" to "realname"~~
~~- [ ] Changing the `OnChangeLocalUserGECOS` to something...~~
    ~~`OnChangeLocalUserName` matches the current `OnChangeName` but does look misleading with the `UserName`.~~

### DIFF
--- a/docs/conf/helpop-full.conf.example
+++ b/docs/conf/helpop-full.conf.example
@@ -145,10 +145,10 @@ everything from people with a host matching *.foo.net, you would do
 
 Sends a notice to a channel indicating you wish to join.">
 
-<helpop key="user" value="/USER <ident> <local host> <remote host> :<GECOS>
+<helpop key="user" value="/USER <ident> <local host> <remote host> :<real name>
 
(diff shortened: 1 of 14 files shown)


SaberUK:
LGTM. Say when you're ready for it to be merged.

PR #1497, opened 2018-05-28
Add --nopid command line option

chrisnovakovic:
I'm trying to eliminate as many unneeded PID files as possible from my system (the init scripts don't use them to identify running services), but there's currently no way to tell InspIRCd not to generate one. This PR implements a `--nopid` command line option, which prevents any PID file from being written to the file system, whether at the default location or the one specified in `<pid file>`.

Although this PR is against the master branch, it's straightforward to backport it to 2.0. I've already done all the work, and can submit a separate PR against the `insp20` branch for that if you like.

### DIFF
--- a/include/configreader.h
+++ b/include/configreader.h
@@ -146,6 +146,18 @@ struct CommandLineConf
 	 */
 	bool writelog;
 
+	/** If this is true, a PID file will be written
+	 * to the file given in the "file" variable of
+	 * the <pid> tag in the config file. This is
+	 * the default.
+	 * Passing --nopid as a command line argument
+	 * sets this to false; in this case, a PID file
+	 * will not be written, even the default PID
+	 * file that is usually written when the <pid>
+	 * tag is not defined in the config file.
+	 */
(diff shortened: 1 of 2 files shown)


SaberUK:
I'm not in objection to this but did you have any reason to make it a command line flag rather than just making it a config (or just specifying a pidfile of /dev/null)?

chrisnovakovic:
Admittedly I had a specific use case in mind, but I can also see it helping others (e.g. the Debian maintainers who were [trying to figure out how to make InspIRCd play nicely with systemd](https://alioth-lists-archive.debian.net/pipermail/pkg-irc-maintainers/2014-July/002069.html), before InspIRCd began shipping with a systemd unit file): in my InspIRCd init script, I want to tell the spawned InspIRCd process never to generate a PID file, regardless of what the configuration file might or might not say. The easiest and most reliable way to do that was to add a command line option that I could pass in the call to `inspircd`.

SaberUK:

PR #1504, opened 2018-07-17
Implemented module to check connections with DNSBL

iwalkalone69:
This is a module to use DNSBL lists but getting configuration from SQL database to allow to change configuration from a backend website.

**MySQL Create table statement**:
CREATE TABLE `dnsbl` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `name` varchar(20) NOT NULL,
  `domain` varchar(100) NOT NULL,
  `regex` varchar(250) NOT NULL,
  `priority` int(11) NOT NULL,
  `web` varchar(250) NOT NULL,
  `active` int(1) NOT NULL DEFAULT '1',
  PRIMARY KEY (`id`),
  UNIQUE KEY `name` (`name`),
  UNIQUE KEY `domain` (`domain`)
) ENGINE=InnoDB;

**Fields explanation**:
- id: internal ID.
- name: name of DNSBL list.
- domain: domain to make DNS queries (for example dnsbl.dronebl.org)
- regex: pcre regular expression to filter some positive queries (for example, ^127.0.0.4$, only accept 127.0.0.4 for answer).
- priority: this is used to order from high to low
- web: website to put in G-Line message to tell the user where he can request delete.
- active: enables/disables list.

**Configuration**
<module name="sqldnsbl" dbid="1" dbtable="dnsbl">
- dbid: id of database connection
- dbtable: it allows to change database table name

**Things to consider**
- It requires DNS and SQL modules.
- Every 100 checks it updates the list automatically from database.

**Things pending or to be improved**
- Allow to change how many checks are needed to update list from database.
- Check that regex and sql are available, if not raise an Exception.

This is the first time I try to contribute to this project and I wonder if this module could be useful for the project and if I can improve or I missed something. Meanwhile, I will try to the pending things.

Thanks.

### DIFF
--- a/src/modules/m_sqldnsbl.cpp
+++ b/src/modules/m_sqldnsbl.cpp
@@ -88,6 +88,7 @@ private:
 	LocalIntExt countExt;
     std::vector<SQLDNSBLList> List;
     int scans = 0;
+    int UpdateAfter = 1;
 protected:
     void UpdateLists();
     std::string GetReverseIP(LocalUser* user)
@@ -139,7 +140,7 @@ public:
     {
         scans++;
 
-        if (scans % 100 == 99) {
+        if (scans % UpdateAfter == UpdateAfter - 1) {
             UpdateLists();
         }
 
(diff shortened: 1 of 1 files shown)


SaberUK:
This seems like a niche module so it should probably go into inspircd-extras for 2.0.

SaberUK:
As previously mentioned this should go into inspircd-extras.

PR #1509, opened 2018-08-12
Changing OnUserQuit to allow quit message modify.

iwalkalone69:
The purpose of this pull request is to allow modifying quit message like we can do with part message in OnUserPart event.

The change is just removing const from parameter to allow its modification in modules.


SaberUK:
It would be better if an `OnUserPreQuit` event was to be introduced which is called with FIRST_MOD_RESULT. That way modules which just want to see the quit don't have to worry if a module after them in the chain has changed the message.

iwalkalone69:
Thanks.

I will add this event and revert the change on the existing one according to you comment.

iwalkalone69:
I've reverted changes on **OnUserQuit** event and added **OnUserPreQuit** event.

Hope it is done right. If not, tell me how should be done. Thanks.

include/modules.h
```diff
@@ -304,6 +304,16 @@ class CoreExport Module : public classbase, public usecountbase
...
 
+	/** Called when before a user quits.
+	 * The details of the exiting user are available to you in the parameter User *user
+	 * This event is only called when the user is fully registered when they quit. To catch
+	 * raw disconnections, use the OnUserDisconnect method.
+	 * @param user The user who is quitting
+	 * @param message The user's quit message (as seen by non-opers)
+	 * @param oper_message The user's quit message (as seen by opers)
+	 */
+	virtual ModResult OnUserPreQuit(User* user, std::string &message);
```
SaberUK: This should take a LocalUser not a User.

Also, the & should be against the type not against the variable name.

src/usermanager.cpp
```diff
@@ -174,16 +174,23 @@ void UserManager::QuitUser(User* user, const std::string& quitreason, const std:
...
-
-	ServerInstance->Logs->Log("USERS", LOG_DEBUG, "QuitUser: %s=%s '%s'", user->uuid.c_str(), user->nick.c_str(), quitreason.c_str());
-	user->Write("ERROR :Closing link: (%s@%s) [%s]", user->ident.c_str(), user->GetRealHost().c_str(), operreason ? operreason->c_str() : quitreason.c_str());
-
 	std::string reason;
 	reason.assign(quitreason, 0, ServerInstance->Config->Limits.MaxQuit);
+
+	ModResult MOD_RESULT;
+	FIRST_MOD_RESULT(OnUserPreQuit, MOD_RESULT, (user, reason));
+	if (MOD_RESULT == MOD_RES_DENY) {
```
SaberUK: Thereis no need to have {} on this.

iwalkalone69:
Applied change suggested in the code commented.

iwalkalone69:
Is everything ok now? Thanks.

SaberUK:
Hi, i'll try to get around to merging this before the last alpha.

iwalkalone69:
Rebased in insp3 branch (PR #1629).

PR #1510, opened 2018-08-12
Allows changing nick specifying password

iwalkalone69:
This allows to change nick specifying password (ircu style).

NICK nick:password

or

NICK nick!password

### DIFF
--- a/src/coremods/core_user/cmd_nick.cpp
+++ b/src/coremods/core_user/cmd_nick.cpp
@@ -27,7 +27,7 @@ CommandNick::CommandNick(Module* parent)
 	: SplitCommand(parent, "NICK", 1, 1)
 {
 	works_before_reg = true;
-	syntax = "<newnick>";
+	syntax = "<newnick[:password]>";
 	Penalty = 0;
 }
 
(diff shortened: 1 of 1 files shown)


SaberUK:
This should be in its own module rather than in a core module and should probably be filed against inspircd-extras for 2.0.

SISheogorath:
And to be honest, why `NICK <nickname>!<password>|<nickname>:<password>` this really makes no sense. Even channels split up passwords and you can't have more than one nick.

So `NICK <nickname> [:]<password>` would be less weird and work with the existing parsers.

iwalkalone69:
Thanks.

I will change it according your comments.

iwalkalone69:
@SaberUK, after reviewing, I see that to deliver it to an external module is not possible because in line **58** of **src/coremods/core_user/cmd_nick.cpp** nick syntax is checked:

`else if (!ServerInstance->IsNick(newnick))`

And the OnUserPreNick is after that, in line 65:

`FIRST_MOD_RESULT(OnUserPreNick, MOD_RESULT, (user, newnick));`

The solution I can see consist in putting OnUserPreNick before nick check and move the code introduced before to an external module.

Am I right?

Thanks.

genius3000:
You can use OnPreCommand to catch it (and strip the password) before the NICK command handler runs.

iwalkalone69:
Thanks @genius3000! It worked for me.

I remove the PR as no changes are needed.

PR #1522, opened 2018-10-25
Use C++11 random if available.

remyabel:
This takes advantage of C++11 features to use the better random facilities. `std::random_device` typically uses `/dev/urandom` on Linux systems to seed `std::default_random_engine` which is typically an LCG. Additional improvements can be added, but I wanted this patch to have as small of a footprint as possible.


SaberUK:
Hello, thanks for your contribution!

In a weeks after the 3.0 branch has forked i'll be pushing a giant commit series to move the master branch to require C++14 and i'll merge this after that happens.

As C++14 will be required anyway you can remove the test and make it fall back to `<random>` instead of rand_s/random on platforms that don't support arc4random_buf.

PR #1529, opened 2018-11-20
m_filter: Add a warn action for notifying opers while still allowing the message through.

Robby-:
m_filter: Add a warn action for notifying opers while still allowing the message through.

Thanks to @H7-25 (Simos) for the idea.

### DIFF
--- a/src/modules/m_filter.cpp
+++ b/src/modules/m_filter.cpp
@@ -39,6 +39,7 @@ enum FilterAction
 {
 	FA_GLINE,
 	FA_ZLINE,
+	FA_WARN,
 	FA_BLOCK,
 	FA_SILENT,
 	FA_KILL,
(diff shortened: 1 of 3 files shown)


src/modules/m_filter.cpp
```diff
@@ -363,6 +364,12 @@ ModResult ModuleFilter::OnUserPreMessage(User* user, const MessageTarget& msgtar
 
 			target = t->name;
 		}
+		if (f->action == FA_WARN)
+		{
+			ServerInstance->SNO->WriteGlobalSno('f', InspIRCd::Format("WARNING: %s their message to %s matched %s (%s)",
```
SaberUK:
```suggestion
			ServerInstance->SNO->WriteGlobalSno('f', InspIRCd::Format("WARNING: %s's message to %s matched %s (%s)",
```

PR #1535, opened 2018-12-14
Make more modules rehash atomically

linuxdaemon:
Atomic rehashing allows us to ensure that an invalid configuration doesn't get partially applied before erroring or remove existing settings and not add them back, thus causing unexpected behavior.

Mostly minor changes to only apply changes to a module's config if it does not error.

### Major changes
- Refactor of `core_info` including changes to the `CommandDie`/`CommandRestart` structure and functionality

- `m_vhost` and `m_customtitle` got new config objects to allow easier loading of their configs instead of reading them on command execution

- Restructured loading and configuration of `m_pbkdf2` providers to make the rehash atomic, added config data objects

- And various `typedef`s for type reuse throughout

### DIFF
--- a/src/coremods/core_channel/core_channel.cpp
+++ b/src/coremods/core_channel/core_channel.cpp
@@ -148,16 +148,8 @@ class CoreModChannel : public Module, public CheckExemption::EventListener
 	void ReadConfig(ConfigStatus& status) CXX11_OVERRIDE
 	{
 		ConfigTag* optionstag = ServerInstance->Config->ConfValue("options");
-		Implementation events[] = { I_OnCheckKey, I_OnCheckLimit, I_OnCheckChannelBan };
-		if (optionstag->getBool("invitebypassmodes", true))
-			ServerInstance->Modules.Attach(events, this, sizeof(events)/sizeof(Implementation));
-		else
-		{
-			for (unsigned int i = 0; i < sizeof(events)/sizeof(Implementation); i++)
-				ServerInstance->Modules.Detach(events[i], this);
-		}
-
-		joinhook.modefromuser = optionstag->getBool("cyclehostsfromuser");
+		bool bypassmodes = optionstag->getBool("invitebypassmodes", true);
+		bool modefromuser = optionstag->getBool("cyclehostsfromuser");
 
 		std::string current;
 		irc::spacesepstream defaultstream(optionstag->getString("exemptchanops"));
@@ -174,26 +166,43 @@ class CoreModChannel : public Module, public CheckExemption::EventListener
 			ServerInstance->Logs->Log(MODNAME, LOG_DEBUG, "Exempting prefix %c from %s", prefix, restriction.c_str());
 			exempts[restriction] = prefix;
 		}
-		exemptions.swap(exempts);
 
 		ConfigTag* securitytag = ServerInstance->Config->ConfValue("security");
 		const std::string announceinvites = securitytag->getString("announceinvites", "dynamic");
+		Invite::AnnounceState newstate;
 		if (stdalgo::string::equalsci(announceinvites, "none"))
-			cmdinvite.announceinvites = Invite::ANNOUNCE_NONE;
+			newstate = Invite::ANNOUNCE_NONE;
 		else if (stdalgo::string::equalsci(announceinvites, "all"))
-			cmdinvite.announceinvites = Invite::ANNOUNCE_ALL;
+			newstate = Invite::ANNOUNCE_ALL;
 		else if (stdalgo::string::equalsci(announceinvites, "ops"))
-			cmdinvite.announceinvites = Invite::ANNOUNCE_OPS;
+			newstate = Invite::ANNOUNCE_OPS;
 		else if (stdalgo::string::equalsci(announceinvites, "dynamic"))
-			cmdinvite.announceinvites = Invite::ANNOUNCE_DYNAMIC;
+			newstate = Invite::ANNOUNCE_DYNAMIC;
 		else
 			throw ModuleException(announceinvites + " is an invalid <security:announceinvites> value, at " + securitytag->getTagLocation());
 
 		// In 2.0 we allowed limits of 0 to be set. This is non-standard behaviour
 		// and will be removed in the next major release.
-		limitmode.minlimit = optionstag->getBool("allowzerolimit", true) ? 0 : 1;
+		size_t minlimit = optionstag->getBool("allowzerolimit", true) ? 0 : 1;
 
+		// Config is valid, apply it
+
+		// Validates and applies <banlist> tags, so do it first
 		banmode.DoRehash();
+
+		exemptions.swap(exempts);
+		limitmode.minlimit = minlimit;
+		cmdinvite.announceinvites = newstate;
+		joinhook.modefromuser = modefromuser;
+
+		Implementation events[] = { I_OnCheckKey, I_OnCheckLimit, I_OnCheckChannelBan };
+		if (bypassmodes)
+			ServerInstance->Modules.Attach(events, this, sizeof(events)/sizeof(Implementation));
+		else
+		{
+			for (unsigned int i = 0; i < sizeof(events)/sizeof(Implementation); i++)
+				ServerInstance->Modules.Detach(events[i], this);
+		}
 	}
 
 	void On005Numeric(std::map<std::string, std::string>& tokens) CXX11_OVERRIDE
--- a/src/coremods/core_channel/core_channel.cpp
+++ b/src/coremods/core_channel/core_channel.cpp
@@ -167,15 +167,15 @@ class CoreModChannel : public Module, public CheckExemption::EventListener
 
 		ConfigTag* securitytag = ServerInstance->Config->ConfValue("security");
 		const std::string announceinvites = securitytag->getString("announceinvites", "dynamic");
-		Invite::AnnounceState newstate;
+		Invite::AnnounceState newannouncestate;
 		if (stdalgo::string::equalsci(announceinvites, "none"))
-			newstate = Invite::ANNOUNCE_NONE;
+			newannouncestate = Invite::ANNOUNCE_NONE;
 		else if (stdalgo::string::equalsci(announceinvites, "all"))
-			newstate = Invite::ANNOUNCE_ALL;
+			newannouncestate = Invite::ANNOUNCE_ALL;
 		else if (stdalgo::string::equalsci(announceinvites, "ops"))
-			newstate = Invite::ANNOUNCE_OPS;
+			newannouncestate = Invite::ANNOUNCE_OPS;
 		else if (stdalgo::string::equalsci(announceinvites, "dynamic"))
-			newstate = Invite::ANNOUNCE_DYNAMIC;
+			newannouncestate = Invite::ANNOUNCE_DYNAMIC;
 		else
 			throw ModuleException(announceinvites + " is an invalid <security:announceinvites> value, at " + securitytag->getTagLocation());
 
@@ -188,7 +188,7 @@ class CoreModChannel : public Module, public CheckExemption::EventListener
 		// In 2.0 we allowed limits of 0 to be set. This is non-standard behaviour
 		// and will be removed in the next major release.
 		limitmode.minlimit = optionstag->getBool("allowzerolimit", true) ? 0 : 1;;
-		cmdinvite.announceinvites = newstate;
+		cmdinvite.announceinvites = newannouncestate;
 		joinhook.modefromuser = optionstag->getBool("cyclehostsfromuser");
 
 		Implementation events[] = { I_OnCheckKey, I_OnCheckLimit, I_OnCheckChannelBan };
--- a/src/coremods/core_channel/core_channel.cpp
+++ b/src/coremods/core_channel/core_channel.cpp
@@ -148,8 +148,6 @@ class CoreModChannel : public Module, public CheckExemption::EventListener
 	void ReadConfig(ConfigStatus& status) CXX11_OVERRIDE
 	{
 		ConfigTag* optionstag = ServerInstance->Config->ConfValue("options");
-		bool bypassmodes = optionstag->getBool("invitebypassmodes", true);
-		bool modefromuser = optionstag->getBool("cyclehostsfromuser");
 
 		std::string current;
 		irc::spacesepstream defaultstream(optionstag->getString("exemptchanops"));
@@ -193,10 +191,10 @@ class CoreModChannel : public Module, public CheckExemption::EventListener
 		exemptions.swap(exempts);
 		limitmode.minlimit = minlimit;
 		cmdinvite.announceinvites = newstate;
-		joinhook.modefromuser = modefromuser;
+		joinhook.modefromuser = optionstag->getBool("cyclehostsfromuser");
 
 		Implementation events[] = { I_OnCheckKey, I_OnCheckLimit, I_OnCheckChannelBan };
-		if (bypassmodes)
+		if (optionstag->getBool("invitebypassmodes", true))
 			ServerInstance->Modules.Attach(events, this, sizeof(events)/sizeof(Implementation));
 		else
 		{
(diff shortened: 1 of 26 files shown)


src/coremods/core_channel/core_channel.cpp
```diff
@@ -148,16 +148,8 @@ class CoreModChannel : public Module, public CheckExemption::EventListener
...
-		if (optionstag->getBool("invitebypassmodes", true))
-			ServerInstance->Modules.Attach(events, this, sizeof(events)/sizeof(Implementation));
-		else
-		{
-			for (unsigned int i = 0; i < sizeof(events)/sizeof(Implementation); i++)
-				ServerInstance->Modules.Detach(events[i], this);
-		}
-
-		joinhook.modefromuser = optionstag->getBool("cyclehostsfromuser");
+		bool bypassmodes = optionstag->getBool("invitebypassmodes", true);
```
SaberUK: `getBool()` never throws so you can just set these after the settings that throw without using a temporary.

src/coremods/core_loadmodule.cpp
```diff
@@ -58,10 +58,11 @@ CmdResult CommandLoadmodule::Handle(User* user, const Params& parameters)
 class CommandUnloadmodule : public Command
 {
  public:
+	bool allowcoreunload;
 	/** Constructor for unloadmodule.
 	 */
 	CommandUnloadmodule(Module* parent)
-		: Command(parent,"UNLOADMODULE", 1)
+		: Command(parent,"UNLOADMODULE", 1), allowcoreunload(false)
```
SaberUK:
```suggestion
		: Command(parent,"UNLOADMODULE", 1)
		, allowcoreunload(false)
```

src/modules/m_customtitle.cpp
```diff
@@ -28,12 +28,39 @@ enum
 	RPL_WHOISSPECIAL = 320
 };
 
+struct CustomTitle
+{
+	const std::string name, password, hash, host, title, vhost;
```
SaberUK: These should be declared each on their own line rather than all at once.

src/modules/m_customtitle.cpp
```diff
@@ -28,12 +28,39 @@ enum
...
+
+	CustomTitle(const std::string& n, const std::string& p, const std::string& h, const std::string& hst, const std::string& t, const std::string& v)
+		: name(n), password(p), hash(h), host(hst), title(t), vhost(v)
```
SaberUK:
```suggestion
		: name(n)
		, password(p)
		, hash(h)
		, host(hst)
		, title(t)
		, vhost(v)
```

src/coremods/core_oper/core_oper.cpp
```diff
@@ -49,13 +37,28 @@ class CoreModOper : public Module
...
 		ConfigTag* security = ServerInstance->Config->ConfValue("security");
-		cmdkill.hidenick = security->getString("hidekills");
-		cmdkill.hideuline = security->getBool("hideulinekills");
+		std::string hidenick = security->getString("hidekills");
+		bool hideuline = security->getBool("hideulinekills");
+
+		ConfigTag* tag = ServerInstance->Config->ConfValue("power");
+		// The hash method for *BOTH* the die and restart passwords
+		const std::string hash = tag->getString("hash");
+		const std::string diepass = tag->getString("diepass", ServerInstance->Config->ServerName);
```
SaberUK: You should ensure these are at least one character log.

src/coremods/core_oper/core_oper.cpp
```diff
@@ -20,18 +20,6 @@
 #include "inspircd.h"
 #include "core_oper.h"
 
-namespace DieRestart
-{
-	bool CheckPass(User* user, const std::string& inputpass, const char* confentry)
```
SaberUK: To avoid duplicating code this should be retained and `correctpass` and `hash` should be passed in as arguments.

src/coremods/core_oper/core_oper.h
```diff
@@ -41,11 +33,18 @@ namespace DieRestart
  */
 class CommandDie : public Command
 {
+ protected:
+	bool CheckPass(User* user, const std::string& inputpass) const;
+
  public:
+	std::string hash, password;
```
SaberUK: `hash` is shared between both the DIE and RESTART commands so you should store it in the `Module` class and pass a reference to it in the constructor.

src/coremods/core_oper/core_oper.h
```diff
@@ -41,11 +33,18 @@ namespace DieRestart
...
+
 	/** Constructor for die.
 	 */
 	CommandDie(Module* parent);
 
+	CommandDie(Module* parent, const std::string& cmd);
```
SaberUK: What is this second constructor for? It doesn't appear to be used anywhere.

src/coremods/core_channel/core_channel.cpp
linuxdaemon: I wasn't sure if the logging on invalid values was going to be replaced with an exception or not

src/coremods/core_oper/core_oper.h
linuxdaemon: it is used by CommandRestart, since that is a subclass of CommandDie

src/coremods/core_oper/core_oper.cpp
linuxdaemon: Isn't an empty hash="" valid for plaintext passwords?

src/coremods/core_channel/core_channel.cpp
linuxdaemon: And from a readability standpoint it made more sense to me to group all config reads separate from usage of the values.

src/coremods/core_oper/core_oper.cpp
linuxdaemon: Both uses of this were replaced with `CommandDie::CheckPass()`, as `CommandRestart` subclasses `CommandDie`. Is there a usage I'm missing? I don't know of any duplicate code in this case

src/coremods/core_channel/core_channel.cpp
SaberUK: Config errors which are recoverable from (e.g. `getBool()` where it uses the default if the value is not a recognised boolean value) are not really an issue so it doesn't make sense for them to ever throw.

I feel that from a usability standpoint having a comment like `// The config was valid so we can apply it` would be better if necessary.

src/coremods/core_oper/core_oper.cpp
SaberUK: I was referring to diepass and restartpass not hash.

src/coremods/core_oper/core_oper.cpp
SaberUK: Ah, I was misreading but looking at the code it seems like CheckPass is now literally just a call to PassCompare so it doesn't seem to make sense to do anything other than just inlining that one function call.

linuxdaemon: Using the `inline` keyword would probably be beneficial, but I think leaving it as a separate method is better as if we wish to change the password logic for /DIE and /RESTART in the future we only have to modify `CommandDie::CheckPass()` instead of `CommanDie::Handle()` and `CommandRestart::Handle()`

SaberUK: In the future there won't be *any* die or restart passwords. The current `<power>` config is just legacy compatibility for 2.0 configs and in 3.0+1 it will be removed entirely with just a check for the server name as done in Plexus and some other server implementations.

src/modules/m_restrictchans.cpp
```diff
@@ -22,21 +22,22 @@
...
 	{
-		allowchans.clear();
+		AllowChans newallows;
 		ConfigTagList tags = ServerInstance->Config->ConfTags("allowchannel");
 		for(ConfigIter i = tags.first; i != tags.second; ++i)
 		{
-			ConfigTag* tag = i->second;
-			std::string txt = tag->getString("name");
-			allowchans.insert(txt);
+			newallows.insert(i->second->getString("name"));
```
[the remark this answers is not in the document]
linuxdaemon: yep, added

src/coremods/core_oper/core_oper.cpp
linuxdaemon: Replaced

src/coremods/core_oper/core_oper.cpp
linuxdaemon: ah, yep added

src/coremods/core_channel/core_channel.cpp
linuxdaemon: My thinking with `getBool()` was if a user set an invalid value for a boolean flag, allowing it to be applied may lead to unexpected or undesired behavior that wouldn't be obvious at first. It is logged, but that assumes the user reads stdout or the log file (if configured). For cases like that, most applications would generally throw an error. If we don't plan to I can adjust this logic, I just wanted to future-proof in case we did.

SaberUK: It is not currently planned for getBool to ever throw in the v3 release.

src/coremods/core_channel/core_channel.cpp
```diff
@@ -174,26 +164,41 @@ class CoreModChannel : public Module, public CheckExemption::EventListener
 			ServerInstance->Logs->Log(MODNAME, LOG_DEBUG, "Exempting prefix %c from %s", prefix, restriction.c_str());
 			exempts[restriction] = prefix;
 		}
-		exemptions.swap(exempts);
 
 		ConfigTag* securitytag = ServerInstance->Config->ConfValue("security");
 		const std::string announceinvites = securitytag->getString("announceinvites", "dynamic");
+		Invite::AnnounceState newstate;
```
SaberUK: Can you give this a less generic name? e.g. `newannouncestate`.

src/coremods/core_info/core_info.cpp
```diff
@@ -99,30 +99,32 @@ class CoreModInfo : public Module
...
-			if (cmdmotd.motds.find(motd) != cmdmotd.motds.end())
+			const std::string motd = tag->getString("motd", "motd");
+			if (newmotds.find(motd) != newmotds.end())
 				continue;
 
 			// We can't process the file if it doesn't exist.
 			ConfigFileCache::iterator file = ServerInstance->Config->Files.find(motd);
 			if (file == ServerInstance->Config->Files.end())
-				continue;
+				throw ModuleException(InspIRCd::Format("MOTD file %s not specified in <files>, at %s", motd.c_str(), tag->getTagLocation().c_str()));
```
SaberUK: This is an exception to the throw-on-invalid rule because the RFC states that `RPL_NOMOTD` should be sent instead of the MOTD if the file is missing.

src/coremods/core_whois.cpp
```diff
@@ -357,17 +357,21 @@ class CoreModWhois : public Module
 	{
 		ConfigTag* tag = ServerInstance->Config->ConfValue("options");
 		const std::string splitwhois = tag->getString("splitwhois", "no");
+		SplitWhoisState newstate;
```
SaberUK: Can you give this a less generic name? e.g. `newsplitstate`.

src/coremods/core_whois.cpp
```diff
@@ -357,17 +357,21 @@ class CoreModWhois : public Module
...
-			cmd.splitwhois = SPLITWHOIS_SPLIT;
+			newstate = SPLITWHOIS_SPLIT;
 		else if (stdalgo::string::equalsci(splitwhois, "splitmsg"))
-			cmd.splitwhois = SPLITWHOIS_SPLITMSG;
+			newstate = SPLITWHOIS_SPLITMSG;
 		else
 			throw ModuleException(splitwhois + " is an invalid <security:splitwhois> value, at " + tag->getTagLocation()); 
 
 		ConfigTag* security = ServerInstance->Config->ConfValue("security");
+		// In the future this may error if an invalid value is set
```
SaberUK: This is no longer relevant I think?

src/modules/m_cgiirc.cpp
```diff
@@ -338,12 +338,12 @@ class ModuleCgiIRC
...
 
-		// The host configuration was valid so we can apply it.
-		hosts.swap(identhosts);
-		cmd.hosts.swap(webirchosts);
-
 		// Do we send an oper notice when a m_cgiirc client has their IP changed?
 		cmd.notify = ServerInstance->Config->ConfValue("cgiirc")->getBool("opernotice", true);
+
+		// The configuration was valid so we can apply it.
+		hosts.swap(identhosts);
```
SaberUK: This was fine as it was before?

src/modules/m_customtitle.cpp
```diff
@@ -88,6 +116,28 @@ class ModuleCustomTitle : public Module, public Whois::LineEventListener
...
+		ConfigTagList tags = ServerInstance->Config->ConfTags("title");
+		CustomVhostMap newtitles;
+		for (ConfigIter i = tags.first; i != tags.second; ++i)
+		{
+			reference<ConfigTag> tag = i->second;
+			std::string name = tag->getString("name", "", 1);
+			if (name.empty())
+				throw ModuleException("<title:name> is empty! at " + tag->getTagLocation());
+
+			std::string pass = tag->getString("password");
```
SaberUK: Password should be validated as non-empty here.

src/modules/m_flashpolicyd.cpp
```diff
@@ -113,6 +112,7 @@ class ModuleFlashPD : public Module
 				ServerInstance->Logs->Log(MODNAME, LOG_DEFAULT, error_message);
 				ServerInstance->SNO->WriteGlobalSno('a', error_message);
 				policy_reply.clear();
+				throw ModuleException(error_message + " at " + tag->getTagLocation());
```
SaberUK: You can get rid of the logging, the snomask write, and the clearing of policy_reply here.

src/modules/m_helpop.cpp
```diff
@@ -146,10 +146,9 @@ class ModuleHelpop : public Module, public Whois::EventListener
 				throw ModuleException("m_helpop: Helpop file is missing important entry 'start'. Please check the example conf.");
 			}
 
-			helpop_map.swap(help);
-
 			ConfigTag* tag = ServerInstance->Config->ConfValue("helpmsg");
 			cmd.nohelp = tag->getString("nohelp", "There is no help for the topic you searched for. Please try again.", 1);
+			helpop_map.swap(help);
```
SaberUK: This was fine before I think?

src/modules/m_hidemode.cpp
```diff
@@ -48,9 +48,10 @@ class Settings
 			if (!modename.empty() && rank)
```
SaberUK: This should throw if the mode name is not empty and/or there is no rank.

src/modules/m_securelist.cpp
```diff
@@ -36,15 +38,22 @@ class ModuleSecureList : public Module
...
-		allowlist.clear();
+		AllowList newallows;
 
 		ConfigTagList tags = ServerInstance->Config->ConfTags("securehost");
 		for (ConfigIter i = tags.first; i != tags.second; ++i)
-			allowlist.push_back(i->second->getString("exception"));
+		{
+			std::string host = i->second->getString("exception");
+			if (host.empty())
+				throw ModuleException("<securehost:exception> is a requirewd field! at " + i->second->getTagLocation());
```
SaberUK:
```suggestion
				throw ModuleException("<securehost:exception> is a required field at " + i->second->getTagLocation());
```

src/modules/m_cgiirc.cpp
linuxdaemon: yeah, the move was done when I thought `getBool()` and similar might throw in the future

src/modules/m_helpop.cpp
linuxdaemon: same as `m_cgiirc`

src/modules/m_pbkdf2.cpp
```diff
@@ -139,39 +139,64 @@ class PBKDF2Provider : public HashProvider
 	}
 };
 
+struct ProviderConfig
+{
+	unsigned long iterations, dkey_length;
```
SaberUK:
```suggestion
	unsigned long iterations;
	unsigned long dkey_length;
```

PR #1538, opened 2018-12-20
Only parse valid durations, don't treat invalid multipliers as seconds

linuxdaemon:
- Only accept valid multipliers
- Add a new signature for `InspIRCd::Duration` to add error checking, `bool InspIRCd::Duration(const std::string& str, unsigned long& duration); which will return true or false depending if the duration string is valid
- Add in checks to most commands and mode parameters to not accept invalid durations


src/configparser.cpp
```diff
@@ -609,14 +609,14 @@ unsigned long ConfigTag::getDuration(const std::string& key, unsigned long def,
 	if (!readString(key, duration))
 		return def;
 
-	if (!InspIRCd::IsValidDuration(duration))
+	unsigned long ret = def;
```
SaberUK:
```suggestion
	unsigned long ret;
```

The value you assign to ret will never actually be used so you don't need a default.

src/coremods/core_channel/cmd_invite.cpp
```diff
@@ -51,7 +51,15 @@ CmdResult CommandInvite::Handle(User* user, const Params& parameters)
 		if (parameters.size() >= 3)
 		{
 			if (IS_LOCAL(user))
-				timeout = ServerInstance->Time() + InspIRCd::Duration(parameters[2]);
+			{
+				unsigned long duration;
+				if (!InspIRCd::Duration(parameters[2], duration))
+				{
+					user->WriteNotice("*** Invalid duration");
```
SaberUK: This isn't a very descriptive response. Can you change this (and other places you've used this message) with something that tells the user what specifically it is a response to.

e.g. 

```
** INVITE: expiry duration is invalid

File: src/helperfuncs.cpp [hunk 3]
```diff
@@ -356,75 +356,74 @@ void InspIRCd::CheckRoot()
...
+	return out;
 }
 
 bool InspIRCd::IsValidDuration(const std::string& duration)
 {
 	for (std::string::const_iterator i = duration.begin(); i != duration.end(); ++i)
 	{
 		unsigned char c = *i;
-		if (((c >= '0') && (c <= '9')) || (c == 's') || (c == 'S'))
+		if (((c >= '0') && (c <= '9')))
```
SaberUK: This if statement can be removed entirely.

src/helperfuncs.cpp
```diff
@@ -356,75 +356,74 @@ void InspIRCd::CheckRoot()
...
 			continue;
 
-		if (duration_multi[c] == 1)
+		if (duration_multi[c] == 0)
```
SaberUK: This might look cleaner? What do you think?

```suggestion
		if (!duration_multi[c])
```

src/coremods/core_channel/cmd_invite.cpp
linuxdaemon: I probably won't do `*** THING: whatever` format specifically as that is too close to a snotice, but yep will do

src/modules/m_chanhistory.cpp
```diff
@@ -72,8 +72,8 @@ class HistoryMode : public ParamMode<HistoryMode, SimpleExtItem<HistoryList> >
 		}
 
 		unsigned int len = ConvToNum<unsigned int>(parameter.substr(0, colon));
-		unsigned int time = InspIRCd::Duration(duration);
-		if (len == 0 || (len > maxlines && IS_LOCAL(source)))
+		unsigned long time;
+		if (!InspIRCd::Duration(duration, time) || len == 0 || (len > maxlines && IS_LOCAL(source)))
```
SaberUK: This causes a duration of 0s to be allowed which wasn't previously. It should be something like this:

```suggestion
		if (!InspIRCd::Duration(duration, time) ||  time == 0 || len == 0 || (len > maxlines && IS_LOCAL(source)))
```

src/helperfuncs.cpp
linuxdaemon: Nope, it can't. If it was removed, any digit in the string would be invalid, as they aren't set to a non-zero value in the array

SaberUK: Ah, sorry yes. I misread.

src/coremods/core_channel/cmd_invite.cpp
SaberUK: This format is used in other places (grep for `\*\*\* [A-Z]+:`) but sure.

src/helperfuncs.cpp
linuxdaemon: hmm, yeah that seems a bit cleaner

src/modules/m_chanhistory.cpp
linuxdaemon: A duration of `0` is valid for chanhistory, there wasn't a check for it previously.

src/helperfuncs.cpp
```diff
@@ -356,75 +356,74 @@ void InspIRCd::CheckRoot()
...
+	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
 };
 
-unsigned long InspIRCd::Duration(const std::string &str)
+bool InspIRCd::Duration(const std::string& str, unsigned long& duration)
 {
-	unsigned char multiplier = 0;
 	long total = 0;
```
SaberUK: It isn't a bug in your code but `total` and `subtotal` here could overflow to a negative number which is undesirible. Can you make them both unsigned?

PR #1539, opened 2018-12-20
Add the ability to disable console colors in stdout logging

linuxdaemon:
- Disables console colors when not attached to a TTY
- Add a preprocessor option to disable console colors completely

### DIFF
--- a/include/consolecolors.h
+++ b/include/consolecolors.h
@@ -19,7 +19,46 @@
 
 #include <ostream>
 
-#ifdef _WIN32
+inline std::ostream& con_green(std::ostream &s);
+inline std::ostream& con_red(std::ostream &s);
+inline std::ostream& con_white(std::ostream &s);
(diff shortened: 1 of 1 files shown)


include/consolecolors.h
```diff
@@ -65,33 +104,53 @@ inline std::ostream& con_reset(std::ostream &s)
 
 #else
 
+#include <unistd.h>   // for isatty()
+#include <stdio.h>    // for fileno()
+
+static inline bool IsRunningInteractive()
```
SaberUK: This should be in an anonymous namespace rather than being static.

SaberUK: Also, you could do something like this here:

```cpp
#if defined INSPIRCD_DISABLE_COLORS
    return false;
#else
    return isatty(fileno(stdout));
#endif
```

PR #1541, opened 2019-01-04
Merge m_restrictchans and m_regonlycreate due to converging functionality

linuxdaemon:
- Add `channels/restricted-create` oper priv to allow an oper to bypass the restrictchans checks
- Allow `<allowchannel:name>` to be a wildcard pattern
- Add `<restrictchans:allowregistered>` to allow registered users to create channels


docs/conf/modules.conf.example
```diff
@@ -1710,12 +1702,17 @@
 #<module name="repeat">
 
 #-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#
-# Restricted channels module: Allows only opers to create channels.
+# Restricted channels module: Allows only opers or registered users to create channels.
```
SaberUK:
```suggestion
# Restricted channels module: Allows only opers or registered users to
# create channels.
```

docs/conf/modules.conf.example
```diff
@@ -1710,12 +1702,17 @@
...
 #
 # You probably *DO NOT* want to load this module on a public network.
 #
 #<module name="restrictchans">
 
+#<restrictchans allowregistered="false">
```
SaberUK: This should be documented (also maybe use "no" instead of "false").

src/modules/m_restrictchans.cpp
```diff
@@ -27,13 +27,38 @@ typedef insp::flat_set<std::string, irc::insensitive_swo> AllowChans;
 class ModuleRestrictChans : public Module
 {
 	AllowChans allowchans;
+	UserModeReference regmode;
+	bool allowregistered;
+
+	bool CanCreateChannel(LocalUser* user, const std::string& name)
+	{
+		if (allowregistered && user->IsModeSet(regmode))
```
SaberUK: Checking if the user has an account name is a more reliable way of checking that users are logged in.

src/modules/m_restrictchans.cpp
```diff
@@ -42,26 +67,24 @@ class ModuleRestrictChans : public Module
...
+		if (!chan && !CanCreateChannel(user, cname))
+			return MOD_RES_DENY;
+
 		return MOD_RES_PASSTHRU;
 	}
 
 	Version GetVersion() CXX11_OVERRIDE
 	{
-		return Version("Only opers may create new channels if this module is loaded",VF_VENDOR);
+		return Version("Only opers may create new channels if this module is loaded", VF_VENDOR);
```
SaberUK: This description isn't accurate anymore.

src/modules/m_restrictchans.cpp
linuxdaemon: That's true, I copied that logic from `regonlycreate` but thinking about it, yeah that would only match if the user is actively using their registered nickname instead of just being logged in. Will fix.

src/modules/m_restrictchans.cpp
```diff
@@ -21,19 +21,44 @@
...
 
 class ModuleRestrictChans : public Module
 {
 	AllowChans allowchans;
+	bool allowregistered;
+
+	bool CanCreateChannel(LocalUser* user, const std::string& name)
+	{
+		const AccountExtItem* accountext = GetAccountExtItem();
+		if (allowregistered && accountext->get(user))
```
SaberUK: GetAccountExtItem() may return NULL if the services_account module is not loaded.

```suggestion
		if (allowregistered && accountext && accountext->get(user))
```

src/modules/m_restrictchans.cpp
```diff
@@ -42,26 +67,24 @@ class ModuleRestrictChans : public Module
...
+		if (!chan && !CanCreateChannel(user, cname))
+			return MOD_RES_DENY;
+
 		return MOD_RES_PASSTHRU;
 	}
 
 	Version GetVersion() CXX11_OVERRIDE
 	{
-		return Version("Only opers may create new channels if this module is loaded",VF_VENDOR);
+		return Version("Allows restricting who can create channels on the network", VF_VENDOR);
```
SaberUK: The module is local to the server so this would be more accurate.

```suggestion
		return Version("Allows restricting who can create channels", VF_VENDOR);
```

PR #1543, opened 2019-01-07
Add translation for m_override mode name

linuxdaemon:
In the 2.0 extras module it's called permitoverride, in 3.0 it's just
override


SaberUK:
I changed the mode in the extras module instead of this. Thanks though.

https://github.com/inspircd/inspircd-extras/compare/052673016f44...f32380dd49a2

PR #1548, opened 2019-01-11
The <ident:bind> setting doesn't exist anymore, remove the reference to it from the config.

Robby-:


SaberUK:
Would it be possible for you to link the commit where this was removed?

Robby-:
This setting was removed in commit 0a1fe6cf293bef91b419f57374b96a6d2aec636e.

PR #1549, opened 2019-01-11
Redo OnSetEndPoint logic to fix duplicate clones

linuxdaemon:
Make sure to recheck clones and connect classes when changing a
user's endpoint

Also moves all rechecks in to `SetClientIP()` so modules don't have to worry about it

### DIFF
--- a/include/inspsocket.h
+++ b/include/inspsocket.h
@@ -313,9 +313,7 @@ class CoreExport StreamSocket : public EventHandler
 	 * @param local The new local endpoint.
 	 * @param remote The new remote endpoint.
 	 */
-	virtual bool OnSetEndPoint(const irc::sockets::sockaddrs& local, const irc::sockets::sockaddrs& remote) {
-		return true;
-	}
+	virtual bool OnSetEndPoint(const irc::sockets::sockaddrs& local, const irc::sockets::sockaddrs& remote);
 
 	/** Send the given data out the socket, either now or when writes unblock
 	 */
--- a/include/inspsocket.h
+++ b/include/inspsocket.h
@@ -313,7 +313,9 @@ class CoreExport StreamSocket : public EventHandler
 	 * @param local The new local endpoint.
 	 * @param remote The new remote endpoint.
 	 */
-	virtual void OnSetEndPoint(const irc::sockets::sockaddrs& local, const irc::sockets::sockaddrs& remote) { }
+	virtual bool OnSetEndPoint(const irc::sockets::sockaddrs& local, const irc::sockets::sockaddrs& remote) {
+		return true;
+	}
 
 	/** Send the given data out the socket, either now or when writes unblock
 	 */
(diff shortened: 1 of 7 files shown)


include/inspsocket.h
```diff
@@ -313,7 +313,9 @@ class CoreExport StreamSocket : public EventHandler
 	 * @param local The new local endpoint.
 	 * @param remote The new remote endpoint.
 	 */
-	virtual void OnSetEndPoint(const irc::sockets::sockaddrs& local, const irc::sockets::sockaddrs& remote) { }
+	virtual bool OnSetEndPoint(const irc::sockets::sockaddrs& local, const irc::sockets::sockaddrs& remote) {
+		return true;
```
SaberUK: Can you put the implementation of this in a .cpp file rather than a header.

src/users.cpp
```diff
@@ -758,14 +759,29 @@ bool LocalUser::SetClientIP(const std::string& address, bool recheck_eline)
...
-			this->exempt = (ServerInstance->XLines->MatchesLine("E", this) != NULL);
+	if (sa == client_sa)
+		return;
 
-		FOREACH_MOD(OnSetUserIP, (this));
-	}
+	ServerInstance->Users->RemoveCloneCounts(this);
+
+	User::SetClientIP(sa);
+	if (recheck_eline)
```
SaberUK: Is the recheck_eline parameter ever false?

If not, can you move the setting of exempt and the CheckLines call to core_xline?

src/modules/m_cgiirc.cpp
```diff
@@ -38,21 +38,7 @@ enum
 static void ChangeIP(LocalUser* user, const irc::sockets::sockaddrs& sa)
```
SaberUK: This is pretty much empty now so it can be removed and the two lines of code inlined to the places that call it.

src/modules/m_haproxy.cpp
```diff
@@ -261,7 +261,8 @@ class HAProxyHook : public IOHookMiddle
 						break;
 				}
 
-				sock->OnSetEndPoint(server, client);
+				if (!sock->OnSetEndPoint(server, client))
+					return 0;
```
SaberUK: I think this should return -1 here?

include/inspsocket.h
```diff
@@ -313,7 +313,7 @@ class CoreExport StreamSocket : public EventHandler
 	 * @param local The new local endpoint.
 	 * @param remote The new remote endpoint.
 	 */
-	virtual void OnSetEndPoint(const irc::sockets::sockaddrs& local, const irc::sockets::sockaddrs& remote) { }
```
SaberUK: Can you update the doxygen documentation for this to add a `@return` line.

src/modules/m_haproxy.cpp
linuxdaemon: ah yep, fixing now

PR #1552, opened 2019-01-11
Add option to make a connect block not match users with an account

linuxdaemon:
Allows setting a connect class which will only match a user who has not
logged in

### DIFF
--- a/docs/conf/inspircd.conf.example
+++ b/docs/conf/inspircd.conf.example
@@ -336,6 +336,16 @@
          port="6697,9999">
 
 <connect
+         name="loggedin"
+         parent="main"
+
+         # account: Control how the class matches users with an account (requires m_services_account)
+         #  allow: Allow a user with an account to connect, not requiring an account though (default)
+         #  deny: Only allow a user that isn't logged in to match this connect class
+         #  require: Only match a user with an account
+         account="require">
+
+<connect
          # name: Name to use for this connect block. Mainly used for
          # connect class inheriting.
          name="main"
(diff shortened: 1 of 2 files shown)


SISheogorath:
Can find a better way to do that? Somehow I already see people putting both options into one connect class wondering why their connects fail. There should be at least some debug logging details and, as mentioned before, maybe we can find a better way to put it.

How about a config called accountpolicy? Which then can be: `required`, `ignored` and `forbidden` with `ignored` as default?

linuxdaemon:
That's a decent idea, just didn't want to break compat with 2.0 in this case. If @SaberUK is fine with losing that compat, then I'll definitely implement it, it'll be a lot easier to deal with.

linuxdaemon:
As suggested by @Robby- the current idea is `<account account="deny/allow/require">` for only match no account, allow any state, require account respectively. Unless anyone has any objections, I may go ahead and update this PR to use that.

With respect to the old syntax though, I think leaving the handling in for it should be fine

linuxdaemon:
I think this now fits the currently agreed terminology from the discussion on IRC. @Robby- thoughts?

Robby-:
@linuxdaemon: I don't really have anything else to add, LGTM

linuxdaemon:
Anything holding this one up @SaberUK ? I've had a few people express interest in it for 3.0 (and a backport for 2.0 I'll do in contrib)

linuxdaemon:
@SadieCat Any movement on this? it came up again, it's very useful for making it so SASL users can bypass dnsbl checks

PR #1563, opened 2019-01-24
Don't allow spaces in UNIX listener paths

linuxdaemon:
Allowing spaces here can cause a whole host of issues as the user's host would contain a space unless set otherwise by webirc, haproxy, or similar.


SaberUK:
Instead of doing this can we sanitise the value when setting the sockaddr path in `ListenSocket::OnEventHandlerRead()`?

linuxdaemon:
What all do we want to sanitize? Pretty much all characters (including newlines) are allowed in file paths

PR #1565, opened 2019-01-26
Give m_setname its own snomask

KoraggKnightWolf:
This pull request gives m_setname its own snomask char, s, as it is not used on InspIRCd 2.x. The reason is so that opers can be informed about users using /SETNAME yet it does not get propogated to the announcement snomask a. This could lead to important messages "drowning". 
The reason for this pull request instead of simply disabling m_setname or making it oper only, is that features should not have to be disabled just because they could be abused as in the end, any command could be abused.

Regards,

Koragg

### DIFF
--- a/src/modules/m_setname.cpp
+++ b/src/modules/m_setname.cpp
@@ -45,7 +45,7 @@ class CommandSetname : public Command
 
 		if (user->ChangeName(parameters[0].c_str()))
 		{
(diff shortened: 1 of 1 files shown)


SaberUK:
The v2 branch is not accepting changes at this time so I have retargeted this to v2. Can you rebase please.

KoraggKnightWolf:
Umm you mean retargeted to v3? Because you said v2 is not accepting yet it is retargeted to v2 should I re-make it for v3? I surely can if you meant that.

PR #1566, opened 2019-01-29
Expand searching in m_httpd_stats, add global handling of GET parameters

linuxdaemon:
- Improves URI parsing to properly read the separate portions (`path`, `query`, `fragment`)
- Add filter and sort options to `/stats/users`
  - `sortby` (accepts `nick` and `idle`)
  - `limit`
  - `showunreg`
  - `localonly`
  - `desc`
- Add more data to `/stats/users`
  - `signon`
  - `lastmsg`
  - `age`
  - `local`

### DIFF
--- a/include/modules/httpd.h
+++ b/include/modules/httpd.h
@@ -31,10 +31,59 @@
 #include <sstream>
 #include <map>
 
+class HTTPQueryParameters : public insp::flat_multimap<std::string, std::string>
+{
+ public:
+	bool get(const std::string& key, std::string& value) const
+	{
+		const_iterator it = find(key);
+		if (it == end())
+			return false;
+
+		value = it->second;
+		return true;
+	}
+
+	std::string getString(const std::string& key, const std::string& def = "") const
+	{
+		const_iterator it = find(key);
+		std::string value;
+		if (!get(key, value))
+			return def;
+
+		return value;
+	}
+
+	template <typename T>
+	T getNum(const std::string& key, T def = 0) const
+	{
+		const_iterator it = find(key);
+		std::string value;
+		if (!get(key, value))
+			return def;
+
+		return ConvToNum<T>(value);
+	}
+
+	unsigned long getDuration(const std::string& key, unsigned long def = 0) const
+	{
+		unsigned long value;
+		if (!InspIRCd::Duration(getString(key, "0"), value))
+			return def;
+
+		return value;
+	}
+
+	bool getBool(const std::string& key, bool def = false) const
+	{
+		return getNum<bool>(key, def);
+	}
+};
+
 struct HTTPRequestURI
 {
 	std::string path;
-	insp::flat_multimap<std::string, std::string> query_params;
+	HTTPQueryParameters query_params;
 	std::string fragment;
 
 	HTTPRequestURI(const std::string& uri)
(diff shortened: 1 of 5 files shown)


include/modules/httpd.h
```diff
@@ -25,11 +25,51 @@
...
 
+struct HTTPRequestURI
+{
+	std::string path;
+	insp::flat_multimap<std::string, std::string> query_params;
+	std::string fragment;
+
+	HTTPRequestURI(const std::string& uri)
+	{
+		http_parser_url url;
```
SaberUK: You don't need to create new instances of this for every request do you?

include/modules/httpd.h
```diff
@@ -25,11 +25,51 @@
...
+		http_parser_url_init(&url);
+		http_parser_parse_url(uri.c_str(), uri.size(), 0, &url);
```
SaberUK: Can all of this code be removed from the public header? I'd like to avoid having internal dependencies leak into the ABI.

src/modules/m_httpd.cpp
```diff
@@ -280,11 +281,13 @@ class HttpServerSocket : public BufferedSocket, public Timer, public insp::intru
...
 		std::string method = http_method_str(static_cast<http_method>(parser.method));
-		HTTPRequest acl(method, uri, &headers, this, ip, body);
+		HTTPRequestURI parsed;
+		ParseURI(uri, parsed);
+		HTTPRequest acl(method, uri, parsed, &headers, this, ip, body);
 		FIRST_MOD_RESULT_CUSTOM(*aclevprov, HTTPACLEventListener, OnHTTPACLCheck, MOD_RESULT, (acl));
 		if (MOD_RESULT != MOD_RES_DENY)
 		{
-			HTTPRequest url(method, uri, &headers, this, ip, body);
+			HTTPRequest url(method, uri, parsed, &headers, this, ip, body);
```
SaberUK: Is there any reason to still pass uri here?

linuxdaemon: `HTTPRequest::GetURI()` returns the plain string prior to parsing and `GetParsedURI()` returns the post-parsing one

include/modules/httpd.h
```diff
@@ -30,6 +30,60 @@
 #include <sstream>
 #include <map>
 
+class HTTPQueryParameters : public insp::flat_multimap<std::string, std::string>
```
SaberUK: I can't imagine the methods in this class would enough use to justify their existence and they seem like a leaky abstraction considering how you can (apparently?) have multiple values for each key.

src/modules/m_httpd.cpp
SaberUK: Yes, but I can't see why would you ever need `GetURI()`.

linuxdaemon: I'm not sure, the original reason was to leave it for people to use, but I guess just storing it on the parsed uri so it can be access from GetParsedURI().uri could work, idk

include/modules/httpd.h
linuxdaemon: The idea was to allow for easier access to simple values where grabbing the first value passed for a key is fine and allow for easy defaults

src/modules/m_httpd.cpp
SaberUK: You're probably better off just deleting it tbh. I can't think of a single reason why anyone would want to use it.

PR #1583, opened 2019-02-20
Improve the handling of config X-lines and filters.

### LINKED ISSUE #1577: Rehashing always sends removal snotices of config-defined xline and filter tags, even when not actually removed
**Description**
Rehashing will always show removal snotices for xlines and filters that are defined in the configuration files but were not actually removed from them, causing rehashes to always be very noisy/floody when having a couple hundred tags defined.

1. Define some `<badhost>`, `<badip>`, `<badnick>` or `<keyword>` tags in the config file
2. Rehash
3. Rehash again without removing any of these tags
4. Observe getting these snotices upon every rehash

See description/steps to reproduce.

To only get snotices for the tags that actually have been removed from the configuration files.

`InspIRCd-3.0.0-318d50b`

genius3000:
#### X-lines:
* Prevent xline_db from saving config X-lines as this is unnecessary and can cause weirdness or 'almost' duplicated X-lines to exist.
- On Rehash:
  * Allow a config X-line to replace an existing X-line if:
    * The existing X-line is not a config line, or
    * The existing X-line is a config line but the reason changed.
  * Compare the existing X-lines to a list of the newly read config X-lines and only expire those that are no longer configured.  

This prevents the extra noise during a rehash by only expiring the `badhost, badip, badnick, and exception` tags that were actually removed. If any of the aforementioned tags have a changed `reason`, the existing X-line will be expired and the new one added (mimicking behavior of the forced removal). Also allow the replacement (expire and add new) of an existing, non-config, X-line with a newly defined config X-line.

#### Filters:
* Change the addition of a config filter to use the existing AddFilter() function. This prevents being able to configure multiple filters with the same patterns. AddFilter() now takes an additional parameter (`bool config`) to set the appropriate flag for `from_config`.

- On Rehash:
  * Continue to remove all config defined filters, as various parameters to each pattern may have been changed.
  * Keep a list of the removed filters, erasing a pattern if it has been successfully added as a filter again.
  * Send an SNOTICE for each of the remaining filter patterns that were actually removed.  

This again prevents the extra noise during a rehash as all config filters are removed and re-added, by only sending notices for the actually removed filters. Also fix a possible bug of defining multiple filters with the same pattern in the config.

Resolves #1577.

### DIFF
--- a/include/xline.h
+++ b/include/xline.h
@@ -515,8 +515,9 @@ class CoreExport XLineManager
 	/** Expire a line given two iterators which identify it in the main map.
 	 * @param container Iterator to the first level of entries the map
 	 * @param item Iterator to the second level of entries in the map
+	 * @param silent If true, doesn't send an expiry SNOTICE.
 	 */
-	void ExpireLine(ContainerIter container, LookupIter item);
+	void ExpireLine(ContainerIter container, LookupIter item, bool silent = false);
 
 	/** Apply any new lines that are pending to be applied.
 	 * This will only apply lines in the pending_lines list, to save on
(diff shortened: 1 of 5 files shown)


genius3000:
As I posted this I realized a downside to not expiring all the config defined X-lines: changing a reason would fail to update. I'll try do something similar to the filter approach to handle this.

PR #1588, opened 2019-03-11
Update most URLs to use HTTPS if available, and fix some dead ones too.

Robby-:


SaberUK:
Merged without the changes to headers and the vendor directory in ac0d5ab.

PR #1599, opened 2019-03-26
Add the ability to hide numbered lines in RFC1459 protocol

EpicNet:
Please make it possible for the server owner to enable and disable (hide) any numbered server messages for display to users. Ways of management through a command set or through a config:

**Syntax:** /setnumlines <on|off|list> [numbers]
**Config:** `<setnumlines hide="numbers">`

**Example:** /setnumlines off 005,307,312,671,317,318
**Config:** `<setnumlines hide="005,307,312,671,317,318">`

The following lines will be hide for users to display:

Number line 005 = "info when connecting to the server"
Number line 307 = "is a registered nick"
Number line 312 = "through which server was the connection"
Number line 671 = "is using a secure connection"
Number line 317 = "idle user silence time..."
Number line 318 = "End of /WHOIS list"

**Example:** /setnumlines list

**Server response:**

> *** Following lines are hide: "005,307,312,671,317,318". Command to enable: /setnumlines on [numbers]

It was an example of certain lines. It is necessary to do this for all possible line numbers that will not affect the important functions of the server.

To allow this command in the operator block in "opers.conf" need to specify: `setnumlines="yes"`


grawity:
...why would a server operator want to disable 005 and 318?

EpicNet:
> ...why would a server operator want to disable 005 and 318?
> 
> I understand trying to hide the rest, which is just miscellaneous WHOIS info, but trying to hide 005 and "end of /foo" just risks breaking clients and doesn't really protect anything at all.

I gave as an example random numbers of lines. The operator may want to update his server in order to hide all unnecessary message lines for the user in order to reduce the received information to a minimum.

SaberUK:
Doing this will completely break your server for most clients. Just because you think those messages don't do anything doesn't mean they don't actually do anything.

If you don't want people to see those lines then write a client that doesn't show them.

PR #1624, opened 2019-05-03
Clarify the require* connect block options and document the sslinfo module operonly option.

### LINKED ISSUE #600: Please document <connect requiressl="trusted">
It is entirely missing in the example configs, and http://wiki.inspircd.org/Secure_Sockets_Layer#Signed_Client_Certificates (which seems to be outdated in places because it talks about v2.1) seems to be the only place on the wiki documenting this option.
In the code, the option is actually used (for instance here: https://github.com/inspircd/inspircd/blob/master/src/modules/m_sslinfo.cpp#L231).

Please note that I've only tested that "trusted" works as it should in the positive case (correct, non-revoked clientcert issued by the CA), but not the negative cases, nor the other, accept-all-ssl case ("true", or "on", or whatever it is).

Robby-:
This closes #600.

### DIFF
--- a/docs/conf/inspircd.conf.example
+++ b/docs/conf/inspircd.conf.example
@@ -305,14 +305,24 @@
          # module be loaded as well.
          modes="+x"
 
-         # requireident, requiressl, requireaccount: require that users of this
-         # block have a valid ident response, use SSL, or have authenticated.
-         # Requires ident, sslinfo, or the services_account module, respectively.
-         requiressl="on"
-         # NOTE: For requireaccount, you must complete the signon prior to full
-         # connection. Currently, this is only possible by using SASL
-         # authentication; passforward and PRIVMSG NickServ happen after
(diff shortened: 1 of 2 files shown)


docs/conf/inspircd.conf.example
```diff
@@ -305,14 +305,25 @@
...
-         requiressl="on"
-         # NOTE: For requireaccount, you must complete the signon prior to full
-         # connection. Currently, this is only possible by using SASL
-         # authentication; passforward and PRIVMSG NickServ happen after
-         # your final connect block has been found.
+         # requireident: Require that users of this block have a valid ident response.
+         # Requires the ident module to be loaded.
+         #requireident="yes"
+
+         # requiressl: Require that users of this block use SSL certificates.
```
SaberUK: requiressl is that all users of the block must use a SSL connection not must use SSL certificates.

PR #1629, opened 2019-05-05
Adding OnUserPreQuit event to allow quit message changes by modules.

iwalkalone69:
The purpose of this pull request is to allow modifying quit message like we can do with part message in OnUserPart event.

The change is just removing const from parameter to allow its modification in modules.


SaberUK:

iwalkalone69:
No need for apologies! Thank you for merging :-)

PR #1633, opened 2019-05-06
Re-order the includes to fix C++03 builds.

ensra:
inspircd.h defines __STDC_LIMIT_MACROS to ensure that C99 int type limits are defined, however, if <stdint.h> is included implicitly before inspircd.h, the build fails due to the C99 integer type limits being undefined.

### DIFF
--- a/src/modules.cpp
+++ b/src/modules.cpp
@@ -24,9 +24,9 @@
  */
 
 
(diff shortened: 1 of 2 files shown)


SaberUK:

PR #1641, opened 2019-05-17
configure: --log-dir in --system mode uses the value for --module-dir

Natolumin:
There was a typo or copy-paste error, making configure invoked with --system overwrite the
passed --log-dir option with the value of --module-dir


SaberUK:

PR #1644, opened 2019-05-19
m_filter registered users flag exception for filters

H7-25:
Added flag_no_registered to m_filter module to exclude registered users from filters

### DIFF
--- a/src/modules/m_filter.cpp
+++ b/src/modules/m_filter.cpp
@@ -126,7 +126,7 @@ class FilterResult
 		std::string flags;
 		if (flag_no_opers)
 			flags.push_back('o');
-        if (flag_no_registered)
+		if (flag_no_registered)
             flags.push_back('r');
 		if (flag_part_message)
 			flags.push_back('P');
--- a/src/modules/m_filter.cpp
+++ b/src/modules/m_filter.cpp
@@ -92,7 +92,7 @@ class FilterResult
 					flag_no_opers = true;
 				break;
 				case 'r':
-				    flag_no_registered = true;
+					flag_no_registered = true;
 				break;
 				case 'P':
 					flag_part_message = true;
--- a/src/modules/m_filter.cpp
+++ b/src/modules/m_filter.cpp
@@ -127,7 +127,7 @@ class FilterResult
 		if (flag_no_opers)
 			flags.push_back('o');
 		if (flag_no_registered)
-            flags.push_back('r');
+			flags.push_back('r');
 		if (flag_part_message)
 			flags.push_back('P');
 		if (flag_quit_message)
--- a/src/modules/m_filter.cpp
+++ b/src/modules/m_filter.cpp
@@ -91,9 +91,6 @@ class FilterResult
 				case 'o':
 					flag_no_opers = true;
 				break;
-				case 'r':
-					flag_no_registered = true;
-				break;
 				case 'P':
 					flag_part_message = true;
 				break;
@@ -108,6 +105,9 @@ class FilterResult
 				break;
 				case 'c':
 					flag_strip_color = true;
+				break;
+				case 'r':
+					flag_no_registered = true;
 				break;
 				case '*':
 					flag_no_opers = flag_no_registered = flag_part_message = flag_quit_message =
@@ -126,8 +126,6 @@ class FilterResult
 		std::string flags;
 		if (flag_no_opers)
 			flags.push_back('o');
-		if (flag_no_registered)
-			flags.push_back('r');
 		if (flag_part_message)
 			flags.push_back('P');
 		if (flag_quit_message)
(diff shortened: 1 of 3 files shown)


src/modules/m_filter.cpp
```diff
@@ -89,6 +91,9 @@ class FilterResult
 				case 'o':
 					flag_no_opers = true;
 				break;
+				case 'r':
+				    flag_no_registered = true;
```
Robby-:
```suggestion
					flag_no_registered = true;
```

src/modules/m_filter.cpp
```diff
@@ -121,6 +126,8 @@ class FilterResult
 		std::string flags;
 		if (flag_no_opers)
 			flags.push_back('o');
+        if (flag_no_registered)
```
Robby-:
```suggestion
		if (flag_no_registered)
```

src/modules/m_filter.cpp
```diff
@@ -121,6 +126,8 @@ class FilterResult
...
+            flags.push_back('r');
```
Robby-:
```suggestion
			flags.push_back('r');
```

src/modules/m_filter.cpp
```diff
@@ -305,8 +312,12 @@ CmdResult CommandFilter::Handle(User* user, const Params& parameters)
 
 bool ModuleFilter::AppliesToMe(User* user, FilterResult* filter, int iflags)
 {
+    const AccountExtItem* accountext = GetAccountExtItem();
```
Robby-:
```suggestion
	const AccountExtItem* accountext = GetAccountExtItem();
```

src/modules/m_filter.cpp
```diff
@@ -305,8 +312,12 @@ CmdResult CommandFilter::Handle(User* user, const Params& parameters)
...
+
 	if ((filter->flag_no_opers) && user->IsOper())
-		return false;
+		return false;
```
Robby-:
```suggestion
		return false;
```

src/modules/m_filter.cpp
```diff
@@ -305,8 +312,12 @@ CmdResult CommandFilter::Handle(User* user, const Params& parameters)
 
 bool ModuleFilter::AppliesToMe(User* user, FilterResult* filter, int iflags)
 {
+    const AccountExtItem* accountext = GetAccountExtItem();
+
 	if ((filter->flag_no_opers) && user->IsOper())
-		return false;
+		return false;		
+	if ((filter->flag_no_registered) && accountext && accountext->get(user))
+        return false;
```
Robby-:
```suggestion
		return false;
```

src/modules/m_filter.cpp
```diff
@@ -89,6 +91,9 @@ class FilterResult
 				case 'o':
 					flag_no_opers = true;
 				break;
+				case 'r':
```
SaberUK: This should be at the end because otherwise old versions will break when parsing flags.

src/modules/m_filter.cpp
```diff
@@ -121,6 +126,8 @@ class FilterResult
 		std::string flags;
 		if (flag_no_opers)
 			flags.push_back('o');
+		if (flag_no_registered)
```
SaberUK: This should be at the end because otherwise old versions will break when parsing flags.

src/modules/m_filter.cpp
```diff
@@ -305,8 +312,12 @@ CmdResult CommandFilter::Handle(User* user, const Params& parameters)
 
 bool ModuleFilter::AppliesToMe(User* user, FilterResult* filter, int iflags)
 {
+	const AccountExtItem* accountext = GetAccountExtItem();
+
 	if ((filter->flag_no_opers) && user->IsOper())
 		return false;
+	if ((filter->flag_no_registered) && accountext && accountext->get(user))
```
SaberUK: This line of code means:

>If the registered flag is set AND accounts exist AND the user is logged into an account

I think you probably meant to do `!accountext->get(user)`?

Robby-:
Thinking this through, up until now, all flags (except `o` and `c`, but those two don't really matter for the point I'm making) have been about blocking, so you could safely specify a `*` and assume all users their messages would get blocked. With this PR, however, `*` will now behave differently: it will now no longer block messages sent by *registered* users, and so `*` can no longer be assumed to be an all-blocking special flag like it always has been.

Therefore, I think the `*` flag should be removed so that it is no longer a valid character, thus requiring operators to specify exactly the flags they want, reducing the chances of mistakes being made.

src/modules/m_filter.cpp
```diff
@@ -104,8 +106,11 @@ class FilterResult
 				case 'c':
 					flag_strip_color = true;
 				break;
+				case 'r':
+					flag_no_registered = true;
+				break;
 				case '*':
-					flag_no_opers = flag_part_message = flag_quit_message =
+					flag_no_opers = flag_no_registered = flag_part_message = flag_quit_message =
```
SaberUK: This is probably too wide to be included with the default flags. Can you remove it for now we can always add it back later if necessary.

src/modules/m_filter.cpp
H7-25: I've also modified helpop.conf.example and inspircd.conf.example, i hope can be good my english is so bad ...

src/modules/m_filter.cpp
H7-25: I've also modified helpop.conf.example and inspircd.conf.example, i hope can be good my english is so bad ...

PR #1646, opened 2019-05-21
Fix sending large pages in m_httpd

linuxdaemon:
Currently m_httpd writes one buffer's worth of data before closing the socket, this fix makes it write the entire buffer and only close once it finishes.

### DIFF
--- a/include/inspsocket.h
+++ b/include/inspsocket.h
@@ -230,6 +230,8 @@ class CoreExport StreamSocket : public EventHandler
 	/** Whether this socket should close once its sendq is empty */
 	bool closeonempty;
 
+	bool closing;
+
 	/** The IOHook that handles raw I/O for this socket, or NULL */
 	IOHook* iohook;
 
@@ -277,6 +279,7 @@ class CoreExport StreamSocket : public EventHandler
 	const Type type;
 	StreamSocket(Type sstype = SS_UNKNOWN)
 		: closeonempty(false)
+		, closing(false)
 		, iohook(NULL)
 		, type(sstype)
 	{
--- a/include/inspsocket.h
+++ b/include/inspsocket.h
@@ -342,10 +342,8 @@ class CoreExport StreamSocket : public EventHandler
 	 */
 	virtual void Close();
 
-	/**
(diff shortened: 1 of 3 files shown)


SaberUK:
Is it not possible to instead add a flag to `Close()` that makes it block closing until all data has been sent?

linuxdaemon:
> Is it not possible to instead add a flag to `Close()` that makes it block closing until all data has been sent?

As in, have it loop until it sends all the data or just merge `WriteAllClose()` with `Close()`? The former seems non-ideal as it could block the entire process, the latter is possible if that's what you want, yeah.

SaberUK:
>As in, have it loop until it sends all the data or just merge WriteAllClose() with Close()?

Make it so it writes as before your change but then it calls an overload of Close() which returns immediately and only actually closes the connection when all data has been written

linuxdaemon:
> Make it so it writes as before your change but then it calls an overload of Close() which returns immediately and only actually closes the connection when all data has been written.

I think I'm still lost, have `DoWrite()` call the overload or call it in `Page()` ? Or do you mean instead of the `if (closing) return;` logic?

include/inspsocket.h
```diff
@@ -227,6 +227,11 @@ class CoreExport StreamSocket : public EventHandler
 	};
 
  private:
+	/** Whether this socket should close once its sendq is empty */
+	bool closeonempty;
+
+	bool closing;
```
SaberUK: Can you document this?

include/inspsocket.h
```diff
@@ -334,6 +341,10 @@ class CoreExport StreamSocket : public EventHandler
 	 * Close the socket, remove from socket engine, etc
 	 */
 	virtual void Close();
+
+	/** If writeblock == True, only close the socket if all data has been sent */
```
SaberUK:
```suggestion
	/** If writeblock is true then only close the socket if all data has been sent. Otherwise, close immediately. */
```

include/inspsocket.h
linuxdaemon: Fixed in 1cec315078077b55d1417661954a54cee97232c8

PR #1653, opened 2019-06-05
Fix a compiler warning introduced in d107093.

genius3000:
Just an order warning, moved the new initialization line.


SaberUK:

PR #1654, opened 2019-06-05
Allow opers with 'channels/auspex' to see a secret channel topic.

genius3000:
Currently an oper cannot see the topic of a secret channel they aren't in using TOPIC. This was reported on IRC by @KoraggKnightWolf. The priv 'channels/auspex' would imply they should be able to, as LIST will show the topic of such a channel.
Add a check to TOPIC to match the behavior of LIST and meet the expectation.


SaberUK:

PR #1661, opened 2019-06-10
Reset diag pragmas after including external headers

linuxdaemon:
This allows the disabled warnings to still be applied to the rest of the source code while ignoring the vendored `http_parser.c`.

### DIFF
--- a/src/modules/extra/m_ldap.cpp
+++ b/src/modules/extra/m_ldap.cpp
@@ -26,12 +26,20 @@
 #include "inspircd.h"
 #include "modules/ldap.h"
 
+#ifdef __GNUC__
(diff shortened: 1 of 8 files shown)


src/modules/extra/m_ldap.cpp
```diff
@@ -26,13 +26,21 @@
 #include "inspircd.h"
 #include "modules/ldap.h"
 
+#ifdef __GNUC__
```
SaberUK: This can be inside the __APPLE__ ifdef.

linuxdaemon: Resolved in 0a1262828d39dd31eaef41bb1fb4d21fa0965c84

PR #1664, opened 2019-06-14
fix example config for m_connflood

benharri:
seconds is an invalid key


SaberUK:
This should be updated to the v3 name but `seconds` should still work, is this not the case?

benharri:
I wasn't seeing any invalid configuration warnings, but I was using the default value. I've since corrected it in my configs.

SaberUK:

PR #1668, opened 2019-06-22
Replace large if/else blocks for target.type with switches

linuxdaemon:
Switch-cases are a bit easier to read and can be optimized a bit nicer in most cases when doing `if (a.b == value) {} else if (a.b == value2)`.. etc

### DIFF
--- a/src/modules/m_silence.cpp
+++ b/src/modules/m_silence.cpp
@@ -386,6 +386,18 @@ class ModuleSilence
 		SilenceEntry::SilenceFlags flag = SilenceEntry::SF_NONE;
 		switch (target.type)
 		{
+			case MessageTarget::TYPE_CHANNEL:
+			{
+				if (is_ctcp)
+					flag = SilenceEntry::SF_CTCP_CHANNEL;
+				else if (details.type == MSG_NOTICE)
+					flag = SilenceEntry::SF_NOTICE_CHANNEL;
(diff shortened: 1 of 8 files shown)


src/modules/m_silence.cpp
```diff
@@ -384,32 +384,37 @@ class ModuleSilence
...
+					flag = SilenceEntry::SF_CTCP_CHANNEL;
+				else if (details.type == MSG_NOTICE)
+					flag = SilenceEntry::SF_NOTICE_CHANNEL;
+				else if (details.type == MSG_PRIVMSG)
+					flag = SilenceEntry::SF_PRIVMSG_CHANNEL;
+
+				return BuildChannelExempts(user, target.Get<Channel>(), flag, details.exemptions);
+				break;
 			}
+			case MessageTarget::TYPE_SERVER:break;
```
SaberUK:
```suggestion
			case MessageTarget::TYPE_SERVER:
				break;
```

SaberUK:
LGTM but in the cases where you have reordered stuff to be user->channel->server can you revert that as it breaks doing `git blame -w` for no real reason.

linuxdaemon:
> LGTM but in the cases where you have reordered stuff to be user->channel->server can you revert that as it breaks doing `git blame -w` for no real reason.

Fixed in c140f5f2e75f9ef0062707f90c35651ac880fd0b

PR #1676, opened 2019-07-04
Fix whitelist check for a badchan redirect.

genius3000:
Reported by @Cronus89 in #inspircd.dev.

### DIFF
--- a/src/modules/m_denychans.cpp
+++ b/src/modules/m_denychans.cpp
@@ -120,9 +120,13 @@ class ModuleDenyChannels : public Module
 				continue;
 
 			// If the redirect channel is whitelisted then it is okay.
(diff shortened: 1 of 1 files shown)


src/modules/m_denychans.cpp
```diff
@@ -120,9 +120,13 @@ class ModuleDenyChannels : public Module
 				continue;
 
 			// If the redirect channel is whitelisted then it is okay.
-			for (GoodChannels::const_iterator j = goodchans.begin(); j != goodchans.end(); ++j)
+			bool whitelisted = false;
+			for (GoodChannels::const_iterator j = goodchans.begin(); j != goodchans.end() && !whitelisted; ++j)
```
SaberUK: This would be clearer if you used break inside the if statement if something is whitelisted.

PR #1677, opened 2019-07-04
DurationString(): Return "0" for a duration of 0.

genius3000:
Usually a duration of 0 is not allowed or handled separately, but it can also be used as a 'no set time' without separation.
Case in point: m_chanhistory calls DurationString() to convert the max time seconds back to a human readable string for the mode serializer. Returning a blank string is bad here.

### DIFF
--- a/src/helperfuncs.cpp
+++ b/src/helperfuncs.cpp
@@ -433,7 +433,7 @@ bool InspIRCd::IsValidDuration(const std::string& duration)
 std::string InspIRCd::DurationString(time_t duration)
 {
 	if (duration == 0)
-		return "0";
+		return "0s";
 
 	time_t years = duration / 31449600;
 	time_t weeks = (duration / 604800) % 52;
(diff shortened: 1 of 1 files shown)


src/helperfuncs.cpp
```diff
@@ -432,6 +432,9 @@ bool InspIRCd::IsValidDuration(const std::string& duration)
 
 std::string InspIRCd::DurationString(time_t duration)
 {
+	if (duration == 0)
+		return "0";
```
SaberUK:
```suggestion
		return "0s";
```

genius3000: I initially did that, but that looks weird in the current use case. "+H 2:0s" doesn't look as proper as "+H 2:0". "0s" may fit better if/when other functions hit this same scenario though.

SaberUK: I feel that always  showing the seconds with a duration specifier is more user friendly as it allows users who aren't familiar with what the mode does to infer what the parameter is for.

PR #1680, opened 2019-07-08
Added logdetails var (bool) to filter tag

H7-25:
Added logdetails var (bool) to filter tag, enabling it, matching lines are logged to file


SaberUK:
Sorry, I can't accept this pull request. The lack of the message body in filter messages is deliberate as it can be trivially used to covertly spy on users without their consent which is all kinds of illegal.

PR #1689, opened 2019-08-01
Move UNIX socket removal to ListenSocket ctor.

### LINKED ISSUE #1681: Unix sockets not recreated on rehash with remove enabled.
When a Unix socket bind tag has `remove` enabled, a rehash will remove the file but not recreate it. It does work great upon startup if there's a dead socket file left behind.  
I'm not very familiar with socket code but having a look, it removes the file but because the listener still exists in `old_ports`, the `new ListenSocket` isn't called which must be doing the file creation. Do we have a safe way of knowing startup vs rehash? Or should we check the `old_ports` list for the socket before removing?

genius3000:
Doing the removal in BindPorts() would remove the socket during a rehash and not recreate it. Now it's only removed if it's about to be created.

Resolves #1681.

### DIFF
--- a/src/listensocket.cpp
+++ b/src/listensocket.cpp
@@ -29,6 +29,15 @@ ListenSocket::ListenSocket(ConfigTag* tag, const irc::sockets::sockaddrs& bind_t
 	: bind_tag(tag)
 	, bind_sa(bind_to)
 {
+	// Are we creating a UNIX socket?
+	if (bind_to.family() == AF_UNIX)
+	{
+		// Is 'replace' enabled?
+		const bool replace = tag->getBool("replace");
+		if (replace && irc::sockets::isunix(bind_to.str()))
+			unlink(bind_to.str().c_str());
+	}
+
 	fd = socket(bind_to.family(), SOCK_STREAM, 0);
 
 	if (this->fd == -1)
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -118,10 +118,6 @@ int InspIRCd::BindPorts(FailedPortList& failed_ports)
 				continue;
 			}
 
-			const bool replace = tag->getBool("replace");
-			if (replace && irc::sockets::isunix(fullpath))
-				remove(fullpath.c_str());
-
 			irc::sockets::untosa(fullpath, bindspec);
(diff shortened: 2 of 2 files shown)


src/listensocket.cpp
```diff
@@ -29,6 +29,18 @@ ListenSocket::ListenSocket(ConfigTag* tag, const irc::sockets::sockaddrs& bind_t
...
+	// Are we creating a UNIX socket?
+	const std::string path = tag->getString("path");
+	if (!path.empty())
+	{
+		// Expand the path relative to the config directory.
+		const std::string fullpath = ServerInstance->Config->Paths.PrependData(path);
+		// Is 'replace' enabled?
+		const bool replace = tag->getBool("replace");
+		if (replace && irc::sockets::isunix(fullpath))
+			remove(fullpath.c_str());
```
SaberUK:
```suggestion
			unlink(fullpath.c_str());
```

(This was a bug in the original code).

src/listensocket.cpp
```diff
@@ -29,6 +29,18 @@ ListenSocket::ListenSocket(ConfigTag* tag, const irc::sockets::sockaddrs& bind_t
 	: bind_tag(tag)
 	, bind_sa(bind_to)
 {
+	// Are we creating a UNIX socket?
+	const std::string path = tag->getString("path");
```
SaberUK: You can check if `bind_sa.family()` is `AF_UNIX` here.

src/listensocket.cpp
```diff
@@ -29,6 +29,18 @@ ListenSocket::ListenSocket(ConfigTag* tag, const irc::sockets::sockaddrs& bind_t
...
+	if (!path.empty())
+	{
+		// Expand the path relative to the config directory.
+		const std::string fullpath = ServerInstance->Config->Paths.PrependData(path);
+		// Is 'replace' enabled?
+		const bool replace = tag->getBool("replace");
+		if (replace && irc::sockets::isunix(fullpath))
```
SaberUK: `fullpath` can just be `bind_sa.str()` I think?

genius3000: Yeah that works successfully.

PR #1698, opened 2019-09-02
Fixing SILENCE command

iwalkalone69:
When silence mask is prefixed by + or -, it should only remove the first character, not the entire string.


SaberUK:
Nice catch, thanks! I'll merge this once Travis is done.

Issue #1674, opened 2019-06-29
[Feature] CHMODE +g could block part/quit messages


PeGaSuS-Coder:
When we use `CHMODE +g` to prevent some "bad words" from being sent to channel, it could probably also block part/quit messages containing those words.
Some users find that they're unable to send those words to the channel (spam, insults, swearing, etc) and they use them in the part/quit message to achieve the their goal.
IMHO, this would be a good enhancement and very helpful for channel owners, if we think in channels where different languages are spoken at the same time and network admins wouldn't need to add every single word to a spamfilter, e.g.

KoraggKnightWolf:
Definitely a good idea, my guess would be is it hooks into PRIVMSG and NOTICE to act upon those, and could perhaps thus also be changed to hook into PART and QUIT as well.

SadieCat:
This was implemented for part messages, we can't really do it for quit messages because those are global not channel-local.

PR #1702, opened 2019-09-14
m_chanfilter: Apply filters to part messages

linuxdaemon:
Partial implementation for #1674

### DIFF
--- a/src/modules/m_chanfilter.cpp
+++ b/src/modules/m_chanfilter.cpp
@@ -110,7 +110,7 @@ class ModuleChanFilter : public Module
 		LocalUser* luser = IS_LOCAL(user);
 
 		std::string oldreason(partmessage);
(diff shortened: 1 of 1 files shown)


src/modules/m_chanfilter.cpp
```diff
@@ -76,40 +95,62 @@ class ModuleChanFilter : public Module
...
+		Channel* chan = memb->chan;
+		ChanFilter::ListItem* match = Match(user, chan, partmessage);
+		if (!match)
+			return;
+
+		// Match() checks the user is local, we can assume from here
+		LocalUser* luser = IS_LOCAL(user);
+
+		std::string oldreason(partmessage);
+		partmessage = "Filtered";
```
SaberUK: In order to match the filter module:

```suggestion
		partmessage = "Reason filtered";
```

PR #1707, opened 2019-09-18
m_noctcp: Fix noctcp user mode not applying to exempted channel ops.

Robby-:


SaberUK:

PR #1708, opened 2019-09-21
Update examples directory when paths are changed interactively.

genius3000:
Currently when you change the install paths interactively in ./configure, the example path remains set to the default. This results in the example files "hiding" in the wrong place. Since this option is more for system maintainers and scripted setups, I've just added a line to update the examples directory during the interactive path changes rather than give the option to define it there.
A quick test on Ubuntu 16.04 showed successful.

Reported by damian on IRC.


SaberUK:

PR #1732, opened 2019-11-15
Improve handling of a link block without a port.

genius3000:
## Summary
This removes the 'no port' check/skip from the allowed IPs list refresh function. Now a link block with no port defined will have it's `allowmask` values and validated `ipaddr` value added to the allowed IPs list. The 'no port' check is now placed in the /connect command handler to prevent connect attempts from even trying.
### ToDo
- Check/modify autoconnects to ensure they won't try connecting to a no port link either.

## Rationale
- Links like services and some relays only connect to an uplink and don't accept incoming connections. There is no need to define a port or allow /connect to these links.

## Testing Environment

**Operating system name and version:** Ubuntu 16.04
**Compiler name and version:** GCC 5.4.0

## Checks
I have ensured that:

  - [ ] No issues arise from having '0' used as the port in the ipaddr validation.

### DIFF
--- a/src/modules/m_spanningtree/main.cpp
+++ b/src/modules/m_spanningtree/main.cpp
@@ -336,6 +336,12 @@ ModResult ModuleSpanningTree::HandleConnect(const CommandBase::Params& parameter
 				return MOD_RES_DENY;
 			}
 
+			if (!x->Port)
+			{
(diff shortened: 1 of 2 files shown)


genius3000:
Going to close this for now. This would be nice/minor improvement but playing well with everything (incl. Unix Domain Sockets) requires a little more work. Will open a new PR later.

PR #1747, opened 2020-01-12
Adding bool options notify and notify_onlyops

iwalkalone69:
## Summary

Adding boolean options notify and notify_onlyops to m_joinflood to setup how it is notified to channel when protection is triggered.

Option notify enables or disables notification. Option notify_onlyops only sends to channel operators (@)  if enabled.

## Rationale

This is done to allow avoiding normal users to see it as it can be annoying when having a network with more than few servers.

## Testing Environment

Used branch insp3, connected 2 users to a channel (only one with op) and tested with 3 options:

- notify="false":  no one receives channel notice
- notify="true" and notify_onlyops="false": everyone receives channel notice
- notify="true" and notify_onlyops="true": only the user with op receives channel notice

**Operating system name and version:** 5.3.0-26-generic
**Compiler name and version:** GCC 9.2.1

## Checks

I have ensured that:

  - [ x ] If ABI changes have been made I have incremented INSPIRCD_VERSION_API.

### DIFF
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -1232,8 +1232,13 @@
 # Closes the channel for N seconds if X users join in Y seconds.
 #<module name="joinflood">
 #
-# The number of seconds to close the channel for:
-#<joinflood duration="1m">
+# This tag takes the following attributes:
+#
+# notify         - If enabled, it will send a notice to channel when
+#                  protection is triggered.
+# notify_onlyops - If enabled, only notify to channel operators.
+#
+#<joinflood duration="1m" notify="true" notify_onlyops="false">
 
 #-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#
 # Anti auto rejoin: Adds support for prevention of auto-rejoin (+J).
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -1234,7 +1234,7 @@
 #
 # This tag takes the following attributes:
 #
-# duration       - The number of seconds to close the channel.
+# duration       - Time to close the channel when protection is triggered.
 # notify         - If enabled, it will send a notice to channel when
 #                  protection is triggered.
 # notify_onlyops - If enabled, only notify to channel operators.
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -1234,6 +1234,7 @@
 #
 # This tag takes the following attributes:
 #
+# duration       - The number of seconds to close the channel.
 # notify         - If enabled, it will send a notice to channel when
 #                  protection is triggered.
 # notify_onlyops - If enabled, only notify to channel operators.
--- a/src/modules/m_joinflood.cpp
+++ b/src/modules/m_joinflood.cpp
@@ -34,6 +34,12 @@ enum
 // The number of seconds the channel will be closed for.
 static unsigned int duration;
 
+// Enable notice to channel when it is triggered
+bool notify;
+
+// Only notify to channel operators
+bool notify_onlyops;
+
 /** Holds settings and state associated with channel mode +j
  */
 class joinfloodsettings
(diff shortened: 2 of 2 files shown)


docs/conf/modules.conf.example
```diff
@@ -1232,8 +1232,13 @@
 # Closes the channel for N seconds if X users join in Y seconds.
 #<module name="joinflood">
 #
-# The number of seconds to close the channel for:
```
genius3000: You should retain the description of 'duration'. Probably best to reword it though, as it can be a time string not just number of seconds.

src/modules/m_joinflood.cpp
```diff
@@ -177,7 +185,12 @@ class ModuleJoinFlood : public Module
 			{
 				f->clear();
 				f->lock();
-				memb->chan->WriteNotice(InspIRCd::Format("This channel has been closed to new users for %u seconds because there have been more than %d joins in %d seconds.", duration, f->joins, f->secs));
+				if (notify) {
+					std::string message = InspIRCd::Format("This channel has been closed to new users for %u seconds because there have been more than %d joins in %d seconds.", duration, f->joins, f->secs);
```
genius3000: I believe Channel::WriteNotice() allows for an optional status character.

docs/conf/modules.conf.example
iwalkalone69: Fixed

src/modules/m_joinflood.cpp
iwalkalone69: You're right about the optional parameter. Changed to simplify the code.

docs/conf/modules.conf.example
```diff
@@ -1232,8 +1232,14 @@
...
-#<joinflood duration="1m">
+# This tag takes the following attributes:
+#
+# duration       - Time to close the channel when protection is triggered.
+# notify         - If enabled, it will send a notice to channel when
```
genius3000, line 1238:
```suggestion
# notify         - If enabled, it will send a notice to the channel when
```

docs/conf/modules.conf.example
```diff
@@ -1232,8 +1232,14 @@
...
+#                  protection is triggered.
+# notify_onlyops - If enabled, only notify to channel operators.
```
genius3000, line 1240:
```suggestion
# notify_onlyops - If enabled, only notify to the channel operators.
```

src/modules/m_joinflood.cpp
```diff
@@ -34,6 +34,12 @@ enum
 // The number of seconds the channel will be closed for.
 static unsigned int duration;
 
+// Enable notice to channel when it is triggered
```
genius3000, line 37:
```suggestion
// Enable notice to the channel when it is triggered.
```

src/modules/m_joinflood.cpp
```diff
@@ -34,6 +34,12 @@ enum
...
+bool notify;
+
+// Only notify to channel operators
```
genius3000, line 40:
```suggestion
// Only notify to the channel operators.
```

src/modules/m_joinflood.cpp
```diff
@@ -177,7 +185,9 @@ class ModuleJoinFlood : public Module
 			{
 				f->clear();
 				f->lock();
-				memb->chan->WriteNotice(InspIRCd::Format("This channel has been closed to new users for %u seconds because there have been more than %d joins in %d seconds.", duration, f->joins, f->secs));
+				if (notify) {
+					memb->chan->WriteNotice(InspIRCd::Format("This channel has been closed to new users for %u seconds because there have been more than %d joins in %d seconds.", duration, f->joins, f->secs), notify_onlyops ? '@' : 0);
+				}
```
genius3000, line 190:
```suggestion
				if (notify)
					memb->chan->WriteNotice(InspIRCd::Format("This channel has been closed to new users for %u seconds because there have been more than %d joins in %d seconds.", duration, f->joins, f->secs), notify_onlyops ? '@' : 0);
```

genius3000 (changes requested):
A couple more wording/styling nitpicks.

PR #1793, opened 2020-08-01
configure: When not generating certs, warn if certs exist in source directory that will overwrite output

digitalcircuit:
## Summary

When running `./configure` and not enabling certificate generation, warn if certificates exist in the root of the source directory which will overwrite any certificates in the build output.

## Rationale

This helps avoid the issue where someone initially generates self-signed certificates for testing, then later replaces the certs in
the output directory without removing them from the source directory.  Previously, it wasn't clear why the certificates were being replaced despite answering "no".

## Testing Environment

**Operating system name and version:** Kubuntu 20.04 LTS
**Compiler name and version:** GCC (`Ubuntu 9.3.0-10ubuntu2`) `9.3.0`
**Perl version:** Perl `v5.30.0`

The output after having previously run `./configure` and generating self-signed certs looks like this…

```
[...]
Would you like to generate a self-signed SSL certificate now? This certificate
can be used for testing but should not be used on a production network.

Note: you can get a free CA-signed certificate from Let's Encrypt. See
https://letsencrypt.org/getting-started/ for more details.

[yes] => no

Warning: InspIRCd will use the existing generated certificates in the root of
the source directory.  If this is not intended, remove the .pem files before
running make.

Writing .configure/cache.cfg ...
[...]
```

## Checks

I have ensured that:

### DIFF
--- a/configure
+++ b/configure
@@ -394,6 +394,13 @@ EOQ
 if (<$RealDir/src/modules/m_ssl_*.cpp>) {
 	if (prompt_bool $interactive, $question, $interactive) {
 		system './tools/genssl', 'auto';
+	} elsif (<$RealDir/*.pem>) {
+		# Answered "no", and (likely self-signed) certs already exist
+		print_warning <<"EOW";
+InspIRCd will use the existing generated certificates in the root of
+the source directory.  If this is not intended, remove the .pem files before
+running make.
+EOW
 	}
 } else {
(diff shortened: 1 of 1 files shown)


digitalcircuit:
I'm happy to change the phrasing, how the warning works (e.g. informational message instead), or anything else.  I just wanted to get this up for discussion after it having caught myself and another off-guard.

PR #1799, opened 2020-10-02
Add m_argon2 to extra

### LINKED ISSUE #1540: Argon2 Support
Please add Support for Argon2, Argon2i and Argon2id

ShutterQuick:
Resolves #1540

## Summary
Adds a module for the Argon2 KDF.

## Testing Environment

WSL 4.4.0-19041-Microsoft

clang version 10.0.0-4ubuntu1
Target: x86_64-pc-linux-gnu

## Checks

I have ensured that:

  - [X] If ABI changes have been made I have incremented MODULE_ABI in `moduledefs.h`.

### DIFF
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -187,6 +187,30 @@
 #           uppercase="ABCDEFGHIJKLMNOPQRSTUVWXYZ">
 
 #-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#
+# Argon2 module: Allows other modules to generate Argon2 hashes,
+# usually for cryptographic uses and security.
+# This module makes the algorithms argon2i, argon2d and argon2id
+# available for use.
+# Note that this module is extra, and must be enabled explicitly
+# to build. It depends on libargon2.
+#<module name="argon2">
+#
+# memory: Memory hardness, in KiB. E.g. 131072 KiB = 128 MiB.
+# iterations: Time hardness in iterations. (def. 3)
+# lanes: How many parallel chains can be run. (def. 1)
+# threads: Maximum amount of threads each invokation can spawn. (def. 1)
+# length: Output length in bytes. (def. 32)
+# saltlength: Salt length in bytes. (def. 16)
+# version: Algorithm version, 10 or 13. (def. 13)
+# The parameters can be customized as follows:
+#<argon2 iterations="3" memory="131074" length="32" saltlength="16">
+# Defines the parameters that are common for all the variants (i/d/id).
+# Can be overriden on individual basis, e.g.
+#<argon2i iterations="4">
+#<argon2d memory="131074"
+#<argon2id iterations="5" memory="262144" length="64" saltlength="32">
+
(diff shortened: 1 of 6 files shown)


docs/conf/modules.conf.example
```diff
@@ -1665,6 +1665,30 @@
...
 # Don't run it on a server you don't trust with your password.
 
+#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#-#
+# Argon2 module: Allows other modules to generate Argon2 hashes,
+# usually for cryptographic uses and security.
+# This module makes the algorithms argon2i, argon2d and argon2id
+# available for use.
+# Note that this module is extra, and must be enabled explicitly
+# to build. It depends on libargon2.
+#<module name="argon2">
```
SadieCat, line 1675: This module begins with "a" so it should be higher up in the config (probably between anticaps and auditorium).

src/modules/extra/m_argon2.cpp
```diff
@@ -0,0 +1,199 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/// $LinkerFlags: find_linker_flags("libargon2" "-llibargon2")
```
SadieCat, line 19: Two spaces after this. Also, you should have a $CompilerFlags just in case its not in the default include path.

src/modules/extra/m_argon2.cpp
```diff
@@ -0,0 +1,199 @@
...
+                return provider;
+            }
+        }
+
+        // We know exactly what providers are loaded. We should always find what
+        // we're looking for.
+        throw ModuleException("Unable to find '" + name + "'. This is a bug.");
+    }
+
+    void ReadConfig(ConfigTag* tag, ProviderConfig* config)
```
SadieCat, line 186: This config can be in the HashProvider then you can just call hashprovider.ReadConfig("argon2x") and get rid of the ProviderConfig class.

src/modules/extra/m_argon2.cpp
```diff
@@ -0,0 +1,199 @@
...
+        for (auto& config : configs)
+        {
+            auto provider = FindProvider(config.first);
+            provider->Configure(config.second);
+        }
+	}
+
+    HashArgon2* FindProvider(const std::string& name)
+    {
+        for (auto provider : providers)
```
SadieCat, line 172: Note: v2 is still C++03+TR1 so C++11 constructs can't be used yet.

SadieCat:
Can you also add a pkg-config auto-enable line to ./configure (grep the file for gnutls and see what it does there) and also add it to the CI? Thanks.

src/modules/extra/m_argon2.cpp
ShutterQuick: I moved the logic and simplified it.
Didn't kill HashProvider though. I used it to hold the values from the argon2 block so it can be used as default values for the other config blocks (argon2i, argon2d, argon2id).

PR #1808, opened 2020-10-23
m_cban: Implement support for channel masks

Techman:
## Summary

This PR updates m_cban to use InspIRCd::Match to enable glob matching.

## Rationale

The existing logic does not actually account for a globbed channel name.

## Testing Environment

Tested on the insp3 git as of commit d0c1741.

**Operating system name and version:** Linux 5.8.16
**Compiler name and version:** Clang 10.0.1

## Checks

I have ensured that:

### DIFF
--- a/docs/conf/helpop.conf.example
+++ b/docs/conf/helpop.conf.example
@@ -572,9 +572,9 @@ the network.
 Sends a message to all users with the +g snomask.
 ">
 
-<helpop key="cban" title="/CBAN <channel> [<duration> [:<reason>]]" value="
-Sets or removes a global channel ban. You must specify all three parameters
-to add a ban, and one parameter to remove a ban (just the channel).
+<helpop key="cban" title="/CBAN <channelmask> [<duration> [:<reason>]]" value="
+Sets or removes a global channel based ban. You must specify all three parameters
+to add a ban, and one parameter to remove a ban (just the channelmask).
 
 The duration may be specified in seconds, or in the format
 1y2w3d4h5m6s - meaning one year, two weeks, three days, four hours,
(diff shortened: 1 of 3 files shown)


SadieCat:
This is a good change but as-is it breaks compatibility with existing servers as cbans may be valid on one side of the link but not the other.

Can you add a config option to make it match using glob and then if that is set expose it in the last parameter of `Version()`?

Techman:
Good note on the compatibility thing. However, the InspIRCd 3 documentation already listed this module as supporting glob matching. In a way, you could look at this as a bug since the module was not conforming to the documented behavior. I didn't dig too far back to see why the documentation changed.

The config option doesn't seem too hard to add (though admittedly I'll have to brush up that Version stuff), but with the pace that InspIRCd 3 moves, improvements are made often. I know that this would introduce an issue with InspIRCd 2 servers, but that is making its way out the door soon in terms of remaining support.

---

I'll provide some background for what I was planning to do here. Depending on if this is merged, I wanted to then go to Anope and provided a PR to enable the ability to SQLine channels. Anope works with glob, so the IRCd end would need to be able expect that.

SadieCat:
>. However, the InspIRCd 3 documentation already listed this module as supporting glob matching. In a way, you could look at this as a bug since the module was not conforming to the documented behavior. I didn't dig too far back to see why the documentation changed.

The v2 docs also mention that it uses a glob pattern. I'd imagine this is a mistake I made when writing the new docs.

>The config option doesn't seem too hard to add (though admittedly I'll have to brush up that Version stuff)

It shouldn't be too hard. If you just add a bool option to the class and then as the third parameter to Version you can just do something like this:

```cpp
return Version("Adds the /CBAN command which allows server operators to prevent channels matching a glob from being created.", VF_COMMON | VF_VENDOR, glob ? "glob" : "");
```


>I know that this would introduce an issue with InspIRCd 2 servers, but that is making its way out the door soon in terms of remaining support.

It would also conflict with older v3 servers which is problematic.

PR #1819, opened 2020-11-29
Document that exemptions configured in <options:exemptchanops> can be negated, and change the <badnick> ChanServ example

Robby-:
## Summary

- Document that exemptions configured in \<options:exemptchanops\> can be negated.
- Change the \<badnick\> example in inspircd.conf.example to another nick, since ChanServ is already covered by examples/services/generic.conf.example.

## Rationale

Because.

## Testing Environment

Not applicable.

**Operating system name and version:** Not applicable.
**Compiler name and version:** Not applicable.

## Checks

I have ensured that:

### DIFF
--- a/docs/conf/inspircd.conf.example
+++ b/docs/conf/inspircd.conf.example
@@ -685,8 +685,9 @@
          #  - stripcolor      Channel mode +S - strips formatting codes from
          #                    messages (requires the stripcolor module).
          #  - topiclock       Channel mode +t - limits changing the topic to (half)ops
-         # You can also configure this on a per-channel basis with a channel mode.
-         # See m_exemptchanops in modules.conf.example for more details.
+         # You can also configure this on a per-channel basis with a channel mode and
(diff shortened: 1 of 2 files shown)


docs/conf/inspircd.conf.example
```diff
@@ -962,10 +963,10 @@
 
 <badnick
          # nick: Nick to disallow. Wildcards are supported.
-         nick="ChanServ"
+         nick="Admin"
```
SadieCat, line 966: Can you use a nick here that it's likely that people won't use? We've had enough trouble caused by the `root@*` example...

docs/conf/inspircd.conf.example
```diff
@@ -962,10 +963,10 @@
...
 
          # reason: Reason to display on /NICK.
-         reason="Reserved for a network service">
+         reason="Reserved for staff">
```
SadieCat, line 969: Opers can't use elined nicks so this is misleading.

PR #1836, opened 2021-01-18
Added <module name="codepage"> to modules.conf.example

R-V6:


SadieCat:
Not sure how that got missed, thanks.

PR #1837, opened 2021-01-18
Implemented customized messages for m_joinflood

iwalkalone69:
## Summary

Be able tu customize m_joinflood messages in configuration.

## Rationale

To allow customizing module messages or translate it to another language.

## Testing Environment

Latest insp branch.

**Operating system name and version:** Debian 10.6
**Compiler name and version:** gcc 8.3.0

## Checks

I have ensured that:

  - [ x ] If ABI changes have been made I have incremented MODULE_ABI in `moduledefs.h`.


docs/conf/modules.conf.example
```diff
@@ -1296,20 +1296,27 @@
...
+#                being throttled by joinflood.
 #
-# splitwait: The number of seconds to disengage joinflood for after
-#            a server splits. This allows users to reconnect without
-#            being throttled by joinflood.
+# splitwait:     The number of seconds to disengage joinflood for after
+#                a server splits. This allows users to reconnect without
+#                being throttled by joinflood.
+#
+# denymessage:   Message to display to the user when join is denied.
```
genius3000, line 1310: Nitpick:
```suggestion
# denymessage:   Message to display to the user when a join is denied.
```

src/modules/m_joinflood.cpp
```diff
@@ -159,6 +161,8 @@ class ModuleJoinFlood
 		duration = tag->getDuration("duration", 60, 10, 600);
 		bootwait = tag->getDuration("bootwait", 30);
 		splitwait = tag->getDuration("splitwait", 30);
+		denymessage = tag->getString("denymessage", "This channel is temporarily unavailable (+j is set). Please try again later.");
```
genius3000, line 164: We should enforce a non-blank message here or the server could, if configured to, send a numeric with no message.
```suggestion
		denymessage = tag->getString("denymessage", "This channel is temporarily unavailable (+j is set). Please try again later.", 1);
```

src/modules/m_joinflood.cpp
```diff
@@ -199,7 +203,7 @@ class ModuleJoinFlood
 			{
 				f->clear();
 				f->lock();
-				memb->chan->WriteNotice(InspIRCd::Format("This channel has been closed to new users for %u seconds because there have been more than %d joins in %d seconds.", duration, f->joins, f->secs));
+				memb->chan->WriteNotice(InspIRCd::Format(notifymessage.c_str(), duration, f->joins, f->secs));
```
genius3000, line 206: This could break pretty easily by changing or removing the format specifiers in the configuration, that's not good. For now, I would recommend removing the changes regarding this message and just keep the changes for the user who tried to join.

genius3000 (changes requested):
I left a few minor suggestions and I'll elaborate more on the removal of any changes to the channel notice message. As far as translation purposes go, there's **many** more places that will benefit from the ability when it is implemented. I don't think exposing strings with format specifiers to the configuration is a good way of doing this though.

src/modules/m_joinflood.cpp
SadieCat: (This is why I haven't merged it yet fwiw.)

iwalkalone69: Removed notifymessage in code and docs.

iwalkalone69:
Sorry, I did a mess. I close and I start a new PR with another branch synced with insp3.

Issue #1414, opened 2017-11-17
Issues with signedness and and problematic shortening of integers.


SadieCat:
**DISCLAIMER**: I do not think that any security issues are caused by this. If you have evidence to the contrary then please PM me on IRC rather than posting it publicly.

---

There are a bunch of issues with the way we handle integers internally. I have fixed a bunch of them in 36040be2952186d56a6646ee7d972aaafdd4e31a but there are still quite a few more left.

Ideally we should enable -Wsign-conversion and -Wshorten-64-to-32 (Clang) by default to help catch these issues.

A few more things that should probably be fixed at the same time:

- [x] Ports should use sa_port_t (aka uint16_t) instead of int/unsigned int.
- [x] Socket engine functions should return ssize_t (usually int but there are no guarantees that `sizeof(ssize_t)` will be `sizeof(int)`) instead of int.
- [x] getDuration should be converted to return an unsigned long rather than a long (this requires some internal config tag fiddling to allow CheckRange to check unsigned values).
- [x] getInt needs an analogue for unsigned integers (currently we use getInt and hope stuff does not overflow)
- [ ] Mode stuff uses unsigned char when the chars inside of a std::string are of undefined signing which can potentially cause problems when iterating across a string of modes.

SaberUK:
getDuration now returns unsigned long (2d36fcb16ef3209fffbe9f4b600971a1edd2ae4b).

getInt now has an analogue for unsigned integer (780dda83ba3857bc3c330d4b5d38bb251a9d879b).

SaberUK:
A bunch of this has been fixed in 0f7cfd46ef2d277f5f82e34a2852c75212d75261.

dominichamon:
I was playing around to see if i could clean some of these up and I found `ConfigTag::getUInt` returns an `unsigned long`. The name suggests it shouldn't (and the return value is often being written to an `unsigned int`) but it uses `strtoul` in the implementation, which suggests `unsigned long` is the correct return value.

Can I assume it is meant to return an `unsigned long` and update the callsites, and should I consider renaming the function too?

SadieCat:
The function name doesn't really matter. getInt returns a signed long too.

dominichamon:
Do you prefer one big PR or to focus on one API at a time? I have a poc and it seems like most fixes are going to be uncontroversial (I think!).

SadieCat:
Providing you don't break API compatibility in the stable branch one pull request is fine.

[...]

PR #1875, opened 2021-05-27
Added -Wshorten-64-to-32 and fixed all warnings

dominichamon:
## Summary

This pull request adds `-Wshorten-64-to-32` to the compilation flags (when supported... or at least non-ICC and non-GCC compilers... let me know if there's a better way to check for clang). It also fixes all existing warnings, identified by adding `-Werror` locally.

## Rationale

Part of fixing #1414. There shouldn't be any security issues, but there are some potentially unsafe casts identified which could require closer scrutiny. 

## Testing Environment

**Operating system name and version:** Linux 5.8.0
**Compiler name and version:** clang 11.0.0-2

## Checks

I have ensured that:

  - [n/a ] I have documented any features added by this pull request.

I think this is accurate. There are no new features, and all API changes are changes from one integer type to another, which should be compatible with current usages. I'm happy to update MODULE_ABI (and rebase to master) if necessary, obviously.

I'm also happy to split the pull request up into parts (somehow!) if you want.

### DIFF
--- a/src/coremods/core_lusers.cpp
+++ b/src/coremods/core_lusers.cpp
@@ -28,8 +28,8 @@
 
 struct LusersCounters
 {
-	unsigned int max_local;
-	unsigned int max_global;
+	unsigned long max_local;
+	unsigned long max_global;
 	unsigned int invisible;
 
 	LusersCounters(unsigned int inv)
@@ -41,7 +41,7 @@ struct LusersCounters
 
 	inline void UpdateMaxUsers()
 	{
-		unsigned int current = ServerInstance->Users->LocalUserCount();
+		unsigned long current = ServerInstance->Users->LocalUserCount();
 		if (current > max_local)
 			max_local = current;
 
@@ -74,10 +74,10 @@ class CommandLusers : public Command
  */
 CmdResult CommandLusers::Handle(User* user, const Params& parameters)
 {
-	unsigned int n_users = ServerInstance->Users->RegisteredUserCount();
+	size_t n_users = ServerInstance->Users->RegisteredUserCount();
 	ProtocolInterface::ServerList serverlist;
 	ServerInstance->PI->GetServerList(serverlist);
-	unsigned int n_serv = serverlist.size();
+	size_t n_serv = serverlist.size();
 	unsigned int n_local_servs = 0;
 	for (ProtocolInterface::ServerList::const_iterator i = serverlist.begin(); i != serverlist.end(); ++i)
 	{
@@ -90,7 +90,7 @@ CmdResult CommandLusers::Handle(User* user, const Params& parameters)
 
 	counters.UpdateMaxUsers();
 
-	user->WriteNumeric(RPL_LUSERCLIENT, InspIRCd::Format("There are %d users and %d invisible on %d servers",
+	user->WriteNumeric(RPL_LUSERCLIENT, InspIRCd::Format("There are %lu users and %d invisible on %zu servers",
 			n_users - counters.invisible, counters.invisible, n_serv));
 
 	if (ServerInstance->Users->OperCount())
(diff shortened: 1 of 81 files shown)


src/coremods/core_lusers.cpp
```diff
@@ -90,7 +90,7 @@ CmdResult CommandLusers::Handle(User* user, const Params& parameters)
 
 	counters.UpdateMaxUsers();
 
-	user->WriteNumeric(RPL_LUSERCLIENT, InspIRCd::Format("There are %d users and %d invisible on %d servers",
+	user->WriteNumeric(RPL_LUSERCLIENT, InspIRCd::Format("There are %lu users and %d invisible on %zu servers",
```
SadieCat, line 93: `%zu` is not available in the v3 branch as it targets C++98 w/ TR1. This should go into master instead.

src/coremods/core_user/cmd_userhost.cpp
```diff
@@ -32,11 +32,12 @@ CmdResult CommandUserhost::Handle(User* user, const Params& parameters)
 
 	std::string retbuf;
 
-	unsigned int max = parameters.size();
+	size_t max = parameters.size();
 	if (max > 5)
 		max = 5;
 
-	for (unsigned int i = 0; i < max; i++)
+  // This cast is safe thanks to the above clamp.
```
SadieCat, line 39: Broken indentation.

src/inspsocket.cpp
```diff
@@ -180,42 +180,46 @@ void StreamSocket::DoRead()
...
-			SocketEngine::ChangeEventMask(this, FD_WANT_FAST_READ | FD_ADD_TRIAL_READ);
-			return 0;
-		}
-		else
-		{
-			error = SocketEngine::LastError();
-			SocketEngine::ChangeEventMask(this, FD_WANT_NO_READ | FD_WANT_NO_WRITE);
-			return -1;
-		}
+	}
```
SadieCat, line 206: Broken indentation.

src/listensocket.cpp
```diff
@@ -86,19 +86,22 @@ ListenSocket::ListenSocket(ConfigTag* tag, const irc::sockets::sockaddrs& bind_t
...
+	}
 
 	if (bind_to.family() == AF_UNIX)
 	{
 		const std::string permissionstr = tag->getString("permissions");
-		unsigned int permissions = strtoul(permissionstr.c_str(), NULL, 8);
+		unsigned long permissions = strtoul(permissionstr.c_str(), NULL, 8);
 		if (permissions && permissions <= 07777)
-			chmod(bind_to.str().c_str(), permissions);
+			chmod(bind_to.str().c_str(), static_cast<__mode_t>(permissions));
```
SadieCat, line 99: What is `__mode_t`? It looks like a platform-specific extension? `man 2 chmod` refers to `mode_t` which is fine but it might need a typedef in the Windows compat header.

src/modules/extra/m_regex_pcre.cpp
```diff
@@ -65,7 +65,8 @@ class PCRERegex : public Regex
 
 	bool Matches(const std::string& text) CXX11_OVERRIDE
 	{
-		return (pcre_exec(regex, NULL, text.c_str(), text.length(), 0, 0, NULL, 0) >= 0);
+    // Possibly unsafe cast here, but pcre_exec expects an int.
```
SadieCat, line 68: Broken indentation.

src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -774,7 +776,8 @@ class OpenSSLIOHook : public SSLIOHook
 		{
 			ERR_clear_error();
 			char* buffer = ServerInstance->GetReadBuffer();
-			size_t bufsiz = ServerInstance->Config->NetBufferSize;
+      // This cast may be unsafe but an int is expected by SSL_read.
```
SadieCat, line 779: Broken indentation.

src/modules/m_ircv3_batch.cpp
```diff
@@ -150,7 +150,8 @@ class IRCv3::Batch::ManagerImpl : public Manager
 		if (id >= MAX_BATCHES)
 			return;
 
-		batch.Setup(id);
+    // This cast is safe thanks to the clamping above.
```
SadieCat, line 153: Broken indentation.

src/inspsocket.cpp
dominichamon: i think this one is ok. i've added an extra clause before casting away signedness above so there is one more level of indentation expected.

src/coremods/core_lusers.cpp
dominichamon: i think this change is required once we change the type of the parameter, so maybe the whole thing should be rebased on master? might be better given the scope of the change too. wdyt?

src/listensocket.cpp
dominichamon: Replaced with `int` which is safe thanks to the check.

src/coremods/core_lusers.cpp
SadieCat: It's probably fine to just cast it to unsigned long for v3 and change it for master. 4,294,967.295 is more users than any IRC server has ever seen and I doubt that'll be changing soon.

You can rebase against master if you want though. It'll probably be safer to do that.

src/inspsocket.cpp
SadieCat: Ah, I see!

src/coremods/core_lusers.cpp
dominichamon: i'll do that. i'll close this and open a new one as trying to fix everything up to base on master is likely going to be hard.

PR #1876, opened 2021-05-27
Fix shorten 64 to 32

dominichamon:
## Summary

This pull request adds `-Wshorten-64-to-32` to the compilation flags (when supported... or at least non-ICC and non-GCC compilers... let me know if there's a better way to check for clang). It also fixes all existing warnings, identified by adding `-Werror` locally.

## Rationale

Part of fixing #1414. There shouldn't be any security issues, but there are some potentially unsafe casts identified which could require closer scrutiny. 

## Testing Environment

**Operating system name and version:** Linux 5.8.0
**Compiler name and version:** clang 11.0.0-2

## Checks

I have ensured that:

  - [ ] This pull request does not introduce any incompatible API changes.

I think this is accurate. There are no new features, and all API changes are changes from one integer type to another, which should be compatible with current usages. I'm happy to update MODULE_ABI if necessary, obviously.

I'm also happy to split the pull request up into parts (somehow!) if you want.


SadieCat:
The CI failures are due to a GitHub Actions issue. I have a probable fix which I'll test and push shortly.

SadieCat:
I have pushed a fix for the GitHub Actions CI issue. Can you rebase?

dominichamon:
> I have pushed a fix for the GitHub Actions CI issue. Can you rebase?

Done. had to force push for some reason, sorry.

dominichamon:
looks like my attempt to only add `-Wshorten-64-to-32` for clang also added it to gcc. i'll take a look at that.

dominichamon:
reconfiguring my local build to include _all_ modules and running some more local tests. sorry about that.

SadieCat:
No worries. You can use `./tools/test-build` to test multiple socket engines and compilers at once if that helps.

dominichamon:
> No worries. You can use `./tools/test-build` to test multiple socket engines and compilers at once if that helps.

it will certainly help reduce the amount you need to trigger workflows (i hope).

src/configreader.cpp
```diff
@@ -347,7 +347,7 @@ void ServerConfig::Fill()
 	}
 	SoftLimit = ConfValue("performance")->getUInt("softlimit", (SocketEngine::GetMaxFds() > 0 ? SocketEngine::GetMaxFds() : LONG_MAX), 10);
 	CCOnConnect = ConfValue("performance")->getBool("clonesonconnect", true);
-	MaxConn = ConfValue("performance")->getUInt("somaxconn", SOMAXCONN);
+	MaxConn = static_cast<unsigned int>(ConfValue("performance")->getUInt("somaxconn", SOMAXCONN, 1, SOMAXCONN));
```
SadieCat, line 350: Is this valid? You can set net.core.somaxconn higher than SOMAXCONN (4096 on my system) with sysctl.

src/listmode.cpp
```diff
@@ -140,8 +140,9 @@ unsigned int ListModeBase::GetLowerLimit()
 	unsigned int limit = UINT_MAX;
 	for (const auto& chanlimit : chanlimits)
 	{
+		// This cast is safe as we start at UINT_MAX and reduce from there.
 		if (chanlimit.limit < limit)
-			limit = chanlimit.limit;
+			limit = static_cast<unsigned int>(chanlimit.limit);
```
SadieCat, line 145: The method returns unsigned long but you cast down to unsigned int here. Is there a reasonf ro that?

src/listensocket.cpp
```diff
@@ -91,14 +91,15 @@ ListenSocket::ListenSocket(std::shared_ptr<ConfigTag> tag, const irc::sockets::s
 	if (bind_to.family() == AF_UNIX)
 	{
 		const std::string permissionstr = tag->getString("permissions");
-		unsigned int permissions = strtoul(permissionstr.c_str(), NULL, 8);
+		unsigned long permissions = strtoul(permissionstr.c_str(), NULL, 8);
 		if (permissions && permissions <= 07777)
-			chmod(bind_to.str().c_str(), permissions);
+			// This cast is safe thanks to the above check.
+			chmod(bind_to.str().c_str(), static_cast<int>(permissions));
```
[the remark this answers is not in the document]
dominichamon: Done

src/listmode.cpp
dominichamon: `GetLowerLimit` (this method) returns `unsigned int` still, which is safe per the above comment. `GetLimit` returns `unsigned long` with my changes in this PR as there was no safe way to keep it as `int`. This is inconsistent though so if you think this (`GetLowerLimit`) should return `unsigned long` too, i can make that change.

SadieCat: Ah yeah, you can alter this method too.

SadieCat:
Seems okay now. If CI passes I'll go through it again tomorrow morning and then merge.

SadieCat (approved):
Looks okay to me.

make/template/main.mk
```diff
@@ -74,7 +74,10 @@ else
 endif
 
 ifneq ($(COMPILER), ICC)
-  CORECXXFLAGS += -Woverloaded-virtual -Wshadow
+    CORECXXFLAGS += -Woverloaded-virtual -Wshadow -Werror
```
genius3000, line 77: I don't think we want -Werror left here, @SadieCat ? Debug builds and CI enable that.

genius3000 (changes requested):
Other than a few nitpicks on indentation and comments, it looks good to me as well.

make/template/main.mk
SadieCat: Ah, yeah. -Werror is enabled in debug builds already. Enabling it in normal builds is a bit excessive.

dominichamon: I left this in by mistake as it was helpful for me to catch issues. I think it could be enabled to prevent backsliding, but having it in debug is probably enough.

src/modules/m_blockamsg.cpp
```diff
@@ -109,7 +109,12 @@ class ModuleBlockAmsg : public Module
 			// OR
 			// The number of target channels is equal to the number of channels the sender is on..a little suspicious.
 			// Check it's more than 1 too, or else users on one channel would have fun.
-			if ((m && (m->message == parameters[1]) && (!irc::equals(m->target, parameters[0])) && ForgetDelay && (m->sent >= ServerInstance->Time()-ForgetDelay)) || ((targets > 1) && (targets == user->chans.size())))
+			if ((m &&
```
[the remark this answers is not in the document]
dominichamon: i did this for debugging but i do think it helps. i think this indentation is better now.

src/coremods/core_dns.cpp
```diff
@@ -344,7 +344,7 @@ class Packet : public Query
 				}
 				else
 				{
-					unsigned long forward = ip.in4.sin_addr.s_addr;
+					in_addr_t forward = ip.in4.sin_addr.s_addr;
```
SadieCat, line 347: Does this typedef exist on Windows? If not you might need to add a typedef to `inspircd_win32wrapper.h` for it.

dominichamon: it should map to `IN_ADDR` i think, but then i'm surprised that access to `s_addr`  works on windows because i _think_ it [should be spelled](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/ns-winsock2-in_addr#:~:text=The%20in_addr%20structure%20is%20the,with%20Windows%20Vista%20and%20later.&text=On%20earlier%20versions%20of%20the,be%20declared%20as%20struct%20in_addr.) `S_addr`.

i put this back to `unsigned int` :)

SadieCat:

PR #1915, opened 2021-07-07
Precompile inspircd.h

ProgVal:
## Summary

`inspircd.h` includes a lot of headers (from inspircd or the stdlib), and is included by essentially every compilation unit.
Precompiling it saves a lot of computation time, at the expense of disk space (71MB).

## Rationale

This reduces the compile time by ~40% on my machine. Before:

```
make -j 4  762,05s user 51,53s system 247% cpu 5:28,87 total
```

after:

```
make -j 4  440,73s user 33,39s system 243% cpu 3:14,82 total
```

## Discussion

I don't have much experience working with this kind of compilation tooling or perl, so this code is probably awful. Suggestions are welcome.

In particular, I can't get rid of this warning:

```
/home/dev-irc/inspircd/src/../include/inspircd.h:32:9: warning: #pragma once in main file
 #pragma once
         ^~~~
```

It seems to be a [known issue](https://stackoverflow.com/q/56563679/539465) :/

## Testing Environment

**Operating system name and version:** Debian 10 (GNU/Linux)

## Checks

I have ensured that:

### DIFF
--- a/make/calcdep.pl
+++ b/make/calcdep.pl
@@ -52,7 +52,10 @@ sub run() {
 	chdir BUILDPATH or die "Could not open build directory: $!";
 	unlink 'include';
 	symlink "${\SOURCEPATH}/include", 'include';
-	mkdir $_ for qw/bin modules obj/;
+	mkdir $_ for qw/bin gch modules obj/;
+
+	# fallback in case gch/inspircd.h.gch can't be used:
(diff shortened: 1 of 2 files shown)


ProgVal:
As discussed on IRC, it's too much trouble for the small benefits we get from this

PR #1919, opened 2021-07-20
Fixed multiple CWE

GermanAizek:
## Summary

Hi again, I review the code again. I found minor flaws, look at them and rate them.
Thanks in advance.

## Tested enviroments

- Linux 5.x
- Windows 10


src/inspstring.cpp
```diff
@@ -123,9 +123,9 @@ std::string Base64::Encode(const void* data, size_t length, const char* table, c
 	for (size_t idx = 0; idx < length; )
 	{
 		// Base64 encodes three octets into four characters.
-		uint32_t octet1 = idx < length ? udata[idx++] : 0;
```
SadieCat, line 126: This seems like a false positive to me? `idx++` may increase udata past the end of the array.

src/listensocket.cpp
```diff
@@ -92,7 +92,7 @@ ListenSocket::ListenSocket(std::shared_ptr<ConfigTag> tag, const irc::sockets::s
 	{
 		const std::string permissionstr = tag->getString("permissions");
 		unsigned long permissions = strtoul(permissionstr.c_str(), NULL, 8);
-		if (permissions && permissions <= 07777)
+		if (permissions && permissions <= 7777)
```
SadieCat, line 95: This is specifically intended to be octal. Octal 7777 is the maximum value for permissions for a file.

src/modules/m_channelban.cpp
```diff
@@ -49,8 +49,12 @@ class ChannelExtBan
 		{
 			if (!InspIRCd::Match(memb->chan->name, target))
 				continue;
-			if (!status || memb->getRank() >= mh->GetPrefixRank())
-				return true;
+
+			if (mh)
```
SadieCat, line 53: This is a false positive. Status is only set to non-`0` if mh is defined.

src/modules/m_connflood.cpp
```diff
@@ -87,7 +87,7 @@ class ModuleConnFlood : public Module
 
 		if (throttled)
 		{
-			if (tdiff > time_t(seconds + timeout))
+			if (tdiff > (time_t)seconds + timeout)
```
SadieCat, line 90: This is causing a compiler error because of the sign conversion in the add.

SadieCat:
In the future can you use feature branches which are rebased against master and only contain the changes you want in your PR so your previous commits aren't included in the branch?

SadieCat:
As you haven't responded to this for over a month I assume you're not interested in fixing it. Please reopen with the fixes I mentioned if you are interested. Thanks.

PR #1921, opened 2021-08-04
Fix argv index error in 'unknown option' message.

ProgVal:
## Summary

It looks like ya_getopt increments `optind` between reading the argument and returning, so it needs to be decremented before accessing argv.

## Rationale

Before

```
$ ./build/GCC-8.3/bin/inspircd --foo bar
Error: unknown option 'bar'.
$ ./build/GCC-8.3/bin/inspircd --help
Error: unknown option '%
```

(`%` means there is no newline at the end)

After:

```
$ ./build/GCC-8.3/bin/inspircd --foo bar
Error: unknown option '--foo'.
Usage: ./build/GCC-8.3/bin/inspircd [--config <file>] [--debug] [--nofork] [--nolog]
                                    [--nopid] [--runasroot] [--version]
$ ./build/GCC-8.3/bin/inspircd --help
Error: unknown option '--help'.
Usage: ./build/GCC-8.3/bin/inspircd [--config <file>] [--debug] [--nofork] [--nolog]
                                    [--nopid] [--runasroot] [--version]

```
## Testing Environment

## Checks

I have ensured that:

### DIFF
--- a/src/inspircd.cpp
+++ b/src/inspircd.cpp
@@ -324,7 +324,7 @@ namespace
 
 				default:
 					// An unknown option was specified.
-					std::cout << con_red << "Error:" <<  con_reset << " unknown option '" << argv[optind] << "'." << std::endl
+					std::cout << con_red << "Error:" <<  con_reset << " unknown option '" << argv[optind-1] << "'." << std::endl
 						<< con_bright << "Usage: " << con_reset << argv[0] << " [--config <file>] [--debug] [--nofork] [--nolog]" << std::endl
 						<< std::string(strlen(argv[0]) + 8, ' ') << "[--nopid] [--runasroot] [--version]" << std::endl;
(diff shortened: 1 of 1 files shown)


SadieCat:
This is weird. I'm pretty sure we changed this because the *opposite* happened at one point.

I wonder if this is a compat difference between ya_getopt (used for Windows) and GNU getopt (used on other platforms)?

ProgVal:
I have no idea :|

SadieCat:
I'll dig into Git history and try to check Windows tomorrow to see if this is still an issue there. We did switch option parser library from our handrolled thing to ya_getopt a while back so it may no longer be an issue.

SadieCat:
Yeah this is definitely also broken on Windows. Weird. 

Looking into Git history now.

SadieCat:
Hmmm it seems like the old Windows getopt was working fine too.

I have no idea why this was changed.

PR #1923, opened 2021-08-14
Send RPL_WHOISACTUALLY instead of RPL_WHOISHOST

ProgVal:
## Summary

Use RPL_WHOISACTUALLY instead of RPL_WHOISHOST

## Rationale

Unlike RPL_WHOISHOST, it contains the user@host and ip as separate parameters,
instead of having them in the free-text parameter.

RPL_WHOISACTUALLY is also somewhat more popular than RPL_WHOISHOST.

## Discussion

To my knowledge, there are three different syntaxes for RPL_WHOISACTUALLY:

1. `<client> <nick> :is actually ...` (bahamut)
2. `<client> <nick> <host> :Is actually using host` (charybdis family)
3. `<client> <nick> <username>@<host> <ip> :Is actually using host`
   (ircu2, ergo, hybrid, ...)

I chose the third one because it contains the most info, and is the most
common one. But the second one would be fine too.

## Testing Environment

## Checks

I have ensured that:

  - [ ] This pull request does not introduce any incompatible API changes.
  - [x] I have documented any features added by this pull request. -> WIP https://github.com/ircdocs/modern-irc/pull/115


src/coremods/core_whois.cpp
```diff
@@ -212,6 +212,8 @@ void CommandWhois::DoWhois(LocalUser* user, User* dest, time_t signon, unsigned
 	whois.SendLine(RPL_WHOISUSER, dest->ident, dest->GetDisplayedHost(), '*', dest->GetRealName());
 	if (!dest->server->IsULine() && (whois.IsSelfWhois() || user->HasPrivPermission("users/auspex")))
 	{
+		/* Clients should prefer RPL_WHOISACTUALLY as it is easier to parse, but we provide RPL_WHOISHOST (despite it being redundant) for backward compatibility */
+		whois.SendLine(RPL_WHOISACTUALLY, InspIRCd::Format("%s@%s", dest->ident.c_str(), dest->GetRealHost().c_str()), dest->GetIPString().c_str(), "is connecting from");
```
SadieCat, line 216: You can use user->MakeHost() here.

PR #1927, opened 2021-09-20
m_repeat: allow opers to be immune to repeat punishment

### LINKED ISSUE #1847: repeat should not affect opers
Currently the repeat module does affect all users, even if opered.

The module should receive a config option or fix to enable opers to be excluded.

examknow:
## Summary

Allows network admins to make opers immune to repeat punishment (mode `+E`) via a config setting (`repeat/opersimmune`)

## Rationale

Resolves #1847

## Testing Environment

**Compiler name and version:** gcc version 8.3.0

## Checks

I have ensured that:

### DIFF
--- a/docs/conf/modules.conf.example
+++ b/docs/conf/modules.conf.example
@@ -1924,14 +1924,12 @@
 #               any effect, as the maximum length of a message on IRC
 #               cannot exceed that.
 # kickmessage - Kick message when * is specified
-# opersimmune - Should opers be affected by repeat checking?
 #<repeat maxbacklog="20"
 #        maxdistance="50"
 #        maxlines="20"
 #        maxtime="0s"
 #        size="512"
-#        kickmessage="Repeat flood"
-#        opersimmune="yes">
+#        kickmessage="Repeat flood">
 #<module name="repeat">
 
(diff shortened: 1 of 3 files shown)


genius3000:
Good idea! Although this will be better implemented as an oper permission, allows more granular control if wanted. We're slowly changing oper bypasses to permissions for that reason.

Have a look at [m_callerid](https://github.com/inspircd/inspircd/blob/insp3/src/modules/m_callerid.cpp#L418) for how to implement it. Only catch is it has to be on a local user, which m_repeat already requires in this function block.
I would suggest a permission name of `channels/ignore-repeat` and then update the example opers.conf.

docs/conf/modules.conf.example
```diff
@@ -2048,7 +2048,7 @@
 
 # You must define <sasl:target> to the name of your services server so
 # that InspIRCd knows where to send SASL authentication messages and
-# when it should enable the SASL capability.
+# when it should enable the SASL capability.
```
SadieCat, line 2051: Can you revert this and the added blank line at the end of the file?

examknow: I apologize on behalf of my silly editor!

PR #1929, opened 2021-09-28
Minor code refactor.

GermanAizek:
Hello again, found minor flaws and fixed them.
Waiting for feedback.

### DIFF
--- a/src/inspircd.cpp
+++ b/src/inspircd.cpp
@@ -292,6 +292,7 @@ namespace
 
 		char** argv = ServerInstance->Config->cmdline.argv;
 		int ret;
+		size_t len = strlen(argv[0]);
 		while ((ret = getopt_long(ServerInstance->Config->cmdline.argc, argv, ":c:", longopts, NULL)) != -1)
 		{
 			switch (ret)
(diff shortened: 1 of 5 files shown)


src/modules/m_spanningtree/main.cpp
```diff
@@ -705,9 +705,8 @@ namespace
 			if (!linkstring.empty())
 				buffer << '=' << linkstring;
 
-			const std::string compatlinkstring = Utils->BuildLinkString(PROTO_INSPIRCD_3, mod);
```
SadieCat, line 708: This is actually a bug. The line below should be checking if `compatlinkstring` is empty not `linkstring`.

src/users.cpp
```diff
@@ -1227,11 +1227,11 @@ ConnectClass::ConnectClass(std::shared_ptr<ConfigTag> tag, char t, const std::ve
 }
 
 ConnectClass::ConnectClass(std::shared_ptr<ConfigTag> tag, char t, const std::vector<std::string>& masks, std::shared_ptr<ConnectClass> parent)
+	: hosts(masks)
```
SadieCat, line 1230: This is wrong because the call to Update() might change the hosts field even though we want the value passed in the constructor to be used.

src/inspircd.cpp
```diff
@@ -309,7 +310,7 @@ namespace
 					// An unknown option was specified.
 					std::cout << con_red << "Error:" <<  con_reset << " unknown option '" << argv[optind-1] << "'." << std::endl
 						<< con_bright << "Usage: " << con_reset << argv[0] << " [--config <file>] [--debug] [--nofork] [--nolog]" << std::endl
-						<< std::string(strlen(argv[0]) + 8, ' ') << "[--nopid] [--runasroot] [--version]" << std::endl;
+						<< std::string(len + 8, ' ') << "[--nopid] [--runasroot] [--version]" << std::endl;
```
SadieCat, line 313: `len` is not used anywhere else so this is fine as it was.

SadieCat:
I've merged in the one valid change from this pr, thanks.

PR #1959, opened 2022-01-16
Insp3 fix delaymsg exemption

iwalkalone69:
This pull request tries to use HasPrivPermission and exemptchanops to be able to exclude operators or some channel ranks like op and voice from being affected by channel +d mode (delaymsg module).

- Condition that channel rank voice or upper is not affected is removed.
- Added condition to exempt operators with channels/ignore-delaymsg priv.
- Added condition to exempt users with CheckExemption::Call "delaymsg" so if you add delaymsg:v into exemptchanops all ranks equal or above voice (v) are not affected.
- Updated inspircd.conf.example to document new delaymsg option for exemptchanops.
- Updated opers.conf.example to document new operators privilege channels/ignore-delaymsg.

I have tested with latest insp3 branch.

**Operating system name and version:** 5.13.0-23-generic
**Compiler name and version:** GCC 11.2.0

## Checks

I have ensured that:

  - [ x ] If ABI changes have been made I have incremented MODULE_ABI in `moduledefs.h`.

### DIFF
--- a/docs/conf/inspircd.conf.example
+++ b/docs/conf/inspircd.conf.example
@@ -699,6 +699,8 @@
          #  - stripcolor      Channel mode +S - strips formatting codes from
          #                    messages (requires the stripcolor module).
          #  - topiclock       Channel mode +t - limits changing the topic to (half)ops
+         #  - delaymsg        Channel mode +d - blocks sending messages until specified
+         #                    seconds have passed since user join
          # You can also configure this on a per-channel basis with a channel mode and
          # even negate the configured exemptions below.
          # See exemptchanops in modules.conf.example for more details.
--- a/docs/conf/inspircd.conf.example
+++ b/docs/conf/inspircd.conf.example
@@ -682,6 +682,8 @@
          #                    (requires the blockcolor module).
          #  - censor          Channel mode +G - censors messages based on the network
          #                    configuration (requires the censor module).
+         #  - delaymsg        Channel mode +d - blocks sending messages until specified
+         #                    seconds have passed since user join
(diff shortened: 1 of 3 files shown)


docs/conf/inspircd.conf.example
```diff
@@ -699,6 +699,8 @@
          #  - stripcolor      Channel mode +S - strips formatting codes from
          #                    messages (requires the stripcolor module).
          #  - topiclock       Channel mode +t - limits changing the topic to (half)ops
+         #  - delaymsg        Channel mode +d - blocks sending messages until specified
```
SadieCat, line 702: The list here is in alphabetical order.

docs/conf/opers.conf.example
```diff
@@ -30,6 +30,7 @@
      #   - channels/ignore-noctcp: allows opers with this priv to send a CTCP to a +C channel.
      #   - channels/ignore-nonicks: allows opers with this priv to change their nick when on a +N channel.
      #   - channels/ignore-repeat: allows opers with this priv to be immune to repeat punishment on a +E channel.
+     #   - channels/ignore-delaymsg: allows opers with this priv to be immune to delaymsg restriction on a +d channel.
```
SadieCat, line 33: The list here is in alphabetical order.

docs/conf/inspircd.conf.example
iwalkalone69: Fixed

docs/conf/opers.conf.example
iwalkalone69: Fixed

SadieCat:

PR #1968, opened 2022-03-19
Add support for the <count> param of WHOWAS

progval:
Note: this PR includes the commit in #1967, because it touches the same line.

## Summary

Add support for the <count> param of WHOWAS, which limits the number of returned replies

## Rationale

Not very useful IMO, but every server but InspIRCd seems to implement it, and it's part of the RFCs:
    
* https://datatracker.ietf.org/doc/html/rfc1459#section-4.5.3
* https://datatracker.ietf.org/doc/html/rfc2812#section-3.6.3

## Testing Environment

**Operating system name and version:** Debian 11
**Compiler name and version:** GCC 10.2.1

## Checks

I have ensured that:

  - [x] I have documented any features added by this pull request. (though it's not on docs.inspircd.org yet, but I'll send a PR if this gets merged)

### DIFF
--- a/src/coremods/core_whowas.cpp
+++ b/src/coremods/core_whowas.cpp
@@ -187,7 +187,7 @@ class CommandWhowas : public Command
 CommandWhowas::CommandWhowas( Module* parent)
 	: Command(parent, "WHOWAS", 1)
 {
-	syntax = "<nick>";
+	syntax = "<nick> [<count>]";
 	Penalty = 2;
 }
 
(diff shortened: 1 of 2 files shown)


src/coremods/core_whowas.cpp
```diff
@@ -200,6 +200,13 @@ CmdResult CommandWhowas::Handle(User* user, const Params& parameters)
 		return CMD_FAILURE;
 	}
 
+	unsigned int count = UINT_MAX;
+	if (parameters.size() > 1) {
+		count = ConvToNum<unsigned int>(parameters[1]);
+		if (count == 0)
+			count = UINT_MAX;
+	}
```
SadieCat, line 208: Untested (pls check it) but something like this might be better:

```suggestion
	WhoWas::Nick::List::const_reverse_iterator last = list.rend();
	if (parameters.size() > 1)
	{
		size_t count = ConvToNum<size_t>(parameters[1]);
		if (count > 0 && std::distance(list.rbegin(), last) < count)
			last = list.rbegin() + count + 1;
	}
```

... and then use `last` instead of `list.rend()`.

PR #1981, opened 2022-04-30
Master

ins64:
Feature request 
Could it be possible to make the filter system more informative?
By adding a new action, warnmsg – server operators could see the entire message which triggered the matching pattern. 

Also could it be possible to show which class a user is connecting from a remote server instead of just showing on local.
Thanks


SadieCat:
> Could it be possible to make the filter system more informative? By adding a new action, warnmsg – server operators could see the entire message which triggered the matching pattern.

We're not making it so server operators can see messages that matched filters as it allows server operators to snoop on messages.

> Also could it be possible to show which class a user is connecting from a remote server instead of just showing on local. Thanks

Snomask `C` will show remote users. 

https://docs.inspircd.org/3/snomasks/

--

In the future please file an issue instead of a pull request.

https://github.com/inspircd/inspircd/issues/new/choose

PR #1984, opened 2022-05-15
Fix segfault in m_ircv3_accounttag

progval:


SadieCat:
This fix didn't cover the case where services_account isn't loaded.

Fixed in dd6b91837ad4a195fefa496d521bc37a40ca3901. Thanks for reporting.

PR #1991, opened 2022-07-10
Add OpenRC support to logrotate

clinew:
## Summary

Add OpenRC "support" to logrotate.

## Rationale

The current method doesn't work on Gentoo.  Gentoo installs files into `/lib/systemd` even though systemd might not be in use, causing the `systemctl` commands to be run (which fail).  Checking for OpenRC and running its specific rehash command before systemd fixes the issue.

## Testing Environment

**Operating system name and version:** Gentoo
**Compiler name and version:** GCC 11.3.0
## Checks

I have ensured that:

### DIFF
--- a/make/template/logrotate
+++ b/make/template/logrotate
@@ -28,7 +28,10 @@
 	rotate 8
 	weekly
 	postrotate
-		if [ -d /lib/systemd ]
+		if [ -r "@RUNTIME_DIR@/inspircd.pid" ]
+		then
+			kill -HUP $(cat "@RUNTIME_DIR@/inspircd.pid")
+		elif [ -d /lib/systemd ]
 		then
 			if systemctl --quiet is-active inspircd
 			then


SadieCat:
We don't currently ship an OpenRC init script so this is a bit out of place. Would you be interested in upstreaming support for that?

clinew:
Maybe.  What would upstreaming OpenRC support entail?

It might be easier to refactor the logrotate file a bit.  Looking at Mumble shows:
```
        postrotate
                test -r /run/murmur/murmur.pid && kill -HUP $(cat /run/murmur/murmur.pid) &>/dev/null
        endscript
```
The OpenRC command is but a fancy wrapper for sending `SIGHUP` as well:
```
            rehash() {                                                         
                    ebegin "Rehashing InspIRCd"                                    
                    start-stop-daemon --signal HUP --pidfile "${pidfile}"
                    eend $?                       
            }                                      
```
So if the first line checks for the standard PID file location and sends SIGHUP if found that would presumably work cross-platform without adding more init-system specific scripts.

Let me know which you'd prefer to see and I'll go from there!

SadieCat:
>Maybe. What would upstreaming OpenRC support entail?

Integrating it into the build system. See `make/template/inspircd.service` for an example on this.

>It might be easier to refactor the logrotate file a bit. Looking at Mumble shows:

Our existing scripts for other init systems (launchd, systemd) don't write a pid file as they're non-forking and handle rehashes internally so doing this might not be possible. It also won't work if the user changes the location of their pid file in the config without updating the init script.

clinew:
So it turns out that OpenRC uses a configuration value for the PID file location, and one needs to change that anyways if one changes the PID in their config.  Thus adding OpenRC support explicitly wouldn't really avoid the double-config modification problem (though it would increase complexity).  I don't see any way around this without adding and calling a tool to parse or display config values.

I think the best thing is to check for a PID file at a sane location, and move on if it's not found.  This allows package maintainers to set sane defaults when using a PID but will also work with other init systems.  The current PR reflects this.  Thoughts?

PR #2037, opened 2023-05-08
Fix cert expiration for dates after year 2049

clinew:
## Summary

Fixes issue where certificates which expire after 2049 might erroneously be marked as expired.

## Rationale

I generated certificates to last 'till 2120 in my test environment so I wouldn't have to deal with tests failing due to expiration.  The irony is rich.

## Testing Environment

Gentoo VM.

**Operating system name and version:** Gentoo!  `Linux test 5.15.88 #1 SMP Sun Feb 12 17:24:09 PST 2023 x86_64 QEMU Virtual CPU version 2.5+ GenuineIntel GNU/Linux`
**Compiler name and version:**  gcc (Gentoo Hardened 12.2.1_p20230428-r1 p2) 12.2.1 20230428

## Checks

I have ensured that:

### DIFF
--- a/src/modules/extra/m_ssl_openssl.cpp
+++ b/src/modules/extra/m_ssl_openssl.cpp
@@ -704,11 +704,11 @@ class OpenSSLIOHook : public SSLIOHook
 		certinfo->activation = GetTime(X509_getm_notBefore(cert));
 		certinfo->expiration = GetTime(X509_getm_notAfter(cert));
 
-		int activated = ASN1_UTCTIME_cmp_time_t(X509_getm_notBefore(cert), ServerInstance->Time());
+		int activated = ASN1_TIME_cmp_time_t(X509_getm_notBefore(cert), ServerInstance->Time());
 		if (activated != -1 && activated != 0)
 			certinfo->error = "Certificate not activated";
 
-		int expired = ASN1_UTCTIME_cmp_time_t(X509_getm_notAfter(cert), ServerInstance->Time());
+		int expired = ASN1_TIME_cmp_time_t(X509_getm_notAfter(cert), ServerInstance->Time());
 		if (expired != 0 && expired != 1)
 			certinfo->error = "Certificate has expired";
 


src/modules/extra/m_ssl_openssl.cpp
```diff
@@ -704,13 +704,23 @@ class OpenSSLIOHook : public SSLIOHook
...
-		int activated = ASN1_UTCTIME_cmp_time_t(X509_getm_notBefore(cert), ServerInstance->Time());
-		if (activated != -1 && activated != 0)
-			certinfo->error = "Certificate not activated";
+		int activated = ASN1_TIME_cmp_time_t(X509_getm_notBefore(cert), ServerInstance->Time());
+		if (activated != -1 && activated != 0) {
+			char date[64];
+			struct tm tm;
+			gmtime_r(&certinfo->activation, &tm);
+			strftime(date, 64, "%b %e %T %Y %Z", &tm);
+			certinfo->error = "Certificate not activated (notBefore '" + std::string(date) + "')";
```
clinew, line 713: This is obviously hacky C-style code and almost certainly inappropriate for C++, but I wasn't going to try and battle search engines to figure out which of the 500 string-concatenation methods I should actually use.  If you let me know what method to use then I can refactor accordingly.

The output was made to mirror that given by `openssl x509 -in foo.pem -text`.  Example from my fork with some enhanced OpenSSL logging:
```
            Sun May 07 2023 21:42:01 m_ssl_openssl: Invalid peer certificate chain from '127.0.0.1' port '6697' ('Certificate has expired (notAfter 'Dec 25 20:25:54 2120 GMT'), Cert chain #1: Certificate has expired (notAfter 'Jun  9 07:12:20 2120 GMT'), Cert chain #2: Certificate has expired (notAfter 'Jun  9 07:12:17 2120 GMT')'):
```

clinew:
Huh.  Apparently gnu++98 doesn't support `%e` string format.  Will deal with that later with probable string refactoring.

SadieCat:
This introduced a regression on older versions of OpenSSL which is now fixed in commit e607175dfa8e53d35964e5f650babb597c1c72b5.

PR #2064, opened 2023-10-14
Do initial config check before ForkIntoBackground

satmd:
## Summary

This PR reorders the initial config check to happen *before* ForkIntoBackground.

## Rationale

Potential fix for issue 1369.

I reorder the call to this->Config->Read and ->Apply to happen before forking so that we exit in the parent process with a proper exit code.
I've tested with `./bin/inspircd -F -d` and `./bin/inspircd -d`. Both return an exitcode 1 for my test case.

The only other use of Config->Apply is in `src/configreader.cpp`, line 721. That code seems to handle rehash and should independently enough from `src/main.cpp`. I *assume* this will not be affected, but I can test it if you want me to.

## Testing Environment

Plain inspircd *master* compile with an additional `<delibertate error` added to inspircd.conf

`src/main.cpp` seems to be the same for insp3 and master, so probably easy backport.

**Operating system name and version:** Linux 6.5
**Compiler name and version:** GCC 13.2.1 (gentoo hardened)

## Checks

I have ensured that:

  - [X] If ABI changes have been made I have incremented MODULE_ABI in `moduledefs.h`.

### DIFF
--- a/src/inspircd.cpp
+++ b/src/inspircd.cpp
@@ -498,8 +498,6 @@ InspIRCd::InspIRCd(int argc, char** argv)
 	SetSignals();
 	if (!Config->CommandLine.runasroot)
 		CheckRoot();
-	if (!Config->CommandLine.nofork)
-		ForkIntoBackground();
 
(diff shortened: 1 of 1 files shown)


SadieCat:
What config was this tested with? I'm unsure how well sockets/modules/etc can survive across forks especially those in modules that have third-party dependencies that we can't affect the internals of.

PR #2071, opened 2023-12-15
Call $compiler with --version instead of -v

satmd:
## Summary

This is an attempt to fix issue 2070

clang++ does not have -v for --version, so call it by the long option

Interestingly the bug manifested very strangely on my machine - maybe user error too.

GOOD: CXX=clang++ ./configure
BAD: CXX=clang++ ./configure --disable-ownership
BAD: CXX=clang++ ./configure --disable-ownership | cat
BAD: CXX=clang++ ./configure | cat

## Rationale

People try to build inspircd using clang and this fails for example in https://bugs.gentoo.org/919508

## Testing Environment

## Checks

I have ensured that:

### DIFF
--- a/make/configure.pm
+++ b/make/configure.pm
@@ -94,7 +94,7 @@ sub __get_template_settings($$$) {
 
 sub __test_compiler($) {
 	my $compiler = shift;
-	return 0 unless run_test("`$compiler`", !system "$compiler -v ${\CONFIGURE_ERROR_PIPE}");
+	return 0 unless run_test("`$compiler`", !system "$compiler --version ${\CONFIGURE_ERROR_PIPE}");
 	return 0 unless run_test("`$compiler`", test_file($compiler, 'compiler.cpp', '-fno-rtti'), 'compatible');
 	return 1;
 }


SadieCat:
Clang *does* have `-v` for compatibility with GCC so this shouldn't be the issue.

I don't have a Gentoo system around but I can't reproduce this on any of the following Clang versions which I should probably be able to do it if it was a Clang issue:

- Alpine clang version 17.0.5
- Ubuntu clang version 16.0.6

Is Gentoo `clang++` a wrapper? I'm wondering if its being weird.

satmd:
```
# which clang
/usr/lib/llvm/17/bin/clang
# which clang++
/usr/lib/llvm/17/bin/clang++
# file $(which clang++)
/usr/lib/llvm/17/bin/clang++: symbolic link to clang++-17
# file $(which clang++-17)
/usr/lib/llvm/17/bin/clang++-17: symbolic link to clang-17
# file $(which clang-17)
/usr/lib/llvm/17/bin/clang-17: ELF 64-bit LSB pie executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, stripped
# clang-17 -v
clang version 17.0.6
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm/17/bin
Configuration file: /etc/clang/x86_64-pc-linux-gnu-clang.cfg
System configuration file directory: /etc/clang
Selected GCC installation: /usr/lib/gcc/x86_64-pc-linux-gnu/13
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Selected multilib: .;@m64
 "/usr/bin/x86_64-pc-linux-gnu-ld.bfd" -pie --hash-style=gnu --eh-frame-hdr -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o a.out /usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../lib64/Scrt1.o /usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../lib64/crti.o /usr/lib/gcc/x86_64-pc-linux-gnu/13/crtbeginS.o -L/usr/lib/gcc/x86_64-pc-linux-gnu/13 -L/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/lib -L/lib -L/usr/lib -z relro -z now -lgcc --as-needed -lgcc_s --no-as-needed -lc -lgcc --as-needed -lgcc_s --no-as-needed /usr/lib/gcc/x86_64-pc-linux-gnu/13/crtendS.o /usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../lib64/crtn.o
/usr/bin/x86_64-pc-linux-gnu-ld.bfd: /usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../lib64/Scrt1.o: in function `_start':
(.text+0x1b): undefined reference to `main'
clang-17: error: linker command failed with exit code 1 (use -v to see invocation)
```

SadieCat:
I think Clang behaving like this on Gentoo is a bug in how Gentoo has packaged it.

Nonetheless, I'll merge this into master as it doesn't really hurt to change this. Unfortunately this doesn't really work for insp3 though as this would break LLVM-GCC which we still support on that branch.

PR #2102, opened 2024-07-19
Used standardised websocket subprotocol names

ItsOnlyBinary:
## Summary

The websocket module should use the standardised subprotocol names as specified here: https://ircv3.net/specs/extensions/websocket

## Testing Environment

Tested with kiwiirc and https://ircdocs.horse/tools/wstester 

> **Note**
> the following must be set in your modules.conf
> `<websocket defaultmode="reject">`

**Operating system name and version:** Archlinux 6.9.7-arch1-1 x86_64.
**Compiler name and version:**  make 4.4.1-2, gcc 14.1.1+r58+gfc9fb69ad62-1

## Checks

I have ensured that:

  - [ ] This pull request does not introduce any incompatible API changes (stable branches only).
  - [ ] If ABI changes have been made I have incremented MODULE_ABI in `moduledefs.h` (stable branches only).

### DIFF
--- a/src/modules/m_websocket.cpp
+++ b/src/modules/m_websocket.cpp
@@ -468,8 +468,8 @@ class WebSocketHook final
 			{
 				proto.erase(std::remove_if(proto.begin(), proto.end(), ::isspace), proto.end());
 
-				bool is_binary = insp::equalsci(proto, "binary.inspircd.org");
-				bool is_text = insp::equalsci(proto, "text.inspircd.org");
+				bool is_binary = insp::equalsci(proto, "binary.ircv3.net");
+				bool is_text = insp::equalsci(proto, "text.ircv3.net");
 
 				if (is_binary || is_text)
 				{


ItsOnlyBinary:
I just read the title of the initial commit related to this pull-request, its vendored due to being draft spec, Feel free to close this pull-request. 

I could add the vendored version to the protocol array in kiwiirc. 

The other big ircd I shall not name does not use vendored if that helps the decision.

SadieCat:
Our policy is to not ship IRCv3 drafts (our websocket module predates IRCv3) because we don't make breaking changes in minor releases. However, I recognise that the current scenario is not great for client developers and it seems unlikely that the IRCv3 prefix names would be changing at this point. I'll see if we can come up with a compromise.

SadieCat:
I'm calling the IRCv3 names a de-facto standard. ;-)

https://github.com/inspircd/inspircd/commit/dafaa684e174d20da46a5b5c6b46e8703d39960b

PR #2106, opened 2024-09-18
Hide channel list on whois for servprotect user

mog422:
## Summary

Hide channel list on whois for servprotect user

## Rationale

Restores [behavior](https://github.com/inspircd/inspircd/blob/855c1f6ab4927e219220ac2a9a14f0d896c18c6f/src/modules/m_servprotect.cpp#L135) lost in v4.

## Testing Environment

**Operating system name and version:** Linux 6.9.7
**Compiler name and version:** gcc version 14.2.1

## Checks

I have ensured that:

### DIFF
--- a/src/modules/m_services.cpp
+++ b/src/modules/m_services.cpp
@@ -23,6 +23,7 @@
 #include "modules/ctctags.h"
 #include "modules/server.h"
 #include "modules/stats.h"
+#include "modules/whois.h"
 #include "timeutils.h"
 #include "xline.h"
 
@@ -517,6 +518,7 @@ class ModuleServices final
 	: public Module
 	, public ServerProtocol::RouteEventListener
 	, public Stats::EventListener
+	, public Whois::LineEventListener
 {
 private:
 	Account::API accountapi;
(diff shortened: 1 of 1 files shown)


SadieCat:
This was an intentional change. If you wish to hide the channels of a services pseudoclient you should use [the hidechans mode](https://docs.inspircd.org/4/modules/hidechans/).

mog422:
Oh, was that intentional? I didn't know because it wasn't in the documentation.

mog422:
However, using only hidechans mode, I cannot make it behave the same as before.
I want to hide it even from the oper if it is service. However, if other users, oper should be visible.

SadieCat:
Hmm I'm open to re-adding this if necessary. I do wonder if having it in the servprotect module is the best place though.

SadieCat:
That part isn't really controversial. I'm just unsure as to whether it being enabled by +k is the best behaviour. The point of servprotect is to protect services from kicks etc. Having it also hide channels feels inappropriate.

Anope already requires hidechans for services pseudoclients so I'm considering making hidechans hide the channel list for services even the source of the WHOIS is a server operator.

mog422:
Yes, I agree with you, but making another hidechans mode that is also hidden from oper seems like over-engineering.

It's not bad that it's also hidden from oper when in (servprotect && hidechans) mode, but in this case, I don't think it has "single clear purpose".

SadieCat:
I think commit 43a5068798cd8ca4b1e4f31a17d330c6dff35a91 should resolve this? If a services pseudoclient sets +I their channel list will no longer be visible even to server operators with the users/auspex privilege (with an opt-out config option). This behaves the same as +k did in v3.

mog422:
yes good job, thanks!

PR #2124, opened 2024-10-28
Set ownership only for file/directories which should be accesible by inspircd

sertonix:
## Summary

Only grant inspircd write access to files/directories that it needs to write to.

## Rationale

Some permissions could be considered dangerous (like write access to `/usr/bin`).

## Testing Environment

I don't really know inspircd so I would appreciate if someone else could test these changes. (I saw this issue while analyzing permission in alpine linux)

**Operating system name and version:** -
**Compiler name and version:** -

## Checks

I have ensured that:

  - [X] This pull request does not introduce any incompatible API changes (stable branches only).
  - [X] If ABI changes have been made I have incremented MODULE_ABI in `moduledefs.h` (stable branches only).

### DIFF
--- a/make/template/main.mk
+++ b/make/template/main.mk
@@ -78,9 +78,9 @@ endif
 # was not passed to configure.
 DISABLE_OWNERSHIP=@DISABLE_OWNERSHIP@
 ifeq ($(DISABLE_OWNERSHIP), 1)
-  INSTFLAGS =
+  INSTOWNER =
 else
-  INSTFLAGS = -g @GID@ -o @UID@
+  INSTOWNER = -g @GID@ -o @UID@
 endif
 
 # Force the use of libc++ on macOS as on some systems it is not the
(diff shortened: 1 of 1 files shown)


sertonix:
> I think this might cause problems for users who are building for another user on their system (e.g. running make install as root when they specified a user/group to run as).

The ownership of the config/run/data directories are still set. Shouldn't that be enough?

> which will not run install on the install directories if they already exist. Would this work for Alpine?

I don't think this will work for Alpine. The while packaging `make install` is run with an empty `DESTDIR` so no directories exist already.

SadieCat:
> > I think this might cause problems for users who are building for another user on their system (e.g. running make install as root when they specified a user/group to run as).
> 
> The ownership of the config/run/data directories are still set. Shouldn't that be enough?

If a user builds as root but wants to run as an unprivileged user this might cause permission issues. We've had issues with users bumping into this in the past

> 
> > which will not run install on the install directories if they already exist. Would this work for Alpine?
> 
> I don't think this will work for Alpine. The while packaging `make install` is run with `DESTDIR` set to an empty directory so other directories don't exist already.

I've been thinking about this and I'm wondering if passing `--disable-ownership` instead of `--gid` and `--uid` and then setting the permissions post-install might actually be the more correct solution here for Alpine. If you're using a package that someone else is building then the uid/gid from build time probably won't be correct anyway?

sertonix:
> If a user builds as root but wants to run as an unprivileged user this might cause permission issues with moving other files about. We've had issues with users bumping into this in the past.

Maybe only change things when `--system` is set then?

> I've been thinking about this and I'm wondering if passing --disable-ownership instead of --gid and --uid and then setting the permissions post-install might actually be the more correct solution here for Alpine.

I would be ok with that.

> If you're using a package that someone else is building then the uid/gid from build time probably won't be correct anyway?

As far as I know the build output only contains the user/group name and not the uid/gid so I think this works fine.

PR #2133, opened 2024-12-04
fix: m_pgsql: Clear vector before adding new values to prevent duplicates

L422Y:
## Summary

The SQL row vector was not being cleared between iterations, causing values to accumulate and duplicate. This adds result.clear() before populating each row.

## Rationale

In order to be able to use the pgsql module effectively

## Testing Environment

**Operating system name and version:** Ubuntu22.04.1 (aarch64) / macOS Sequoia 15.1
**Compiler name and version:** GCC 11.4 / clang 16.0.0 (clang-1600.0.26.4)

## Checks

I have ensured that:

  - [ ] If ABI changes have been made I have incremented MODULE_ABI in `moduledefs.h` (stable branches only).

### DIFF
--- a/src/modules/extra/m_pgsql.cpp
+++ b/src/modules/extra/m_pgsql.cpp
@@ -177,7 +177,8 @@ public:
 		if (currentrow >= PQntuples(res))
 			return false;
 		int ncols = PQnfields(res);
-
+		
+		result.clear();
 		for(int i = 0; i < ncols; i++)
 		{
 			result.push_back(GetValue(currentrow, i));


SadieCat:
Clearing the output vector in this case is the responsibility of the caller. Have you encountered a module which is causing problems here?

L422Y:
> Clearing the output vector in this case is the responsibility of the caller. Have you encountered a module which is causing problems here?

Derp, that would make sense. Thought I had found a big bug haha :)

It was an issue in a module I am working on for persisting channel history to a database.

Is this documented somewhere I could read up on? Thank you!

SadieCat:
I'm fine with changing the behaviour so keep this open. I was just curious if you were having problems with a module.

PR #2135, opened 2024-12-17
Two simple performance optimizations for large servers

shaggie76:
## Summary

On our server hosting 150K users was running hot and some perf captures turned up a few easy optimizations we thought we would share.

## Rationale

These two low-risk changes combined seem to have cut a few percent off of the CPU capture. 

## Testing Environment

These changes have been live in our production environment all day and there has been no problems.

Ubuntu 22.04.4 LTS
gcc version 11.4.0

## Checks

I have ensured that:

### DIFF
--- a/include/channels.h
+++ b/include/channels.h
@@ -43,7 +43,7 @@ class CoreExport Channel final
 public:
 	/** A map of Memberships on a channel keyed by User pointers
 	 */
-	typedef std::map<User*, insp::aligned_storage<Membership>> MemberMap;
+	typedef std::unordered_map<User*, insp::aligned_storage<Membership>> MemberMap;
 
 private:
 	/** Set default modes for the channel on creation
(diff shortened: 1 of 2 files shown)


shaggie76:
Oh one thing I failed to mention is that these changes were made against our copy of inspircd-3.14.0 not 4 but since they're so benign I can't imagine there's a problem with 4; I just can't field-test 4 yet with 150K users.

include/channels.h
```diff
@@ -43,7 +43,7 @@ class CoreExport Channel final
 public:
 	/** A map of Memberships on a channel keyed by User pointers
 	 */
-	typedef std::map<User*, insp::aligned_storage<Membership>> MemberMap;
+	typedef std::unordered_map<User*, insp::aligned_storage<Membership>> MemberMap;
```
SadieCat, line 46: I think changing this will cause any user list (e.g. `/NAMES`) to no longer be in alphabetical order.

shaggie76: std::map in this case would use `std::less<User*>` which does direct pointer value compares and would be in arbitrary order.

I checked `/names #inspircd.dev` on irc.teranova.net (InspIRCd-4) and confirmed it wasn't alphabetical without this change.

PR #2210, opened 2026-05-14
Fix two use-after-free crashes in ModuleManager::DoSafeUnload

aasb13:
## Summary

Fix two use-after-free crashes in ModuleManager::DoSafeUnload

## Rationale

When unloading a module, DoSafeUnload erases its entries from the DataProviders
map and then reads the erased iterator to call OnServiceDel. This is a
use-after-free that crashes in downstream handlers when they access the
ServiceProvider name. The shutdown also logs mod->ModuleFile after adding the
module to GlobalCulls, which can be freed before the log call. These crashes
are reproducible on graceful shutdown when the unloaded module provides
services.

## Testing Environment

**Operating system name and version:** Arch Linux 7.0.5-zen1-1-zen
**Compiler name and version:** GCC 16.1.1

## Checks

I have ensured that:

- [x] Generative AI (Copilot, ChatGPT, etc) was not used to create any part of this pull request.
- [x] This pull request does not introduce any incompatible API changes (stable branches only, delete if not applicable).

### DIFF
--- a/src/modules.cpp
+++ b/src/modules.cpp
@@ -440,6 +440,8 @@ void ModuleManager::DoSafeUnload(Module* mod)
 
 	std::map<std::string, Module*>::iterator modfind = Modules.find(mod->ModuleFile);
 
+	std::string modfile = mod->ModuleFile;
+
 	// Unregister modes before extensions because modes may require their extension to show the mode being unset
 	UnregisterModes(mod, MODETYPE_USER);
 	UnregisterModes(mod, MODETYPE_CHANNEL);
@@ -476,8 +478,9 @@ void ModuleManager::DoSafeUnload(Module* mod)
 		DataProviderMap::iterator curr = i++;
 		if (curr->second->creator == mod)
 		{
+			ServiceProvider* provider = curr->second;
 			DataProviders.erase(curr);
-			FOREACH_MOD(OnServiceDel, (*curr->second));
+			FOREACH_MOD(OnServiceDel, (*provider));
 		}
 	}
 
@@ -488,7 +491,7 @@ void ModuleManager::DoSafeUnload(Module* mod)
 	Modules.erase(modfind);
 	ServerInstance->GlobalCulls.AddItem(mod);
 
-	ServerInstance->Logs.Normal("MODULE", "The {} module was unloaded", mod->ModuleFile);
+	ServerInstance->Logs.Normal("MODULE", "The {} module was unloaded", modfile);
 }
 
 void ModuleManager::UnloadAll()
--- a/src/modules.cpp
+++ b/src/modules.cpp
@@ -440,8 +440,6 @@ void ModuleManager::DoSafeUnload(Module* mod)
 
 	std::map<std::string, Module*>::iterator modfind = Modules.find(mod->ModuleFile);
 
-	std::string modfile = mod->ModuleFile;
-
 	// Unregister modes before extensions because modes may require their extension to show the mode being unset
 	UnregisterModes(mod, MODETYPE_USER);
 	UnregisterModes(mod, MODETYPE_CHANNEL);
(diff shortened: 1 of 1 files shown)


src/modules.cpp
```diff
@@ -488,7 +491,7 @@ void ModuleManager::DoSafeUnload(Module* mod)
 	Modules.erase(modfind);
 	ServerInstance->GlobalCulls.AddItem(mod);
 
-	ServerInstance->Logs.Normal("MODULE", "The {} module was unloaded", mod->ModuleFile);
+	ServerInstance->Logs.Normal("MODULE", "The {} module was unloaded", modfile);
```
sadiepowell: I don't think this is a use after free? The module isn't deleted immediately, its added to the global cull list and is then deleted at the end of the main loop (or, more likely on shutdown, from ModuleManager::UnloadAll).

sadiepowell:
Thanks for your report. In the future please raise crash issues privately with the team [following our security policy](https://github.com/inspircd/inspircd/security/policy).

Do you have a method of reproducing the crash? I see the invalidation issue with OnServiceDel but I can't actually get it to crash in testing.

aasb13:
Here's an example of a module that crashes on my machine:

```cpp
#include "inspircd.h"

class TestServiceDelModule final
    : public Module
{
public:
    TestServiceDelModule()
        : Module(VF_NONE, "Test")
    {
    }

    void OnServiceDel(ServiceProvider& provider) override
    {
        // use-after-free
        ServerInstance->Logs.Debug("TEST", "OnServiceDel called for service: {}", provider.name);

        if (provider.creator)
            ServerInstance->Logs.Debug("TEST", "Service provider module: {}", provider.creator->ModuleFile);
    }
};

MODULE_INIT(TestServiceDelModule)
```

Then, I have these modules

```
<module name="ssl_openssl">      
<module name="sslrehashsignal">

<module name="banexception">
<module name="callerid">
<module name="cban">
<module name="conn_join">
<module name="cap">
<module name="services">
<module name="sasl">
<module name="account">
<module name="cloak">           
<module name="cloak_user">
<module name="conn_umodes">
<module name="customprefix">
<module name="delayjoin">
<module name="delaymsg">
<module name="filter">
<module name="gecosban">
<module name="hidechans">
<module name="ident">
<module name="ircv3">
<module name="joinflood">
<module name="muteban">
<module name="namedmodes">
<module name="nicklock">
<module name="noctcp">
<module name="operlevels">
<module name="opermotd">
<module name="override">
<module name="remove">
<module name="sethost">         
<module name="swhois">
<module name="timedbans">
<module name="topiclock">
<module name="watch">

<module name="securelist">
<module name="restrictchans">

<module name="botmode">
<module name="commonchans">
<module name="deaf">
<module name="regex_stdlib">

<module name="chanhistory">
<chanhistory maxduration="4w"
             maxlines="500"
             prefixmsg="yes"
             savefrombots="yes"
             sendtobots="yes">

<module name="password_hash">
<mkpasswd operonly="no">
<module name="password_hash">
<module name="argon2">
<module name="bcrypt">
<module name="pbkdf2">

<module name="md5">
<module name="testservicedel">
```

And after inspircd starts, you can just gracefully shut it down and it will crash

`
kill -TERM $(pgrep inspircd)
`

or 

`
./inspircd restart
`

And then it crashes:

```
$ coredumpctl info 8048
        PID: 8048 (inspircd)
        Signal: 11 (SEGV)
        Command Line: /tmp/inspircd/run/bin/inspircd --config=/tmp/inspircd/run/conf/inspircd.conf
        Executable: /tmp/inspircd/run/bin/inspircd
        Unit: user@1000.service
        Slice: user-1000.slice
        Message: Process 8048 (inspircd) of user 1000 dumped core.
        Stack trace of thread 8048:
        #0  0x00007f366bf257ad n/a (/tmp/inspircd/run/modules/m_testservicedel.so + 0x47ad)
        #1  0x000055584e12adef n/a (/tmp/inspircd/run/bin/inspircd + 0xabdef)
        ELF object binary architecture: AMD x86-64
```

sadiepowell:
I can reproduce the `OnServiceDel` crash with your example module. Thanks.

Were you also getting a crash from `mod->ModuleFile`? I'm unable to reproduce that and as I mentioned in my review comment I think that use is fine.

aasb13:
About `mod->ModuleFile`, I didn't actually get a crash from accessing it, but it looked bad because the module is already in the cull list and marked for removal. Dereferencing there feels misleading. Saving the string earlier for logging would be essentially free. The actual crash I got came from `OnServiceDel`

sadiepowell:

PR #2212, opened 2026-05-29
mysql ssl - detect + add 5.7/mariadb fallback

grooverdan:
For detecting MYSQL_OPT_SSL_MODE, look at the define rather than the version.

Add a 5.7/MariaDB fallback using MYSQL_OPT_SSL_VERIFY_SERVER_CERT and MYSQL_OPT_SSL_ENFORCE.

Based on suggestion by Sergei Golubchik on:
https://jira.mariadb.org/browse/CONC-729

And cross reference with https://dev.mysql.com/doc/c-api/5.7/en/mysql-options.html

## Testing Environment

## Checks

I have ensured that:

- [X] Generative AI (Copilot, ChatGPT, etc) was not used to create any part of this pull request.
- [X] If the pull request contains a security fix I have followed the reporting rules mentioned in [the security policy](https://github.com/inspircd/inspircd/security/policy) (delete if not applicable).
- [N/A] I have documented any features added by this pull request (delete if not applicable).
- [X] This pull request does not introduce any incompatible API changes (stable branches only, delete if not applicable).
- [N/A] If ABI changes have been made I have incremented MODULE_ABI in `moduledefs.h` (stable branches only, delete if not applicable).

### DIFF
--- a/src/modules/extra/m_mysql.cpp
+++ b/src/modules/extra/m_mysql.cpp
@@ -334,9 +334,13 @@ public:
 		mysql_options(connection, MYSQL_OPT_CONNECT_TIMEOUT, &timeout);
 
 		// Enable SSL if requested.
-#if defined LIBMYSQL_VERSION_ID && LIBMYSQL_VERSION_ID > 80000
+#ifdef MYSQL_OPT_SSL_MODE
 		unsigned int ssl = config->getBool("ssl") ? SSL_MODE_REQUIRED : SSL_MODE_PREFERRED;
 		mysql_options(connection, MYSQL_OPT_SSL_MODE, &ssl);
+#elif defined MYSQL_OPT_SSL_VERIFY_SERVER_CERT && defined MYSQL_OPT_SSL_ENFORCE
+        unsigned my_bool ssl = config->getBool("ssl"), on= 1;
+        mysql_options(connection, MYSQL_OPT_SSL_VERIFY_SERVER_CERT, &ssl);
+        mysql_options(connection, MYSQL_OPT_SSL_ENFORCE, &on);
 #endif
 
 		// Attempt to connect to the database.


src/modules/extra/m_mysql.cpp
```diff
@@ -334,9 +334,13 @@ class SQLConnection final
...
 
 		// Enable SSL if requested.
-#if defined LIBMYSQL_VERSION_ID && LIBMYSQL_VERSION_ID > 80000
+#ifdef MYSQL_OPT_SSL_MODE
 		unsigned int ssl = config->getBool("ssl") ? SSL_MODE_REQUIRED : SSL_MODE_PREFERRED;
 		mysql_options(connection, MYSQL_OPT_SSL_MODE, &ssl);
+#elif defined MYSQL_OPT_SSL_VERIFY_SERVER_CERT && defined MYSQL_OPT_SSL_ENFORCE
+        unsigned my_bool ssl = config->getBool("ssl"), on= 1;
+        mysql_options(connection, MYSQL_OPT_SSL_VERIFY_SERVER_CERT, &ssl);
+        mysql_options(connection, MYSQL_OPT_SSL_ENFORCE, &on);
```
sadiepowell, line 343: Is this correct? It always sets `MYSQL_OPT_SSL_ENFORCE` to `1` regardless of the config value setting.

Additionally, this uses `MYSQL_OPT_SSL_VERIFY_SERVER_CERT` which I think introduces different behaviour for MariaDB than MySQL which I don't think does this with `SSL_MODE_REQUIRED`. We should probably introduce a strict mode for this in the future but for now it would be an incompatibility.


---

It'd be really nice if MariaDB could implement the recent MySQL changes or stop pretending to be MySQL. 🙃

grooverdan: Good points on logic flow, I was looking too casually at the implementation I copied off. I'll dig a bit deeper.

An unraveling of the SSL options and a good mode for https://jira.mariadb.org/browse/CONC-729 that is clear would be my preference too. Having a document of clear compatible usage for codebases that support connectors based on the same codebase would be good.

I'd like to know at what layer or where the "pretending to be MySQL" is? I've tried at the many levels to stop the pretending in messaging, documentation getting rid of "drop in replacement" and as much junk that could lead to such a claim. Does having a fork mean we're still pretending because the same base APIs are there? What would we realisticly do to convince people like yourself that we aren't pretending to be MySQL?

Honestly, we're upset that our userbase still calls us MYSQL - https://mariadb.org/i-am-not-building-cadillacs-anymore/. So please I'm keen to know what we can do to be recognised as different from the codebase we diverged from 17 years ago.

src/modules/extra/m_mysql.cpp
```diff
@@ -334,9 +334,13 @@ class SQLConnection final
 		mysql_options(connection, MYSQL_OPT_CONNECT_TIMEOUT, &timeout);
 
 		// Enable SSL if requested.
-#if defined LIBMYSQL_VERSION_ID && LIBMYSQL_VERSION_ID > 80000
+#ifdef MYSQL_OPT_SSL_MODE
```
sadiepowell, line 337: `MYSQL_OPT_SSL_MODE` is an enum value not a `#define` (in MySQL 8.4.9).

grooverdan:
Thanks @sadiepowell

PR #2213, opened 2026-05-30
Add support for HAProxy certificate fingerprint in PROXY protocol

xatlas:
## Summary

Add support for a custom HAProxy PROXY protocol v2 TLV (`PP2_TYPE_CERTFP = 0xE0`) that carries the hex encoded fingerprint of the client's TLS certificate, forwarded by HAProxy. When present, the fingerprint is used to populate the fake `ssl_cert` that `m_haproxy` already creates for SSL-terminated connections.

## Rationale

When HAProxy terminates TLS on behalf of InspIRCd, the server never sees the client's certificate. This means `SASL EXTERNAL` authentication against services (e.g. Atheme's saslserv/external) doesn't work, and `/WHOIS` shows no certificate fingerprint.

HAProxy already supports forwarding the client certificate fingerprint as a custom TLV. This change reads that TLV, and attaches it to the fake `ssl_cert` so the rest of InspIRCd treats the connection exactly as if TLS terminated locally.

Connections that don't include the `0xE0` TLV are unaffected, so the existing behaviour is preserved.

Example of required HAProxy configuration using multiple hash types:
```
backend irc_backend
    mode tcp
    balance roundrobin
    # Using the native ssl_c_sha1 function
    server inspircd_node1 <ip>:<port> send-proxy-v2-ssl set-proxy-v2-tlv-fmt(0xe0) "%[ssl_c_sha1,hex,lower]"
    # Using the native ssl_c_der function
    server inspircd_node5 <ip>:<port> send-proxy-v2-ssl set-proxy-v2-tlv-fmt(0xe0) "%[ssl_c_der,sha1,hex,lower]"
    server inspircd_node2 <ip>:<port> send-proxy-v2-ssl set-proxy-v2-tlv-fmt(0xe0) "%[ssl_c_der,sha2(256),hex,lower]"
    server inspircd_node3 <ip>:<port> send-proxy-v2-ssl set-proxy-v2-tlv-fmt(0xe0) "%[ssl_c_der,sha2(384),hex,lower]"
    server inspircd_node4 <ip>:<port> send-proxy-v2-ssl set-proxy-v2-tlv-fmt(0xe0) "%[ssl_c_der,sha2(512),hex,lower]"
```
Obviously, all servers should use the same hash conversion. These are just examples of use.

## Testing Environment

I tested this change in a production-like setup consisting of multiple InspIRCd nodes behind an HAProxy load balancer with TLS termination. HAProxy was configured to forward the client certificate SHA-1 fingerprint via the custom `0xE0` TLV using `%[ssl_c_sha1,hex,lower]`. The connection from HAProxy to InspIRCd is plain TCP (non-TLS), as is expected in this architecture; the HAProxy listener should be restricted to trusted local or internal network traffic only.

On connect, clients correctly received their certificate fingerprint in `/WHOIS` and `/SSLINFO`. I then registered the fingerprint with Atheme services using `/ns CERT ADD`, and on subsequent reconnections Atheme automatically recognised the certificate and logged the user in without a password prompt, confirming end-to-end SASL EXTERNAL authentication works as expected.

**Operating system name and version:** Linux 6.12.90+deb13-cloud-amd64
**Compiler name and version:** gcc (Debian 14.2.0-19) 14.2.0

## Checks

I have ensured that:

- [x] Generative AI (Copilot, ChatGPT, etc) was not used to create any part of this pull request.
- [ ] I have documented any features added by this pull request (delete if not applicable).

### DIFF
--- a/src/modules/m_haproxy.cpp
+++ b/src/modules/m_haproxy.cpp
@@ -21,7 +21,6 @@
 
 #include "inspircd.h"
 #include "iohook.h"
-#include "stringutils.h"
 #include "modules/ssl.h"
 
 enum
--- a/src/modules/m_haproxy.cpp
+++ b/src/modules/m_haproxy.cpp
@@ -158,7 +158,7 @@ private:
 	std::string certificate_fingerprint;
 
 	// The fake certificate we attach for this connection, if any.
-	ssl_cert* certificate = nullptr;
+	reference<ssl_cert> certificate;
 
 	// The current state of the PROXY parser.
 	HAProxyState state = HPS_WAITING_FOR_HEADER;
--- a/src/modules/m_haproxy.cpp
+++ b/src/modules/m_haproxy.cpp
@@ -58,6 +58,9 @@ enum
 	// The identifier for a TLS TLV entry.
 	PP2_TYPE_SSL = 0x20,
 
+	// The identifier for a TLS certificate fingerprint TLV entry.
+	PP2_TYPE_CERTFP = 0xE0,
+
 	// The minimum length of a PP2_TYPE_SSL TLV entry.
 	PP2_TYPE_SSL_LENGTH = 5,
 
@@ -150,6 +153,12 @@ private:
 	// The API for interacting with user TLS internals.
 	UserCertificateAPI& sslapi;
 
+	// The fingerprint forwarded by the proxy for the client certificate.
+	std::string certificate_fingerprint;
+
+	// The fake certificate we attach for this connection, if any.
+	ssl_cert* certificate = nullptr;
+
 	// The current state of the PROXY parser.
 	HAProxyState state = HPS_WAITING_FOR_HEADER;
 
@@ -172,15 +181,62 @@ private:
 		}
 
 		// What type of TLV are we parsing?
-		switch (recvq[start_index])
+		switch (static_cast<uint8_t>(recvq[start_index]))
 		{
 			case PP2_TYPE_SSL:
 				if (!ReadProxyTLVSSL(sock, start_index + PP2_TLV_LENGTH, length))
 					return 0;
 				break;
+
+			case PP2_TYPE_CERTFP:
+				if (!ReadProxyTLVCertFP(sock, start_index + PP2_TLV_LENGTH, length))
+					return 0;
+				break;
 		}
 
 		return PP2_TLV_LENGTH + length;
+	}
+
+	bool ReadProxyTLVCertFP(StreamSocket* sock, size_t start_index, uint16_t buffer_length)
+	{
+		// If the socket is not a user socket we don't have to do
+		// anything with this TLV's information.
+		if (sock->type != StreamSocket::SS_USER)
+			return true;
+
+		// The fingerprint must be a 20-byte raw SHA-1 digest.
+		if (buffer_length != 20)
+		{
+			ServerInstance->Logs.Debug(MODNAME, "Ignoring PP2_TYPE_CERTFP TLV with unexpected length {} (expected 20)", buffer_length);
+			return true;
+		}
+
+		std::string& recvq = GetRecvQ();
+
+		// Convert to lowercase hex to match InspIRCd's fingerprint format.
+		static constexpr char hex[] = "0123456789abcdef";
+		certificate_fingerprint.resize(40);
+		for (size_t i = 0; i < 20; ++i)
+		{
+			const uint8_t byte = static_cast<uint8_t>(recvq[start_index + i]);
+			certificate_fingerprint[i * 2]     = hex[byte >> 4];
+			certificate_fingerprint[i * 2 + 1] = hex[byte & 0x0F];
+		}
+
+		ServerInstance->Logs.Debug(MODNAME, "Received certificate fingerprint from HAProxy: {}", certificate_fingerprint);
+
+		// If the SSL TLV was already processed, patch the existing certificate
+		// with the fingerprint now that we have it.
+		if (certificate)
+		{
+			certificate->fingerprints.push_back(certificate_fingerprint);
+			certificate->invalid = false;
+			certificate->trusted = true;
+			certificate->unknownsigner = false;
+			certificate->revoked = false;
+			certificate->error.clear();
+		}
+		return true;
 	}
 
 	bool ReadProxyTLVSSL(StreamSocket* sock, size_t start_index, uint16_t buffer_length)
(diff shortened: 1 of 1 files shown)


src/modules/m_haproxy.cpp
```diff
@@ -172,17 +181,64 @@ class HAProxyHook final
...
+
+		// Convert to lowercase hex to match InspIRCd's fingerprint format.
+		static constexpr char hex[] = "0123456789abcdef";
+		certificate_fingerprint.resize(40);
+		for (size_t i = 0; i < 20; ++i)
+		{
+			const uint8_t byte = static_cast<uint8_t>(recvq[start_index + i]);
+			certificate_fingerprint[i * 2]     = hex[byte >> 4];
+			certificate_fingerprint[i * 2 + 1] = hex[byte & 0x0F];
+		}
```
sadiepowell: Include `stringutils.h` and you can use `Hex::Encode(void*,size_t)` here.

src/modules/m_haproxy.cpp
```diff
@@ -172,17 +181,64 @@ class HAProxyHook final
...
 	}
 
+	bool ReadProxyTLVCertFP(StreamSocket* sock, size_t start_index, uint16_t buffer_length)
+	{
+		// If the socket is not a user socket we don't have to do
+		// anything with this TLV's information.
+		if (sock->type != StreamSocket::SS_USER)
+			return true;
+
+		// The fingerprint must be a 20-byte raw SHA-1 digest.
```
sadiepowell: Is it possible to somehow make this modular? Most people use SHA-256 fingerprints nowadays.

src/modules/m_haproxy.cpp
```diff
@@ -150,6 +153,12 @@ class HAProxyHook final
 	// The API for interacting with user TLS internals.
 	UserCertificateAPI& sslapi;
 
+	// The fingerprint forwarded by the proxy for the client certificate.
+	std::string certificate_fingerprint;
+
+	// The fake certificate we attach for this connection, if any.
+	ssl_cert* certificate = nullptr;
```
sadiepowell:
```suggestion
	reference<ssl_cert> certificate;
```

If you're storing this for longer term it should be held as a reference (the reference code is confusing but consider it similar to `std::shared_ptr<>`, which is what its been replaced with in master). You can assign the pointer to it as normal.

sadiepowell:
Thanks for your pull request. I was wondering if there was a way to do this when I wrote the HAProxy code but I couldn't find anything at the time.

src/modules/m_haproxy.cpp
xatlas: Cool, thanks

src/modules/m_haproxy.cpp
xatlas: Fixed, thanks!

src/modules/m_haproxy.cpp
xatlas: HAProxy's built-in TLS fetch functions only expose `ssl_c_sha1`, so there is no native way to forward a SHA-256 fingerprint today: https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#7.3.4

I think we can use Lua scripting to inject arbitrary data into custom TLVs, so a SHA-256 digest via `0xE0` could be technically possible. But this is a path that I haven't explored yet.

With that in mind, I've relaxed the validation from a strict `buffer_length != 20` check to accept any non-empty digest up to 64 bytes (SHA-512). What do you think?

sadiepowell: If its not possible at the moment then that's okay. Relaxing the check seems like a good option because if they ever add support for SHA-256 to HAProxy in the future it'll just work.

xatlas: I'm testing a Lua script to send a SHA-256 from HAProxy and found some issues sending the raw digest to InspIRCd. I will delegate the hex conversion to HAProxy and let InspIRCd get a clean fingerprint, so it will be easier to accept different hash types.

xatlas: Actually, we can get the client's certificate in DER format:
- https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#7.3.4-ssl_c_der
- https://www.haproxy.com/documentation/haproxy-configuration-manual/latest/#7-sha2

With this function, we can get the certificate and calculate the hash using `%[ssl_c_der,sha2(256),hex,lower]` natively in HAProxy. So we don't need to rely on Lua scripting.

sadiepowell:
This looks good to me now. It might take me a few days to get around to testing it but I'd definitely like to see this in the next release. Thanks! <3

src/modules/m_haproxy.cpp
```diff
@@ -150,6 +153,12 @@ class HAProxyHook final
 	// The API for interacting with user TLS internals.
 	UserCertificateAPI& sslapi;
 
+	// The fingerprint forwarded by the proxy for the client certificate.
+	std::string certificate_fingerprint;
```
sadiepowell, line 157: If this is assigned to then immediately moved into `certificate->fingerprints` is there a need for it to be a non-local variable?

xatlas: Yes, it is used below at line 274.

Since `0xE0` falls in the [application-specific TLV range](https://github.com/haproxy/haproxy/blob/83634a4c9a2721c90411bd232e6d947c9171d0b5/doc/proxy-protocol.txt#L701) (`PP2_TYPE_MIN_CUSTOM`) and, AFAIK, the PROXY protocol does not guarantee TLV ordering, I decided to handle both possible orderings:

- `PP2_TYPE_SSL` arrives first, indicating the client is connecting over TLS. At this point `certificate_fingerprint` is still empty, so a fake cert is created without a fingerprint and attached to the user via `sslapi->SetCertificate`. The certificate member retains a reference to it. If `PP2_TYPE_CERTFP` arrives afterwards, it validates and stores the fingerprint, then checks if (certificate) to detect that the SSL TLV was already processed, and patches the fingerprint directly onto the existing cert object.

- `PP2_TYPE_CERTFP` arrives first, the fingerprint is validated and stored in `certificate_fingerprint`. When `PP2_TYPE_SSL` is subsequently processed, it finds a non-empty `certificate_fingerprint` and creates the fake cert already populated with it.

I hope this clarifies the design. Happy to answer any further questions.

sadiepowell:
I've tested this and it seems to work good and I'd be happy to merge as-is but I've come up with two minor suggestions that you might want to consider:

- It would be good to `.clear()` and `.shrink_to_fit()` `certificate_fingerprint` once the certificate has been initialised to reduce memory usage.

- If possible it'd be good to deduplicate some of the code initialising the members of `this->certificate`.

xatlas:
Glad to read that! I agree with your suggestions and added a method to deduplicate some code.

I also ran a quick microbenchmark test with 100 million iterations, compiled at `-O2`, and found that `std::move` cuts the cost per operation from about 49 ns to around 29 ns for 40, 64, and 128-byte digests.

From what I could see in the tests, adding `clear()` or `shrink_to_fit()` right after the move actually seems to add a small overhead penalty (likely due to internal checks, given that the string's capacity is transferred to the vector upon moving)

If I'm missing something, please let me know. But based on those numbers, leaving `std::move` on its own seems to be an optimal approach here.

sadiepowell:
Moving instead of copying and then resetting also works here.

I think this PR is ready to merge now. Unless you have any last changes you want to make I'll merge this tomorrow (about 12~ hours from now). I'll also put your example HAProxy backend config on the docs site when I make the next release.

Thanks for your contribution! <3

sadiepowell:
Merged, thanks for your contribution! 🫡

PR #2215, opened 2026-06-20
Fix websock pings

Kufat:
## Summary

Pass StreamSock to Ping() so FD_ADD_TRIAL_WRITE can be set; set it for native websocket pings.

## Rationale

With nativeping enabled, websocket clients will currently time out when they're not receiving events (e.g. no/idle channels.)

## Testing Environment

**Operating system name and version:** Ubuntu 24.04.4 LTS (5.10.102) under WSL
**Compiler name and version:** GCC 13.3.0

## Checks

Tested locally with and without the `ChangeEventMask()` call. Timed out without, worked with.

I have ensured that:

- [X] Generative AI (Copilot, ChatGPT, etc) was not used to create any part of this pull request.
- [X] If ABI changes have been made I have incremented MODULE_ABI in `moduledefs.h` (stable branches only, delete if not applicable).

### DIFF
--- a/include/moduledefs.h
+++ b/include/moduledefs.h
@@ -22,7 +22,7 @@
 class Module;
 
 /** The version of the InspIRCd ABI which is presently in use. */
-#define MODULE_ABI 4012UL
+#define MODULE_ABI 4011UL
 
 /** Stringifies the value of a symbol. */
 #define MODULE_STRINGIFY_SYM1(DEF) MODULE_STRINGIFY_SYM2(DEF)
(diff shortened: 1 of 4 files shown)


include/moduledefs.h
```diff
@@ -22,7 +22,7 @@
 class Module;
 
 /** The version of the InspIRCd ABI which is presently in use. */
-#define MODULE_ABI 4011UL
+#define MODULE_ABI 4012UL
```
sadiepowell: We will bump this when the next release happens so there's no need for you to do it.

